PR #78, opened 2023-06-27
fix: enable toggle collapsible without providing v-model

khairulhaaziq:
- [ ] enable toggle collapsible without providing v-model
- [ ] enable custom visibility components inside with v-slot

### DIFF
--- a/packages/playground-vue/src/components/Demo/CollapsibleDemo.vue
+++ b/packages/playground-vue/src/components/Demo/CollapsibleDemo.vue
@@ -7,12 +7,11 @@ import {
 } from "radix-vue";
 import { Icon } from "@iconify/vue";
 
-const rootOpen = ref(false);
 const rootDisabled = ref(false);
 </script>
 
 <template>
-  <CollapsibleRoot class="w-[300px]" v-model:open="rootOpen"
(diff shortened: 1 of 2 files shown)


mujahidfa:
> Before: we need to pass v-model to toggle collapsible state

Not sure what you mean by this, because it should work with and even without v-model?

Original version for reference:

```vue
<script setup lang="ts">
import { provide, toRef } from "vue";

const props = withDefaults(defineProps<CollapsibleRootProps>(), {
  open: false,
});
const emit = defineEmits<{
  (e: "update:open", value: boolean): void;
}>();

provide<CollapsibleProvideValue>(COLLAPSIBLE_INJECTION_KEY, {
  contentId: "1",
  disabled: toRef(() => props.disabled),
  open: toRef(() => props.open),
  onOpenToggle: () => {
    emit("update:open", !props.open);
  },
});
</script>

<template>
  <div
    :data-state="props.open ? 'open' : 'closed'"
    :data-disabled="props.disabled ? 'true' : undefined"
  >
    <slot :open="props.open" />
  </div>
</template>
```

mujahidfa:
Also v-model doesn't work with this PR. Reverting this change in #81 (with added v-slot).

Issue #70, opened 2023-06-26
Feat: asChild or as="template" implementation


khairulhaaziq:

zernonia:
Hi guys! I love the structure and how clean the approach is for developing Radix-Vue!

I saw that `asChild` is a huge challenge for Svelte, but I came across [Ark UI](https://ark-ui.com/) recently,  and saw how they implement it.
Perhaps this could be an inspiration!

https://github.com/chakra-ui/ark/blob/4c7937abd15a5f184b746c20a4b1e68dee197e26/packages/vue/src/factory.tsx#L126

I would love to contribute to this (soon to be) amazing library if you guys are ok with it!! 
 

(btw, I'm a fellow Malaysian too! 🤩🙌🏼)

PR #84, opened 2023-06-27
feat: add Primitive component, with asChild utility

zernonia:
## Description

Implement `asChild` props similar to Radix UI (by using a custom wrapped component, namely `Primitive`.)
Inspired by [Ark UI]() approaches, but instead of using `jsx`, we implemented a `vue-sfc` approach.

## Related Issue
#70 

 
## How Has This Been Tested?
Updated the demo on `CollapsibleDemo`. You could inspect the element via browser devtools to see that there's only 1 button element.

image


khairulhaaziq (changes requested):
Simply genius! Looks good! however I think we should refrain from using the namescaped naming, is it necessary, also I think we should find a better name, maybe AsChildWrapper? if you think we should discuss the name further lets do on discord

zernonia:
@khairulhaaziq  I've changed from using namespaced, to individual component. 
I think naming `Primitive` will help other dev to understand that it serves the same purpose as the "Primitive" component in Radix UI

khairulhaaziq:
> @khairulhaaziq I've changed from using namespaced, to individual component. I think naming `Primitive` will help other dev to understand that it serves the same purpose as the "Primitive" component in Radix UI

cant the html/tag element be inferred from the slots? meaning the Primitive is dynamic, inferred from the child. is that possible?

zernonia:
> Edit: Ah sorry I just take my time on it, so the PrimitiveButton only defaults to button, and if the asChild attribute is added, it resolve to the child element of any tag, correct?

Yup. How it works is that by default it will render a `button` (using PrimitiveButton). But if we pass `asChild` props, it will render whatever child tag

khairulhaaziq (approved):
lgtm!

Issue #22, opened 2023-06-23
Feat: Switch


khairulhaaziq:
[reference](https://www.radix-ui.com/docs/primitives)

### Root

Contains all the parts of a switch. An `input` will also render when used within a `form` to ensure events propagate correctly.
| Status | Name              | Required | Type     | Default | Description                                                                                                                   |
|--------|-------------------|----------|----------|---------|-------------------------------------------------------------------------------------------------------------------------------|
| asChild           | false    | boolean  | false   | Change the default rendered element for the one passed as a child, merging their props and behavior.                         |
| defaultChecked    |          | boolean  |         | The state of the switch when it is initially rendered. Use when you do not need to control its state.                         | ⚠️   | 
| checked           |          | boolean  |         | The controlled state of the switch. Must be used in conjunction with `onCheckedChange`.                                      |  ⚠️  | 
| onCheckedChange   |          | function |         | Event handler called when the state of the switch changes.                                                                    |  ⚠️    | 
 | disabled          |          | boolean  |         | When true, prevents the user from interacting with the switch.                                                                |  ✅  | 
| required          |          | boolean  |         | When true, indicates that the user must check the switch before the owning form can be submitted.                             |  ✅   | 
 | name              |          | string   |         | The name of the switch. Submitted with its owning form as part of a name/value pair.                                         |  ✅  | 
| value             |          | string   | on      | The value given as data when submitted with a `name`.                                                                         |  ✅   | 

| Status | Attribute          | Values                |
|--------|--------------------|-----------------------|
| [data-state]       | checked, unchecked    |  ✅   | 
| [data-disabled]    | Present when disabled |  ✅   | 

### Thumb
The thumb that is used to visually indicate whether the switch is on or off.
| Status | Name              | Required | Type     | Default | Description                                                                                                                   |
|--------|-------------------|----------|----------|---------|-------------------------------------------------------------------------------------------------------------------------------|
| asChild           | false    | boolean  | false   | Change the default rendered element for the one passed as a child, merging their props and behavior.                         |  ✅  | 

| Status | Attribute          | Values                |
|--------|--------------------|-----------------------|
| [data-state]       | checked, unchecked    |   ✅   | 
 | [data-disabled]    | Present when disabled |   ✅   | 

## Accessibility
| Status | Keys      | Description                 |
|--------|-----------|-----------------------------|
| Space     | Toggles the component's state. |  ✅   | 
 | Enter     | Toggles the component's state. |  ✅  |

PR #90, opened 2023-06-28
fix: make switch controllable and unctrollable

khairulhaaziq:
make switch controllable via v-model or uncontrolled(works without v-model)
#22

### DIFF
--- a/packages/radix-vue/src/Switch/SwitchRoot.vue
+++ b/packages/radix-vue/src/Switch/SwitchRoot.vue
@@ -24,24 +24,31 @@ import { ref, provide, watch, toRef } from "vue";
 
 const props = withDefaults(defineProps<SwitchRootProps>(), {
   disabled: false,
+  modelValue: undefined,
 });
 
 const emit = defineEmits(["update:modelValue"]);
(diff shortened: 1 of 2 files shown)


zernonia:
Could we introduce `@vueuse/core` into the dependencies? I think `useVModel` handle this case nicely.

khairulhaaziq:
> Could we introduce `@vueuse/core` into the dependencies? I think `useVModel` handle this case nicely.

@zernonia we have thought and discussed about it, while we are only using small number of utilities, we think we better try to write/copy it than importing a huge dependency.

zernonia:
Vueuse does have tree-shaking capabilities.. so I guess that should be fine right?
But yeah.. No problem with current approach. I just foresee that we might eventually need vueuse 😁

khairulhaaziq:
> Vueuse does have tree-shaking capabilities.. so I guess that should be fine right? But yeah.. No problem with current approach. I just foresee that we might eventually need vueuse 😁

do you want to test and try adding it? see how much it effects the build size and the time to build? if not too much of difference we can consider it

PR #97, opened 2023-06-28
feat: ScrollArea (1/3)

zernonia:


zernonia:
fianlly some progress!! this component is more complicated than I thought! 

https://github.com/radix-vue/radix-vue/assets/59365435/72fd4c3d-bd73-41d4-9fb3-287d967e2987

khairulhaaziq (approved):
Lgtm!

packages/radix-vue/src/ScrollArea/ScrollAreaViewport.vue
```diff
@@ -0,0 +1,61 @@
...
+       */
+      overflowX: injectedValueFromRoot?.scrollbarXEnabled.value
+        ? 'scroll'
+        : 'hidden',
+      overflowY: injectedValueFromRoot?.scrollbarYEnabled.value
+        ? 'scroll'
+        : 'hidden',
+    }"
+  >
+    <div ref="contentElement" :style="{ minWidth: '100%', display: 'table' }">
```
simonbrunel, line 45: @zernonia I'm experiencing some layout issues with this component and it seems caused by this `display: 'table'`. What the reason behind using "table" here?

zernonia: We're not 100% sure regarding the reason for this, but I've tested out multiple case and seems like we don't really need it

Issue #98, opened 2023-06-28
fix: fix all components that should be controllable and uncontrollable


khairulhaaziq:
Fix relevant components to be able to be used with or without v-model(controllable or uncontrollable)
Use the useVueModel, refer to #95 
Related to all components with open/close
- [ ] #1  ~wait until built first
- [x] #2 
- [x] #5
- [x] #6 
- [x] #8  
- [x] #9 
- [x] #11 
- [ ] #13 ~wait until built first
- [ ] #14 ~wait until built first
- [x] #15 
- [ ] #17 ~needs refactor to support without v-model first
- [ ] #19 ~wait until built first
- [ ] #21 ~needs refactor to support without v-model first
- [x] #22 #91 
- [ ] #23 ~needs refactor to support without v-model first
- [x] #25 
- [ ] #26 ~needs refactor to support without v-model first
- [x] #28

PR #99, opened 2023-06-28
fix: enable components to be controlled or uncontrolled

khairulhaaziq:
Enable components to be controlled or uncontrolled, functioning either by providing state via v-model or without.

## Description

Make the state reactive via v-model input and able to work on its own without needing to provide state.

## Related Issue

- [ ] #98 

## Motivation and Context

Some components shouldnt need to be provided state manually to work, e.g. collapsible or popover. Some others needs to be handled via outside function that mutates the v-model binded, a 2-way binding must work perfectly.

## How Has This Been Tested?

In demo page we will add a custom button that mutates the v-model input to the components. Other than that, we will try and remove v-model to see if it works without it.


khairulhaaziq:
this isnt completed, merge first, we'll complete it later

Issue #102, opened 2023-06-28
feat: declaring props to all components


khairulhaaziq:
implementing props is one thing, declaring is another.

By declaring every props and typing them Im hoping that we can speed up the development even further.

PR #103, opened 2023-06-28
feat: declare all props to every component

khairulhaaziq:
- [ ] #102 

These left:
- [x] Alert Dialog
- [ ] Context Menu
- [x] Dialog
- [x] DropdownMenu
- [x] HoverCard
- [x] Popover

note: DropdownMenuGroup props is incorrect


khairulhaaziq:
Every built component done except context menu, and minor incorrect props in DropdownMenuGroup, merge first, will handle later.

PR #113, opened 2023-07-02
feat: allow defaultValue props, able to work without v-model

zernonia:
Remove dependencies on v-model. So tabs should now be working without v-model, as it will have it's own local state now.


khairulhaaziq (approved):
⚡️

PR #121, opened 2023-07-05
fix: isPresent still true when no animation is provide

zernonia:
add a state to indicate if there's animation present in the element.

If `no`, we would immediately set `isPresent.value` false as there's no event listener of `animationend`


khairulhaaziq (approved):
lgtm

Issue #129, opened 2023-07-07
[Chore]: Primitive `asChild` props throws warning message.


zernonia:
When using `asChild` prop, Vue logged warning of `Non-function value encountered for default slot. Prefer function slots for better performance.`.

- [ ] Change to use function slots
- [ ] Make sure the warning goes away

PR #150, opened 2023-07-07
fix: primitive `aschild` not working properly

zernonia:
because `AccordionTrigger` is wrapped with `CollapsibleTrigger`, somehow that makes the `asChild` doesn't work.

this would closed off #129  as well, as we are not passing `slots.default?.()` now


khairulhaaziq (approved):

zernonia:
Tested with nested `asChild`, as wriitten in the docs's PR,  seems to be working too!

image

Issue #13, opened 2023-06-23
Feat: Menubar


khairulhaaziq:
[reference](https://www.radix-ui.com/docs/primitives)

PR #158, opened 2023-07-10
[Menubar]feat: 1/3

khairulhaaziq:
Menubar initial component

## Todo:
- [x] Align content text with trigger
- [x] Redirect initial focus to first item
- [x] save last selected item tabindex to be used for next initial focus
- [x] focus management
- [x] handle escape key

### Related
#13

### DIFF
--- a/packages/radix-vue/src/Menubar/MenubarItem.vue
+++ b/packages/radix-vue/src/Menubar/MenubarItem.vue
@@ -28,7 +28,7 @@ const subInjectedValue = inject<MenubarSubProvideValue>(
 const props = defineProps<MenubarItemProps>();
 
 function handleClick() {
-  rootInjectedValue!.setIsOpen(false);
+  rootInjectedValue!.changeValue(undefined);
   rootInjectedValue!.changeSelected(
     rootInjectedValue!.triggerElement.value as HTMLElement
   );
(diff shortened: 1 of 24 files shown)


packages/radix-vue/src/Menubar/MenubarItem.vue
```diff
@@ -0,0 +1,71 @@
...
+  MENUBAR_SUB_INJECTION_KEY
+);
+
+const props = defineProps<MenubarItemProps>();
+
+function handleClick() {
+  if (rootInjectedValue?.selectedElement.value) {
+    rootInjectedValue.hideTooltip();
+  }
+}
```
zernonia, line 34: this doesn't seems to close the menu dropdown. maybe `rootInjectedValue.setIsOpen(false)` ?

packages/radix-vue/src/Menubar/MenubarRoot.vue
```diff
@@ -0,0 +1,116 @@
...
+}
+
+export const MENUBAR_INJECTION_KEY =
+  Symbol() as InjectionKey<MenubarProvideValue>;
+
+export type MenubarProvideValue = {
+  selectedElement: Ref<HTMLElement | undefined>;
+  isOpen: Readonly<Ref<boolean>>;
+  setIsOpen: (val: boolean) => void;
+  toggleIsOpen: () => void;
```
zernonia, line 22: maybe we should use `useVModel` for `props.open`, since that is a prop?

packages/radix-vue/src/Menubar/MenubarMenu.vue
```diff
@@ -0,0 +1,75 @@
+<script lang="ts">
+import type { Ref, InjectionKey } from "vue";
+import type { DataOrientation } from "@/shared/types";
+import { useVModel } from "@vueuse/core";
+
+export interface MenubarMenuRootProps {
+  modelValue?: boolean;
```
zernonia, line 7: I think `MenubarMenu` doesn't required a modelValue, but instead a value.
If this value matches the MenubarRoot's  modelValue, then the menu will open.

reference: https://www.radix-ui.com/docs/primitives/components/menubar#menu

zernonia (approved):

Issue #19, opened 2023-06-23
Feat: Select


khairulhaaziq:
[reference](https://github.com/radix-ui/primitives/blob/main/packages/react/select/src/Select.tsx)

PR #159, opened 2023-07-10
[Select]feat: 1/3

khairulhaaziq:
Todo:
- [x] Support single and multiple options
- [x] Fix autoplacement on collision
## Related
#19


khairulhaaziq:
it is scrappy but check it out @zernonia

packages/radix-vue/src/Select/SelectGroup.vue
```diff
@@ -0,0 +1,73 @@
...
+});
+
+const emits = defineEmits(["update:modelValue"]);
+
+const { primitiveElement, currentElement: parentElement } =
+  usePrimitiveElement();
+
+provide<SelectGroupProvideValue>(SELECT_GROUP_INJECTION_KEY, {
+  type: props.type,
+  modelValue: toRef(() => props.modelValue),
```
zernonia, line 45: perhaps we could use `useVModel` here?

khairulhaaziq: sure, there's a ton of refactoring needed here, especially to support multiple select, I will and do tonight. This especially, we dont actually need a v-model in the group since no value are gonna passed to the group, this is just because I reused from the previous dropdownmenu component

zernonia: @khairulhaaziq  is there a multiple select features for Select component?

khairulhaaziq: > @khairulhaaziq is there a multiple select features for Select component?

Ouh just saw radix doesnt have it, we may extend it, I really like how easy to do multiple select with headless UI, API can be similar to radix’s accordion

zernonia: True. I think this could be an enhancement in coming day

Issue #14, opened 2023-06-23
Feat: Navigation Menu


khairulhaaziq:
[reference](https://www.radix-ui.com/docs/primitives)

PR #167, opened 2023-07-11
[Navigation Menu]feat: 1/3

khairulhaaziq:
## Todos
- [x] Make modals inside viewport
- [x] Make arrow center relative to button
- [x] Make viewport center relative to list container
- [x] Make viewport dynamic width/height

### Related
#14


khairulhaaziq:
@zernonia do you think the navigation menu viewport has any similarity with scroll area's viewport?
https://www.radix-ui.com/docs/primitives/components/navigation-menu

zernonia:
> @zernonia do you think the navigation menu viewport has any similarity with scroll area's viewport? https://www.radix-ui.com/docs/primitives/components/navigation-menu

yeah there's some similarity there

khairulhaaziq (commented):
There are several issues:
- [ ] When focus is trapped inside navigation menu items, when shift+tab on first item it should return to trigger, and pressing tab on last item should focus on next trigger.
- [ ] When a menu is opened with keyboard, it cant be closed when clicking outside
- [x] We want to disable loop navigation for the items 45b81d2

zernonia:
The keyboard nav should be similar RadixUI now. I've check against the list above, and seems like work fine! 😁

@khairulhaaziq

khairulhaaziq (commented):
so good!

Issue #172, opened 2023-07-12
[Docs]feat: Convert mdx to mdc


khairulhaaziq:
Currently the docs pages under '/components' were all copied from radix. What we need to do first is we need to convert from the current mdx syntax to [mdc](https://content.nuxtjs.org/guide/writing/mdc/). We have converted the Alert Dialog as an example to be applied to every other components.

### Todos
- [ ] Convert all current component pages to mdc syntax, refer to Alert Dialog page.
- [ ] Some components arent imported from Radix because we just had done it recently, (DropdownMenu, ContextMenu, Menubar...), so add it first then convert it.
- [ ] Review all api/props, examples, make sure everything is correct and working.

DennisSmuda:
This can be closed, I guess right? @khairulhaaziq

PR #176, opened 2023-07-13
[Docs] Convert mdx to mdc

DennisSmuda:
Converts mdx to mdc for documentation content files.

## Description
Currently the docs pages under '/components' were all copied from radix. What we need to do first is we need to convert from the current mdx syntax to [mdc](https://content.nuxtjs.org/guide/writing/mdc/). We have converted the Alert Dialog as an example to be applied to every other components.

## Related Issue
https://github.com/radix-vue/radix-vue/issues/172

### DIFF
--- a/docs/content/3.components/accordion.md
+++ b/docs/content/3.components/accordion.md
@@ -16,11 +16,11 @@ associated section of content.
 #previewSlot
   :demo-accordion
 #codeSlot
-::hero-code-group{folder="Accordion"}
-::
+  ::hero-code-group{folder="Accordion"}
 ::
 
 ## Features
+
 :br
 ::list{type="success"}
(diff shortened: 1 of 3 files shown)


khairulhaaziq:
@DennisSmuda how good are you with css/tailwind?

DennisSmuda:
@khairulhaaziq I personally would prefer [yaml syntax](https://content.nuxtjs.org/guide/writing/mdc#yaml-method) for the table data, what do you think?

khairulhaaziq:
@DennisSmuda appreciate the work! But please see discord. We should work on new docs instead of this which is the old docs

DennisSmuda:
based off of `docs` instead of `new-docs` 🔥

Issue #145, opened 2023-07-07
[Tabs]feat: 2/3


khairulhaaziq:
Phase 2 of this project development is about making all API such as props, emits and data-attributes, work as expected. Below is the list of API needed to be completely working to get through the phase 2 of development.

The list is structured as below:
Prop/emit/data-attributes | Type | Default
### Props
- [ ] asChild boolean false
- [ ] defaultValue string - 
- [ ] value string -
- [ ] onValueChange function -
- [ ] orientation "horizontal"|"vertical" "horizontal"
- [ ] dir "ltr" | "rtl" -
- [ ] activationMode "automatic" | "manual" -

### Data Attributes
- [ ] [data-orientation] "vertical"|"horizontal"

### Props
- [ ] asChild boolean false
- [ ] loop boolean true

### Data Attributes
- [ ] [data-orientation] "vertical"|"horizontal"

### Props
- [ ] asChild boolean false
- [ ] value* string -
- [ ] disabled boolean false

### Data Attributes
- [ ] [data-state] "active"|"inactive"
- [ ] [data-disabled] present when disabled
- [ ] [data-orientation] "vertical"|"horizontal"

## Related
#23

onmax:
I am working on this. I see that there is a "Thumb" Component. I guess this is a mistake from copy and paste from another issue?

khairulhaaziq:
Ah yes probably,

> I am working on this. I see that there is a "Thumb" Component. I guess this is a mistake from copy and paste from another issue?

Ah yes sorry. Please refer to original radix for confirmation

onmax:
Should I fork the repo and create the draft PR from my forked repo or should I request access to push?

khairulhaaziq:
@onmax you can fork it, create a new branch and then create draft pr

khairulhaaziq:
closed with #182

PR #182, opened 2023-07-14
Tabs

onmax:
Implemented `Tabs` components

## Description
Using the [Radix API](https://www.radix-ui.com/docs/primitives/components/tabs#content)

## Related Issue
See #145 

## Motivation and Context

Someone has to do it duh

## How Has This Been Tested?
Just the playground and multiple demos using this component

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ For changelog, visit [radix-vue.com/overview/releases](https://radix-vue.com/ove
 | [Separator](https://github.com/radix-vue/radix-vue/issues/20)       |   ✅   | ✅           |                        |
 | [Slider](https://github.com/radix-vue/radix-vue/issues/21)          |   ✅   |              |                        |
 | [Switch](https://github.com/radix-vue/radix-vue/issues/22)          |   ✅   | ✅           |                        |
-| [Tabs](https://github.com/radix-vue/radix-vue/issues/23)            |   ✅   |              |                        |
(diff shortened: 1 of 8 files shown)


onmax:
I have implemented all the props/events that are described [here](https://www.radix-ui.com/docs/primitives/components/tabs) except for the `forceMount` in the `TabContent`.

I think that prop has to do something with React and therefore nothing to do in Vue?

khairulhaaziq:
there are some issues:
- [ ] When activation mode is manual, it works as expected where only when clicked the tabs are changed, however focus/active style arent showing
- [ ] On vertical navigation, need to preventDefault so it doesnt scroll. I think on horizontal need to add that aswell

Other than that, I think it's perfect!

khairulhaaziq (commented):
there are some issues:
- [ ] When activation mode is manual, it works as expected where only when clicked the tabs are changed, however focus/active style arent showing
- [ ] On vertical navigation, need to preventDefault so it doesnt scroll. I think on horizontal need to add that aswell

Other than that, I think it's perfect!
p/s: update the readme table for tabs when you're done!

onmax:
Don't merge yet. I just realised that it might be important to add `onValueChange` event. I completely forgot 🤦

khairulhaaziq:
> Don't merge yet. I just realised that it might be important to add `onValueChange` event. I completely forgot 🤦.
> 
> Or maybe we don't want it, since user can access a similar event?

no need. we're using v-model, so every props of value + onValueChange is replaced by a single v-model value

PR #190, opened 2023-07-15
Feat: MDC Conversion

DennisSmuda:
Converts mdx to mdc for documentation content files.

## Description
Currently the docs pages under '/components' were all copied from radix. What we need to do first is we need to convert from the current mdx syntax to [mdc](https://content.nuxtjs.org/guide/writing/mdc/). We have converted the Alert Dialog as an example to be applied to every other components.

## Related Issue

https://github.com/radix-vue/radix-vue/issues/172


khairulhaaziq:
hi I need to update about this issue. We arent going to convert to mdc because with vitepress we can write vue component directly. You can refer to the alertdialog or accordion for the almost-finished pages.

DennisSmuda:
Won't do mdc anymore

PR #197, opened 2023-07-16
fix: NavigationMenu recursive issue causing crash

zernonia:
Remove `setRef` causing the rendering recursive issue


khairulhaaziq (approved):
lgtm

PR #199, opened 2023-07-16
[Bug] ScrollArea showing scrollbar when the `type=hover`

zernonia:
fix the scrollbar showing when it shouldn't


khairulhaaziq (approved):
lgtm

PR #202, opened 2023-07-17
[Progress]feat: 2/3

onmax:
I completed the API from Radix as well as created multiple variants in the playground

## Description
In `ProgressRoot`:
- I created `modelValue` using `useVModel`. This value is set by the user, we watch and validate the changes (`validateValue`). By default this value is `null`, which will make the state of the progress as `indeterminate`
- I created `max` model using `useVModel`. This value is set by the user, we watch and validate the changes (`validateMax`). By default this value is `100`.
- I also created a bunch of stories that I think are relevant for the future which will help us to debug and maintain the component. There is a small issue in the story, where I use `setInterval` and for some reason is being triggered twice every tick. I do not bother to fix this issue, since is not relevant for the goal of Radix.

As we spoke in Discord, the idea in the future would be to create a custom `useVModel`, where we can pass a validator function.

## Related Issue
#139


khairulhaaziq (approved):
lgtm

Issue #131, opened 2023-07-07
[Accordion]feat: 2/3


khairulhaaziq:
Phase 2 of this project development is about making all API such as props, emits and data-attributes, work as expected. Below is the list of API needed to be completely working to get through the phase 2 of development.

The list is structured as below:
Prop/emit/data-attributes | Type | Default
### Props
- [x] asChild boolean false
- [x] type "single"|"multiple" - 
- [x] defaultValue string - 
- [ ] ~value string -~
- [ ] ~onValueChange function -~
- [x] v-model string | string[] -
- [ ] collapsible boolean false
- [ ] disabled boolean false
- [ ] dir "ltr" | "rtl" "ltr"
- [ ] orientation "horizontal"|"vertical" "vertical"

### Data Attributes
- [ ] [data-orientation] "vertical"|"horizontal"

### Props
- [ ] asChild boolean false
- [ ] disabled boolean false
- [ ] value string -

### Data Attributes
- [x] [data-state] "open"|"closed"
- [x] [data-disabled] present when disabled
- [ ] [data-orientation] "vertical"|"horizontal"

### Props
- [x] asChild boolean false

### Data Attributes
- [ ] [data-state] "open"|"closed"
- [x] [data-disabled] present when disabled
- [ ] [data-orientation] "vertical"|"horizontal"

### Props
- [x] asChild boolean false

### Data Attributes
- [x] [data-state] "open"|"closed"
- [x] [data-disabled] present when disabled
- [ ] [data-orientation] "vertical"|"horizontal"

### Props
- [x] asChild boolean false

### Content
- [ ] [data-state] "open"|"closed"
- [ ] [data-disabled] present when disabled
- [ ] [data-orientation] "vertical"|"horizontal"

PR #201, opened 2023-07-16
[chore]:  Extends PrimitiveProps to components

zernonia:
(sorry there's way too many component, wasting too much time to list them out 1 by 1 🥲)

PR #211, opened 2023-07-19
[Accordion]feat: 2/3

onmax:
## Description
Finished Accordion API and a11y

## Related Issue
#131 

## Motivation and Context
There is a major change. I removed all the `AccordionImpl` components. I don't think we need them.
The main current problem now is that as far as I know we cannot express the following logic into Vue `props`:

```ts
{
  type: AccordionType.Single;
  modelValue?: string;
} | {
  type: AccordionType.Single | AccordionType.Multiple;
  modelValue?: string[];
}
```

Basically change the type of `modelValue` depending on the type. Since we cannot do that, I just declare `modelValue`, `defaultValue` and `onValueChange` as a combination of: `string`, `undefined` and `string[]`. 

Please verify that the exports are ok.

Let me know if you see any bugs or room to improvments!

Thank you!


khairulhaaziq:
Alot of things to unpack here.
1. When typing props, you need to extend the existing props if the child is another component. E.g. if the child is CollapsilbleTrigger, import the CollapsibleTriggerProps and extend them, that way you dont need to declare asChild default to false because it is already done in the bottom component(the primitives). This is what @zernonia done [here](https://github.com/radix-vue/radix-vue/pull/201) but he probably missed some in accordion.
2. We dont do onValueChange, every `value` or `open` we bind with v-model and in components we implement useVModel. You can refer to other components on how these are implemented.
3. Thanks for the refactors, we had that in planned! Actually there is a very relevant example on how to implement `single` and `multiple` props. They are `toggle group` and `select`. I think you should refer those and see how they are implemented.
4. This is just enhancer, I think we should remove the dependancies on Collapsible components and use the Primitives right away, although these are almost the same features but removing dependancies would make it much more maintainable and cleaner.

Alot to take in, overall pretty good and very nearly ready. Let me know if you need some help!

packages/radix-vue/src/Accordion/AccordionContent.vue
```diff
@@ -1,32 +1,37 @@
...
   ACCORDION_ITEM_INJECTION_KEY
 );
+
+const props = withDefaults(defineProps<{ asChild?: boolean }>(), {
+  asChild: false,
+});
 </script>
 
 <template>
   <CollapsibleContent
```
khairulhaaziq, line 24: remove the Collapsible dependancy

packages/radix-vue/src/Accordion/AccordionItem.vue
```diff
@@ -1,180 +1,119 @@
...
-  ACCORDION_VALUE_INJECTION_KEY
-);
+const injectedRoot = inject<AccordionProvideValue>(ACCORDION_INJECTION_KEY);
 
 const open = computed({
-  get: () =>
-    (props.value &&
-      accordionValueInjectedValue?.modelValue.value.includes(props.value)) ||
-    false,
+  get: () => {
```
khairulhaaziq, line 54: we dont need computedref/setters, this can be computed only

packages/radix-vue/src/Accordion/AccordionItem.vue
```diff
@@ -1,180 +1,119 @@
...
-        movePrev();
-      }
-      break;
-  }
-
-  const clampedIndex = nextIndex % triggerCount;
-  triggerCollection[clampedIndex].focus();
-};
+  const trigger =
+    newSelectedElement.querySelector<HTMLButtonElement>("[aria-controls]");
```
khairulhaaziq, line 102: find another method to select the trigger. maybe put in inject as how other component does

packages/radix-vue/src/Accordion/AccordionHeader.vue
```diff
@@ -1,33 +1,34 @@
...
-);
-const accordionItemInjectedValue = inject<AccordionItemProvideValue>(
+const injectedRoot = inject<AccordionProvideValue>(ACCORDION_INJECTION_KEY);
+const injectedItem = inject<AccordionItemProvideValue>(
   ACCORDION_ITEM_INJECTION_KEY
 );
+
+const props = withDefaults(defineProps<AccordionHeaderProps>(), {
+  asChild: false,
+});
```
khairulhaaziq, line 23: we extended AccordionHeaderProps with PrimitiveProps, so we dont need to declare asChild default to false

khairulhaaziq (changes requested):
refer to the other comment

onmax:
Great feedback. Working on it

packages/radix-vue/src/Accordion/AccordionRoot.vue
```diff
@@ -1,34 +1,214 @@
...
+        `Invalid prop \`defaultValue\` of value \`${props.defaultValue}\` supplied to \`AccordionRoot\`, which type is \`single\`. The \`defaultValue\` prop must be:
+  - a string
+  - \`undefined\`
+
+Defaulting to \`undefined\`.`
+      );
+      return undefined;
+    }
+  }
+}
```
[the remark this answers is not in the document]
onmax: But, do you want to keep the `validate` functions?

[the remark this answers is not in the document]
khairulhaaziq: > But, do you want to keep the `validate` functions?

yeah validate is ok. look how small togglegroup is, try to follow that without losing any functionality/maintainability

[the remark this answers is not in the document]
khairulhaaziq: > An idea: what about a composable like this:
> 
> [39a27fc](https://github.com/radix-vue/radix-vue/commit/39a27fce5ce35fa387d82aba2cfe943f58424a3f)

idk how it is useful, show how it can be used accross components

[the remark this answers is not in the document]
onmax: With:

```ts
const { modelValue, changeModelValue } = useSingleOrMultipleValue(props, emits);
```

It is basically the code that we have in[ `ToggleGroupRoot`](https://github.com/onmax/radix-vue/blob/39a27fce5ce35fa387d82aba2cfe943f58424a3f/packages/radix-vue/src/ToggleGroup/ToggleGroupRoot.vue#L67)


<details><summary>And doing a quick search in VSCode, we have like 8</summary>

</details>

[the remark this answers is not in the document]
onmax: So we don't repeat the same validations and the same `changeValueModel` logic

onmax:
I think everything is ready. I will recheck everything tomorrow in case I miss something I don't see now.

I can squash the commits to clean the tree before merging

onmax:
@khairulhaaziq you mention to remove `Collapsible`. I feel like I have to copy and paste a lot of the logic.

See [Radix implementation](https://github.com/radix-ui/primitives/blob/eca6babd188df465f64f23f3584738b85dba610e/packages/react/accordion/src/Accordion.tsx#L8):

zernonia:
> @khairulhaaziq you mention to remove `Collapsible`. I feel like I have to copy and paste a lot of the logic.
> 
> See [Radix implementation](https://github.com/radix-ui/primitives/blob/eca6babd188df465f64f23f3584738b85dba610e/packages/react/accordion/src/Accordion.tsx#L8):

I dont recommend remove Collapsible. What's the issue that leads to the idea of removing ya?

khairulhaaziq:
not working for me: 

https://github.com/radix-vue/radix-vue/assets/101852870/fddfcef3-d7d9-4436-acdf-1d903729e2df

khairulhaaziq:
this has been a slow progress do to my absence, sorry. @zernonia you can take it over

packages/radix-vue/src/Accordion/story/AccordionSingle.story.vue
```diff
@@ -104,11 +104,11 @@ const selectedValue = ref("item-2");
       </AccordionRoot>
     </Variant>
 
-    <Variant title="Collapsible">
+    <Variant title="Not Collapsible">
       <AccordionRoot
         class="w-[300px] rounded-md bg-[--line-color] shadow-lg"
         type="single"
-        :collapsible="true"
+        :collapsible="false"
```
zernonia, line 111: I think let's revert this back to `Collapsible`. As the default value for `collapsible` is `false`. this Variant here allow us to test `collapsible: true`

packages/radix-vue/src/Accordion/AccordionRoot.vue
```diff
@@ -1,34 +1,107 @@
...
+  orientation: "vertical",
+  collapsible: false,
+});
 
-const emit = defineEmits<{
+const emits = defineEmits<{
   (e: "update:modelValue", value: string | string[] | undefined): void;
 }>();
+
+const { modelValue, changeModelValue } = useSingleOrMultipleValue(props, emits);
```
zernonia, line 86: I think you missed out passing `defaultValue` here.

causing the Accordion not open by default in uncontrollable state


```
    <AccordionRoot
        class="w-[300px] rounded-md bg-[--line-color] shadow-lg"
        type="single"
        default-value="item-3"
      >
```
image

packages/radix-vue/src/Accordion/AccordionRoot.vue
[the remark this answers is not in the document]
zernonia: I think this composable is good. It handles logic for single/multiple for `modelValue`, and with validation 👍🏻.

packages/radix-vue/src/Accordion/AccordionItem.vue
```diff
@@ -17,165 +27,88 @@ export interface AccordionItemProps
...
   );
-  const triggerIndex = triggerCollection.findIndex(
-    // (item) => item.ref.current === target
-    (item) => item === target
-  );
-  const triggerCount = triggerCollection.length;
+});
 
-  if (triggerIndex === -1) return;
+const dataDisabled = computed(() => (disabled.value ? "" : undefined));
```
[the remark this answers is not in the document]
onmax: No. If we return boolean, then the element will have `data-disabled="true"` or `data-disabled="false"`. What we need is to set `data-disabled` to the element or we remove it.

The Radix documentation states: "`data-disabled` is present when is disabled". [Docs](https://www.radix-ui.com/docs/primitives/components/accordion#item)

See [Radix code implementation](https://github.com/radix-ui/primitives/blob/main/packages/react/accordion/src/Accordion.tsx#L437)

packages/radix-vue/src/Accordion/story/AccordionSingle.story.vue
onmax: True, my bad

packages/radix-vue/src/Accordion/AccordionRoot.vue
onmax: I will get the value directly from the props. I think is cleaner

onmax:
@zernonia I pushed some changes. 

[This comment ](https://github.com/radix-vue/radix-vue/pull/211#discussion_r1270205779)is the missing thing now

khairulhaaziq:
@onmax since our reviews halt the development abit, do consider working on another component meanwhile. @zernonia i think can merge?

onmax:
Yeah, I forgot to create a new branch for the accordion. So now I was waiting for the merge, to start working in a new component.

Nonetheless, today I will be travelling, so I can wait 😄

zernonia:
@khairulhaaziq  @onmax  this is working well! And loves the validation messages. Clear and easily actionable!

zernonia (approved):

Issue #149, opened 2023-07-07
[Tooltip]feat: 2/3


khairulhaaziq:
Phase 2 of this project development is about making all API such as props, emits and data-attributes, work as expected. Below is the list of API needed to be completely working to get through the phase 2 of development.

The list is structured as below:
Prop/emit/data-attributes | Type | Default

### Props
- [ ] delayDuration number 700
- [ ] skipDelayDuration number 300
- [ ] disableHoverableContent boolean -

### Props
- [ ] defaultOpen boolean - 
- [ ] open boolean -
- [ ] onOpenChange function -
- [ ] delayDuration number 700
- [ ] disableHoverableContent boolean -

### Props
- [ ] asChild boolean false

### Data Attributes
- [ ] [data-state] "closed" | "delayed-open" | "instant-open"

### Props
- [ ] forceMount boolean -
- [ ] container HTMLElement document.body

### Props
- [ ] asChild boolean false
- [ ] aria-label string -
- [ ] onEscapeKeyDown function -
- [ ] onPointerDownOutside function -
- [ ] forceMount boolean -
- [ ] side "top" | "right" | "bottom" | "left" "bottom"
- [ ] sideOffset number 0
- [ ] align "start" | "center" | "end" "center"
- [ ] alignOffset number 0
- [ ] avoidCollisions boolean true
- [ ] collisionBoundary Element | null | Array<Element | null> []
- [ ] collisionPadding number|Padding 0
- [ ] arrowPadding number 0
- [ ] sticky "partial"|"always" "partial"
- [ ] hideWhenDetached boolean false

### Data Attributes
- [ ] [data-state] "open"|"closed"
- [ ] [data-side] "left"|"right"|"bottom"|"top"
- [ ] [data-align] "start"|"end"|"center"

### Props
- [ ] asChild boolean false
- [ ] width number 10
- [ ] height number 5

## Related
#28

PR #231, opened 2023-07-23
[Tooltip] Complete API

onmax:
## Description
Complete the Tooltip API.

Relevant changes:
- All props for `Tooltip*` components with comments
- I created `ScreenReaderOnly` component
- Refactor `useHoverDelay` -> `useHover`
- Removed `TooltipPortal` in favor of `Teleport`
- Added comments to `PopperContentProps`
- Radix React uses `onEscapeKeyDown` and `onPointerDownOutside` callback functions as props in `TooltipContent`. I think we decided that we instead we would emit events.

#### Regarding @focus

At some point I think we have to update the Tooltip focus event. We don't really want to open it for every event, just for events triggered by keyboard.

I opened this https://github.com/vueuse/vueuse/pull/3254 so we can use that composable or we just copy and paste that function into our project, as we want.

In Radix, for example, they just check if the event was triggered by a PointerDown, see [Radix Code](https://github.com/radix-ui/primitives/blob/main/packages/react/tooltip/src/Tooltip.tsx#L302)

## Related Issue
#149 

## How Has This Been Tested?
Visually and I created some stories. Not sure how to do propper stories, that's why I did the minimum to test the props


onmax:
So this PR is complete. Now I am facing an issue in the `stories/Tooltip.content.story.vue`.

I have no idea what is going on and how I can debug it. Any one has an idea why I get this error?
image

Going to `stories/Tooltip.story.vue` seems to work fine

onmax:
At some point I think we have to update the Tooltip `focus` event. We don't really want to open it for every event, just for events triggered by `keyboard`.

I opened this [PR in `VueUse`](https://github.com/vueuse/vueuse/pull/3254) so we can use that composable or we just copy and paste that function into our project, as we want.

In Radix, for example, they just check if the event was triggered by a `PointerDown`, see [Radix Code](https://github.com/radix-ui/primitives/blob/main/packages/react/tooltip/src/Tooltip.tsx#L302)

packages/radix-vue/src/Tooltip/TooltipContent.vue
```diff
@@ -1,38 +1,116 @@
...
-      :asChild="props.asChild"
-    >
+<PopperContent v-if="injectedValue?.open.value" ref="contentElement" :as-child="props.asChild" :side="props.side"
+    :sideOffset="props.sideOffset" :align="props.align" :alignOffset="props.alignOffset"
+    :avoidCollisions="props.avoidCollisions" :collisionBoundary="props.collisionBoundary"
+    :collisionPadding="props.collisionPadding" :arrowPadding="props.arrowPadding" :sticky="props.sticky"
+    :hideWhenDetached="props.hideWhenDetached" style="
+      --radix-tooltip-content-transform-origin: var(
+        --radix-popper-transform-origin
+      );
```
zernonia, line 100: what's the reason replacing `v-bind=props` with explicitly props ya? It kinda congested the code, and serves no different as v-bind props.

packages/radix-vue/src/HoverCard/HoverCardTrigger.vue
```diff
@@ -1,60 +1,35 @@
...
-async function handleMouseleave(e: MouseEvent) {
-  injectedValue!.isHover = false;
-  const result = await useMouseleaveDelay(e, injectedValue?.closeDelay);
-  if (result && !injectedValue?.isHover) {
-    injectedValue?.hideTooltip();
-  }
-}
+const triggerElement = ref<HTMLElement>();
+
+const isHovered = useHover(triggerElement, {
```
zernonia, line 20: this `useHover` broke `HoverCard`. I suggest let's add back `useHoverDelay`, and we can remove that once we refactor HoverCard.

packages/radix-vue/src/Tooltip/TooltipTrigger.vue
```diff
@@ -4,55 +4,35 @@ export type TooltipTriggerDataState =
...
-      :data-state="dataState"
-      :aria-expanded="injectedValue?.open.value || false"
-      :as-child="props.asChild"
-      @mouseenter="handleMouseEnter"
-      @mouseleave="injectedValue?.hideTooltip"
-      @focus="injectedValue?.showTooltip"
-      @blur="injectedValue?.hideTooltip"
-      style="cursor: default"
-    >
+  <PopperAnchor ref="triggerElement" :as-child="props.asChild">
```
zernonia, line 31: this is meant to be `asChild="true"`  always

packages/radix-vue/src/Tooltip/TooltipContent.vue
onmax: Basically to avoid `props` drilling; Avoid passing to `PopperContent` other variables that they don't use like `ariaLabel`.
For sure, I would prefer using `v-bind` as it is concise, but I just did this to be safe

packages/radix-vue/src/Tooltip/TooltipContent.vue
```diff
@@ -1,38 +1,116 @@
...
+      --radix-tooltip-content-available-height: var(
+        --radix-popper-available-height
+      );
+      --radix-tooltip-trigger-width: var(--radix-popper-anchor-width);
+      --radix-tooltip-trigger-height: var(--radix-popper-anchor-height);
+    " @keydown.esc="onEscapeKeyDown($event)">
+    <PrimitiveDiv :data-state="injectedValue?.dataState.value" :data-side="props.side" :data-align="props.align"
+      role="tooltip" tabindex="-1">
       <slot />
+      <ScreenReaderOnly :id="injectedValue?.contentId" :ariaLabel="ariaLabel" />
```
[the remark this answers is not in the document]
onmax: Just updated the code, I had to update `VisuallyHidden` component with a new slot.

In `NavigationMenuTrigger`, we have the following code:
```vue
<template v-if="open">
  <VisuallyHidden
    aria-hidden
    :tabIndex="0"
    :ref="setFocusProxyRef"
    @focus="handleVisuallyHiddenFocus"
  >
  </VisuallyHidden>
  <span :aria-owns="contentId" v-if="context?.viewport"></span>
</template>
```

But I don't fully understand this code. I would appreciate any feedback of [my current implementation](https://github.com/radix-vue/radix-vue/pull/231/commits/e450c4518b68752a308afd75d778e11c8add813d#diff-8ebfb3d10d04967175c21e45f46923067aca5719ae03b6cc6152c36a70b38344R110-R112), which imitates [Radix implementation.](https://github.com/radix-ui/primitives/blob/main/packages/react/tooltip/src/Tooltip.tsx#L553-L555)

zernonia (approved):
lgtm!

PR #237, opened 2023-07-23
fix: bind props values to data-side and data-align, remove unnecessar…

MellKam:
I think someone forgot to change static values for data-side and data-align attributes


khairulhaaziq (approved):
lgtm, thanks!

PR #241, opened 2023-07-24
Make the Histoire component center

### LINKED ISSUE #239: Title Tooltip trigger position not visible on the Histoire page
When I'm trying to fix the tooltip component, the tooltip trigger position is in the top left corner. In that case, the tooltip content is not visible. 

Screenshot 2023-07-24 at 06 23 06

It's also a problem with most of the components. I think it would be better if the components were displayed in the center of the page.

I would be happy to help with this issue if you think it's a good idea.

misbahansori:
Closes #239 

The commit looks like a lot of changes, but basically, I just added this wrapper div to make the component center. 

```html
<div class="min-h-screen w-full flex items-center justify-center">
```
I have added this wrapper to AlertDialog, Avatar, and Checkbox components, if you think this is good, I will add this wrapper to other components as well.


zernonia:
Let's not do this.
As I've suggested in Discord, it is better to look at how we implement Accordion's story, which is using `type:grid`.

The reason is we are going to populate more story in the future, which your approach above would be eventually replace. I wouldn't want your effort to go to waste, could you try changing to use `type: grid` for the layout? 😁


image

misbahansori:
Okay I will do that instead

Issue #226, opened 2023-07-21
[Tooltip]: asChild doesn't keep all props


steakscience:
**Describe the bug**
Related to #224
When using asChild with Tooltip, some props are not kept

**To Reproduce**
Use slots in tooltip trigger
```html
<TooltipTrigger asChild>
  <slot />
</TooltipTrigger>
```

I have props on my Button element
```html
<Tooltip>
  <Button type="tertiary" iconOnly>
    ...
  </Button>
</Tooltip>
```

**Result:**
<Button> loses the `type="tertiary"` prop, but keeps `iconOnly` boolean prop. Not sure why.

**Expected result:**
It should keep all props

@zernonia

zernonia:
Hi @steakscience .. how do you know that the props are not kept ya? Does it not appear in your component?
But if you mean you check the element html via the console, it is missing, that is a [correct behavior](https://vuejs.org/guide/components/attrs.html#nested-component-inheritance).

steakscience:
@zernonia Good point. 

In my case, I have class names that will be added if `type="tertiary"` but they are missing.

However, class names from `iconOnly` are there

khairulhaaziq:
> @zernonia Good point. 
> 
> 
> 
> In my case, I have class names that will be added if `type="tertiary"` but they are missing.
> 
> 
> 
> However, class names from `iconOnly` are there

Can you share your logic for the conditional classes for types props?

zernonia:
Can you do a console.log for "prop.type"? I wanna see if the component recevied this props

steakscience:
@khairulhaaziq Sure, it's the same as the [class bindings in the Vue docs](https://vuejs.org/guide/essentials/class-and-style.html#binding-html-classes)
<img width="910" alt="image" src="https://github.com/radix-vue/radix-vue/assets/2727095/582f7f43-3794-4adc-8f3d-f0b46ceadee3">

steakscience:
@zernonia Looks like it received the props? But also `button` isn't a prop
<img width="1391" alt="image" src="https://github.com/radix-vue/radix-vue/assets/2727095/51de0b24-2f61-4a0e-af76-9eb9509ef69e">

zernonia:
Hmmm.. icic.. cause the `TooltipTrigger` is having a type 'button' here, which is why it cause the bug.
I would need to overwrite the type if the child component (in this case your Button) has a prop type.

Thanks for the info @steakscience 

[...]

PR #242, opened 2023-07-24
[Bug]: fix props not inheritted correctly

zernonia:
Fixes https://github.com/radix-vue/radix-vue/issues/226

Added test to check
- asChild keep the props, if the props/attrs exist then replace the one inheritted.
- ... and previous bug


khairulhaaziq (approved):
looks great to me

Issue #146, opened 2023-07-07
[Toggle]feat: 2/3


khairulhaaziq:
Phase 2 of this project development is about making all API such as props, emits and data-attributes, work as expected. Below is the list of API needed to be completely working to get through the phase 2 of development.

The list is structured as below:
Prop/emit/data-attributes | Type | Default
### Props
- [x] asChild boolean false
- [x] defaultPressed boolean - 
- [x] pressed boolean -
- [ ] onPressedChange function -
- [x] disabled boolean false

### Data Attributes
- [x] [data-state] "on"|"off"
- [x] [data-disabled] present when disabled

## Related
#24

DennisSmuda:
@khairulhaaziq - why is this closed? 😅 
I'm almost done, just have to change `modelValue` to `pressed` 🙏

khairulhaaziq:
> @khairulhaaziq - why is this closed? 😅 I'm almost done, just have to change `modelValue` to `pressed` 🙏

sorry I closed a bunch of issues and overlook it, please continue 🙏

PR #243, opened 2023-07-24
Feat/toggle root component

DennisSmuda:
## Description
- [x] Refactor code to match  style
- [ ] Props 
- [x] Data Attributes (on/off)
- [ ] Data - Disabled

## Related Issue
https://github.com/radix-vue/radix-vue/issues/146

### DIFF
--- a/packages/radix-vue/src/Toggle/Toggle.story.vue
+++ b/packages/radix-vue/src/Toggle/Toggle.story.vue
@@ -1,10 +1,11 @@
 <script setup lang="ts">
-import { Toggle } from "./";
 import { Icon } from "@iconify/vue";
 import { ref } from "vue";
+import { Toggle, ToggleRoot } from "./";
 
-const toggleState = ref(true);
+const toggleState = ref(false);
 </script>
+
 <template>
   <Story title="Toggle" :layout="{ type: 'single', iframe: true }">
     <Variant title="default">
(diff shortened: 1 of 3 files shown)


packages/radix-vue/src/Toggle/index.ts
DennisSmuda, line 1: @khairulhaaziq - I would just delete the old `Toggle.vue` and export the new `ToggleRoot as Toggle` in the end 👍

packages/radix-vue/src/Toggle/ToggleRoot.vue
```diff
@@ -0,0 +1,78 @@
...
+      @change="togglePressed"
+      :checked="pressed"
+      aria-hidden="true"
+      :disabled="props.disabled"
+      :data-state="dataState"
+      :data-disabled="props.disabled"
+      style="opacity: 0; position: absolute; inset: 0"
+    />
+    <slot />
+  </PrimitiveDiv>
```
zernonia, line 77: @DennisSmuda  this section is kinda different from [RadixUI's Toggle](https://github.com/radix-ui/primitives/blob/main/packages/react/toggle/src/Toggle.tsx#L43), can we refactor them to `PrimitiveButton`, and remove the `input`? This way we have only 1 root, and `asChild` props would work 😁

DennisSmuda: Yep, I was asking myself the same, thank you :+1:

DennisSmuda:
@zernonia, @khairulhaaziq - I think I'm done 😄

packages/radix-vue/src/Toggle/Toggle.story.vue
zernonia, line 1: Can we have a Toggle/Chromatic story? 😁

https://github.com/radix-ui/primitives/blob/eca6babd188df465f64f23f3584738b85dba610e/packages/react/toggle/src/Toggle.stories.tsx

zernonia: It's ok. we can have mulitple story. I would love to at least get the "chromatic" in for every components for phase (2/3) 😁

zernonia (approved):
lgtm!

Issue #229, opened 2023-07-22
[Vue warn]: injection "Symbol()" not found.


Gerbuuun:
**Describe the bug**
```html
<template>
  <DropdownMenuRoot>
    <DropdownMenuTrigger>
      <Icon name="radix-icons:hamburger-menu" />
    </DropdownMenuTrigger>

    <DropdownMenuPortal>
      <DropdownMenuContent>
        <DropdownMenuItem>Test</DropdownMenuItem>
      </DropdownMenuContent>
    </DropdownMenuPortal>
  </DropdownMenuRoot>
</template>

<script setup lang="ts">
import {
  DropdownMenuRoot,
  DropdownMenuTrigger,
  DropdownMenuPortal,
  DropdownMenuContent,
  DropdownMenuItem,
} from "radix-vue";
</script>
```
When I use this component in my Nuxt v3.6.5 project, I get the following warning
```bash
[Vue warn]: injection "Symbol()" not found.
```
Github copilot gave the following explanation:
```
This warning is usually caused by a mismatch between the version of Vue used by your application and the version of a Vue library or plugin that you are using. The Symbol() injection is used by Vue to provide a unique identifier for each component instance, and if it is not found, it can cause errors or unexpected behavior.

To resolve this issue, you can try the following steps:

1. Make sure that you are using compatible versions of Vue and any Vue libraries or plugins that you are using. Check the documentation for each library or plugin to see which versions of Vue they are compatible with.

2. If you are using Vue 3, make sure that any Vue libraries or plugins that you are using are also compatible with Vue 3. Some libraries and plugins may not yet be updated to work with Vue 3.

3. If you are using a build tool like Webpack or Rollup, make sure that you are correctly configuring the build to include all necessary dependencies.

4. If you are using a library or plugin that provides its own version of Vue, make sure that you are not including multiple versions of Vue in your application. This can cause conflicts and errors.
```

**Desktop (please complete the following information):**
 - OS: MacOS Ventura 13.4.1c
 - Browser: Brave
 - Version 1.52.130 Chromium: 114.0.5735.198

smastrom:
@khairulhaaziq @uyloal Hi guys, I can submit a PR that fixes those SSR warnings anywhere needed as I have already dealt with this issue.

The reason why this is happening is that on the server, `inject` (imported from `vue`) simply doesn't work as it would in a vanilla Vue setup. Nuxt executes the setup function twice: once on the server and once on the client.

[...]

PR #246, opened 2023-07-25
[DropdownMenu]fix: [Vue warn]: injection "Symbol()" not found.

uyloal:
When open `DropdownMenu`, console will logging `[Vue warn]: injection "Symbol()" not found.`

## Description

The reason is that `DropdownMenuItem` and `DropdownMenuSub` will try to find the `DropdownMenuSubProvideValue`, even parent component  didn't provide it. And I make `DropdownMenuSubProvideValue` to be `undefined` in `DropdownMenuRoot`
## Related Issue

- #229

## Motivation and Context

Remove the warning, make code better than better.

## How Has This Been Tested?

- It tested with stories, no warning again.
- No error or warning about the code of `DropdownMenu`, when run build command.

#### Console logging
image

## Beside The Point

In the process of modifying the problem, I notice that the hook called `useCollection` also cause the same warning.
I didn't change code about it, becasue too much code involved. Probably shouldn't be modified with `DropdownMenu`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ For changelog, visit [radix-vue.com/overview/releases](https://radix-vue.com/ove
 | [Toggle](https://radix-vue.com/components/toggle)                   |   ✅   | ✅           |                        |
 | [Toggle Group](https://radix-vue.com/components/toggle-group)       |   ✅   | ✅           |                        |
 | [Toolbar](https://radix-vue.com/components/toolbar)                 |   ✅   |              |                        |
-| [Tooltip](https://radix-vue.com/components/tooltip)                 |   ✅   |              |                        |
+| [Tooltip](https://radix-vue.com/components/tooltip)                 |   ✅   | ✅           |                        |
 
 ## Contributing
 
--- a/packages/radix-vue/package.json
+++ b/packages/radix-vue/package.json
@@ -1,6 +1,6 @@
 {
   "name": "radix-vue",
-  "version": "0.1.11",
+  "version": "0.1.12",
   "description": "Vue port for Radix UI Primitives.",
   "type": "module",
   "main": "./dist/index.umd.cjs",


khairulhaaziq (changes requested):
this seems wrong. why would the root need to import from its sub and make another provide? I think there is another solution

uyloal:
@khairulhaaziq That's not quite right. At the time of the first modification, I create a component named `DropdownMenuSubItem`. The following code was removed from `DropdownMenuItem` and added to `DropdownMenuSubItem`. This could solve the warnings from `DropdownMenuItem`.

```ts

const subInjectedValue = inject<DropdownMenuSubProvideValue>(
  DROPDOWN_MENU_SUB_INJECTION_KEY
);
```

But warnings from `DropdownMenuSub` are much more problematic. It looks like always need to find inject value, and Vue not allow to find if value did not provide. Do you have a better idea? Maybe needs a root component to wrap `DropdownMenuSub`

```ts
const parentContext = inject(DROPDOWN_MENU_SUB_INJECTION_KEY);

provide<DropdownMenuSubProvideValue>(DROPDOWN_MENU_SUB_INJECTION_KEY, {
  modelValue,
  showTooltip: () => {
    modelValue.value = true;
  },
  hideTooltip: () => {
    modelValue.value = false;
  },
  triggerElement,
  itemsArray: [],
  orientation: props.orientation,
  triggerId: useId(),
  contentId: useId(),
  parentContext,
});
```

khairulhaaziq:
> After inspecting this deeply, This is a warning when `inject` doesn't find the associated `provided` value. This warning is expected as some of our component is mounting `recursively`, namely "DropdownMenuItem".
> 
> Although @uyloal suggestion does eliminate the warning, however it does not make sense to provide that in the root element.
> 
> Furthermore, this warning doesn't cause any effect after build.
> 
> My suggestion, let's toggle of this warning (we can toggle it on during dev if we need it), since we already have most of the skeleton of components, this warning doesn't concern much 😁
> 
> wydt @khairulhaaziq ?

agreed, yeah if we cant find an immediate solution and people are bothered by the warning, just hide it 😆

Issue #248, opened 2023-07-25
Move prettier to root workspace


Elliot-Alexander:
**Is your feature request related to a problem? Please describe.**
Inconsistent code style across packages/docs/playgrounds etc in repo

**Describe the solution you'd like**
Move the prettier installation from the radix-vue package to the root workspace. Additionally, include docs and playgrounds as workspaces.

**Describe alternatives you've considered**
Alternative is to main the formatting in each package/workspace within project - however this leads to potentially inconsistent code style between them and duplicated configurations.

**Additional context**
N/A

PR #249, opened 2023-07-25
Move prettier to root 

Elliot-Alexander:
## Description
Moves the prettier installation from the package level to the root of the workspace. It also moves to having all playgrounds and docs as workspaces. Currently, for prettier, I have scoped it to just radix-vue however this can be changed to be run across all code files. 

## Related Issue

#248

## Motivation and Context

#248

## How Has This Been Tested?

Ran CI in personal fork, ran prettier format from top level to see impacted files (did not leak out of scope)


zernonia:
we should move eslint into the root as well

Elliot-Alexander:
Yeah definitely, I can update in a bit :D

khairulhaaziq:
> Yeah definitely, I can update in a bit :D

I think [antfu's config](https://antfu.me/posts/why-not-prettier) is sufficient for our use case.

zernonia:
> I think [antfu's config](https://antfu.me/posts/why-not-prettier) is sufficient for our use case.

Yup agree. We can only use eslint, and remove prettier entirely

zernonia:
Thanks for the PR @Elliot-Alexander . With the help of @mukundshah  we already add `antfu/eslint-config` to root, and resolves many complications with that. 

Thus I'm closing this ticket ya 😁

Issue #252, opened 2023-07-26
[Vue warn] Component emitted an undefined event.


uyloal:
**Describe the bug**

`defineEmits` has a different event name with the key of `useVModel`.

`useVModel` will emit two events: `update:modelValue` and `update:open`

```ts
...

const emit = defineEmits<{
  (e: "update:modelValue", value: boolean): void;
}>();

const open = useVModel(props, "open", emit, {
  defaultValue: props.defaultOpen,
  passive: true,

...
```

Included components:
- AlertDialogRoot
- DialogRoot

**Screenshots**
![image](https://github.com/radix-vue/radix-vue/assets/37807381/06dcab5b-63e8-4dc9-a49b-1d5acf0b17a1)

**Desktop (please complete the following information):**
 - Device: desktop
 - OS: macOS 13.4
 - Browser [chrome]
 - Version [latest]

khairulhaaziq:
hi do you want to work on this issue? idk what's exactly the issue

uyloal:
@khairulhaaziq Yes, I have created a pr #253

PR #253, opened 2023-07-26
fix: component emitted an undefined event.

uyloal:
The following warning will appear when the `key` of `useVModel`'s parameters is not defined in `defineEmits`.

```console
[Vue warn]: Component emitted event "xxxx" but it is neither declared in the emits option nor as an "xxxx" prop.
```

## Description

1. update the event name in `defineEmits` to the same as the `key` of `useVModel`'s parameters. `AlertDialogRoot`,`DialogRoot`
2. remove redundant event names. `MenubarRoot`

## Related Issue

- #252 

## Motivation and Context

make a better DX

## How Has This Been Tested?

Tested it with histoire.


khairulhaaziq:
thanks!

Issue #258, opened 2023-07-26
Tooltip: Trigger overrides cursor styles


steakscience:
**Describe the bug**
`<TooltipTrigger>` has an inline style `cursor: default`

This should not be included and will override any styles from child elements

khairulhaaziq:
hi yeah sorry about that, do you want to create a pr and fix the issue?

khairulhaaziq:
> @khairulhaaziq Got it

thanks so much! haha I love getting more people involved!

PR #262, opened 2023-07-26
fix(tooltip): Remove unnecessary inline style

steakscience:
## Description
Removes unnecessary inline style (cursor: default) in TooltipTrigger

## Related Issue
#258

## Motivation and Context
Prevents cursor styles from child elements from being overridden


khairulhaaziq:
thanks!!!

khairulhaaziq (approved):
lgtm

PR #264, opened 2023-07-27
[useSingleOrMultiple] fix: Toogle value

onmax:
## Description
Rather that activating always the single element, we toggle the element `on` and `off`


zernonia (approved):
lgtm!

PR #268, opened 2023-07-27
[AlertDialog] fix possible errors with watchEffect, implement radix-ui accesibility attirbutes that depends on element id ...

MellKam:
1. Move cleanup logic from watchEffect to onCleanup hook. 
2. Changed role for DialogContent from "dialog" to "alertdialog". (to match radix-ui)
3. Add id attributes to title, description and dialog content to implement aria-labelledby, and other accasibility things. Now should be the same as in radix
4. Removed unnecessary generic casting for inject/provide values.

### DIFF
--- a/packages/radix-vue/src/AlertDialog/AlertDialogAction.vue
+++ b/packages/radix-vue/src/AlertDialog/AlertDialogAction.vue
@@ -1,25 +1,20 @@
 <script setup lang="ts">
 import { inject } from "vue";
 import { PrimitiveButton, type PrimitiveProps } from "@/Primitive";
-import {
-  ALERT_DIALOG_INJECTION_KEY,
-  type AlertDialogProvideValue,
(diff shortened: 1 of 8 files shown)


khairulhaaziq:
Hi thanks for the work! great job! since dialog features are similar to alertdialog, would you mind refactoring that too? to keep it consistent

MellKam:
> Hi thanks for the work! great job! since dialog features are similar to alertdialog, would you mind refactoring that too? to keep it consistent

Okay, I can do that. I think it's better to do it in a separate pullrequest?

khairulhaaziq:
> > Hi thanks for the work! great job! since dialog features are similar to alertdialog, would you mind refactoring that too? to keep it consistent
> 
> Okay, I can do that. I think it's better to do it in a separate pullrequest?

thanks!

khairulhaaziq (approved):

Issue #143, opened 2023-07-07
[Slider]feat: 2/3


khairulhaaziq:
Phase 2 of this project development is about making all API such as props, emits and data-attributes, work as expected. Below is the list of API needed to be completely working to get through the phase 2 of development.

The list is structured as below:
Prop/emit/data-attributes | Type | Default
### Props
- [ ] asChild boolean false
- [ ] defaultValue number[] - 
- [ ] value number[] -
- [ ] onValueChange function -
- [ ] onValueCommit function -
- [ ] name string -
- [ ] disabled boolean false
- [ ] orientation "horizontal"|"vertical" "vertical"
- [ ] dir "ltr" | "rtl" -
- [ ] inverted boolean false
- [ ] min number 0
- [ ] max number 100
- [ ] step number 1
- [ ] minStepsBetweenThumbs number 0

### Data Attributes
- [ ] [data-disabled] present when disabled
- [ ] [data-orientation] "vertical"|"horizontal"

### Props
- [ ] asChild boolean false

### Data Attributes
- [ ] [data-disabled] present when disabled
- [ ] [data-orientation] "vertical"|"horizontal"

### Props
- [ ] asChild boolean false

### Data Attributes
- [ ] [data-disabled] present when disabled
- [ ] [data-orientation] "vertical"|"horizontal"

### Props
- [ ] asChild boolean false

### Data Attributes
- [ ] [data-disabled] present when disabled
- [ ] [data-orientation] "vertical"|"horizontal"

## Related
#21

Issue #21, opened 2023-06-23
Feat: Slider


khairulhaaziq:
[reference](https://www.radix-ui.com/docs/primitives)

## API Reference
### Root

Contains all the parts of a slider. It will render an `input` for each thumb when used within a `form` to ensure events propagate correctly.

| Name                   | Required | Type                                            | Default | Description                                                                                                                                                                                                                                                                                                                                   | Status |
|------------------------|----------|-------------------------------------------------|---------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|--------|
| asChild                | false    | boolean                                         | false   | Change the default rendered element for the one passed as a child, merging their props and behavior. Read our [Composition](../guides/composition) guide for more details.                                                                                                                                                                 | ❌      |
| defaultValue           |          | number[]                                        |         | The value of the slider when initially rendered. Use when you do not need to control the state of the slider.                                                                                                                                                                                                                                 | ❌      |
| value                  |          | number[]                                        |         | The controlled value of the slider. Must be used in conjunction with `onValueChange`.                                                                                                                                                                                                                                                         | ❌      |
| onValueChange          |          | onValueChange?(value: number[]): void           |         | Event handler called when the value changes.                                                                                                                                                                                                                                                                                                  | ❌      |
| onValueCommit          |          | onValueCommit?(value: number[]): void           |         | Event handler called when the value changes at the end of an interaction. Useful when you only need to capture a final value e.g. to update a backend service.                                                                                                                                                                              | ❌      |
| name                   |          | string                                          |         | The name of the slider. Submitted with its owning form as part of a name/value pair.                                                                                                                                                                                                                                                         | ❌      |
| disabled               |          | boolean                                         | false   | When `true`, prevents the user from interacting with the slider.                                                                                                                                                                                                                                                                              | ❌      |
| orientation            |          | "horizontal" \| "vertical"                       | "horizontal" | The orientation of the slider.                                                                                                                                                                                                                                                                                                               | ❌      |
| dir                    |          | "ltr" \| "rtl"                                  |         | The reading direction of the slider. If omitted, inherits globally from `DirectionProvider` or assumes LTR (left-to-right) reading mode.                                                                                                                                                                                                     | ❌      |
| inverted               |          | boolean                                         | false   | Whether the slider is visually inverted.                                                                                                                                                                                                                                                                                                     | ❌      |
[...]

PR #272, opened 2023-07-28
[Slider]: feat: 2/3

edtorba:
### Props
- [x]  asChild boolean false
- [x]  defaultValue number[]
- [x]  onValueCommit function
- [x]  name string
- [x]  disabled boolean false
- [x]  orientation "horizontal"|"vertical"
- [x]  dir "ltr" | "rtl"
- [x]  inverted boolean false
- [x]  min number 0
- [x]  max number 100
- [x]  step number 1
- [x]  minStepsBetweenThumbs number 0

### Data Attributes
- [x]  [data-disabled] present when disabled
- [x]  [data-orientation] "vertical"|"horizontal"
### Props
- [x]  asChild boolean false
### Data Attributes
- [x] [data-disabled] present when disabled
- [x] [data-orientation] "vertical"|"horizontal"
### Props
- [x] asChild boolean false
### Data Attributes
- [x]  [data-disabled] present when disabled
- [x]  [data-orientation] "vertical"|"horizontal"
### Props
- [x]  asChild boolean false
### Data Attributes
- [x]  [data-disabled] present when disabled
- [x]  [data-orientation] "vertical"|"horizontal"

## Related
https://github.com/radix-vue/radix-vue/issues/21 https://github.com/radix-vue/radix-vue/issues/143 https://github.com/radix-vue/radix-vue/issues/288


edtorba:
A minor update, I've been busy with other unrelated stuff. Should be able to continue work on this early next week 👍

edtorba:
The "inverted" implementation is actually erroneous. I mistakenly implemented it differently from what the "dir" should be. In essence, "inverted" is intended to merely reverse the "dir" value. For instance, when "dir" is set to "rlt" and "inverted" is set to "true," the outcome should be "ltr."

edtorba:
@khairulhaaziq @zernonia Taking into account that we mimic radix-ui, but still, should the `value` prop be dropped and `modelValue` (already there) used instead? As technically that would be a more appropriate property name for the Vue land?

zernonia:
> @khairulhaaziq @zernonia Taking into account that we mimic radix-ui, but still, should the `value` prop be dropped and `modelValue` (already there) used instead? As technically that would be a more appropriate property name for the Vue land?

Hey @edtorba .. yes let's drop the `value`, and use `modelValue` 😁

packages/radix-vue/src/Slider/SliderThumb.vue
```diff
@@ -25,68 +26,84 @@ const props = withDefaults(defineProps<SliderThumbProps>(), {
...
+
+  return style;
+});
 </script>
 
 <template>
-  <span
-    :style="`transform: translateX(-50%); position: absolute; left: calc(${injectedValue?.modelValue?.value}%)`"
-  >
+  <span :style="thumbStyle">
```
razvantimis, line 106: can you add inheritAttrs false because we duplicate attrs on both span
```
defineOptions({
  inheritAttrs: false
})
```

edtorba:
@zernonia @razvantimis I haven't wired the "minStepsBetweenThumbs" prop because I don't fully understand it. Feel free to push a commit or provide a hint about what needs to be done regarding it. As for the other pieces, please feel free to go through them. 👀

zernonia:
@edtorba  could you add the Chromatic story for Slider (https://github.com/radix-ui/primitives/blob/main/packages/react/slider/src/Slider.stories.tsx#L323). This would help to check that the Slider is working correctly.

Also, I would expect this should fix https://github.com/radix-vue/radix-vue/issues/288  too. However in your branch, if I set `:max="1000"` it's still behaving weirdly.

edtorba:
Had max prop working, maybe accidentally broke it, will check 👍

edtorba:
Right... that PR is defo not ready... Somehow "Supports multiple thumbs." part went over my head... 🤦 I'll add chromatic story but that'll be missing muti thumb for now ...

edtorba:
> minStepsBetweenThumbs

I now understand what it's meant to do as well 😂🤣👍

edtorba:
I'll have it fully ready in next couple of days. And sorry for taking time 😅😂

zernonia (commented):
1. The activeIndex seems like not targeting current thumb.

https://github.com/radix-vue/radix-vue/assets/59365435/15727270-bdfe-4e22-a76c-c7865f4dc53a

2. The positioning of thumb is wrong sometimes

https://github.com/radix-vue/radix-vue/assets/59365435/89eefe15-6ef2-4f01-869d-b8f8a45af9db

3. Controlled slider feels clunky. Might be something to do with `setValue` ]

https://github.com/radix-vue/radix-vue/assets/59365435/bd2e84c0-f0df-497a-af9a-099a305fc208

zernonia (approved):

edtorba:
Am away for a few days, will work on perf and other minor glitches when will be back 👀👍

Issue #273, opened 2023-07-28
[AlertDialog] prevent body from content shift when overflow is hidden


MellKam:
I think `useScrollLock` hook from vueuse doesn't fit for our purpose, because it sets overflow to "hidden" and causes content shift.

https://github.com/radix-vue/radix-vue/blob/24124e7f43b6dc04b01be4946cefa2ad3a70d8b2/packages/radix-vue/src/AlertDialog/AlertDialogRoot.vue#L44-L48

In original radix alert dialog they just use `pointer-events: none;`. My suggestion:

```ts
watch(open, (open) => {
  if (open) {
    document.body.style.pointerEvents = "none";
  } else {
  	document.body.style.pointerEvents = null;
  }
})
```

khairulhaaziq:
you were the one removed it hahah. yeah just create a PR and propose your solution

MellKam:
> you were the one removed it hahah. yeah just create a PR and propose your solution

Sorr of, I just merged the main branch with these changes and thought that it worked the same as before. XDDD

khairulhaaziq:
> > you were the one removed it hahah. yeah just create a PR and propose your solution
> 
> Sorr of, I just merged the main branch with these changes and thought that it worked the same as before. XDDD

create a pr for this, if it takes too long I might revert first

PR #274, opened 2023-07-29
[AlertDialog] fix: remove the useScrollLock hook and toggle body pointer events instead

MellKam:
#273

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -150,6 +150,52 @@ importers:
         specifier: ^1.8.1
         version: 1.8.1(typescript@5.1.3)
 
+  playground/vue3:
+    dependencies:
+      radix-vue:
+        specifier: link:../../packages/radix-vue
+        version: link:../../packages/radix-vue
+      vue:
+        specifier: ^3.2.47
+        version: 3.3.4
+    devDependencies:
+      '@antfu/eslint-config':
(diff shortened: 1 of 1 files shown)


khairulhaaziq (approved):

khairulhaaziq:
so where's the scroll lock function? nvm Im reverting for now.

MellKam:
Are you talking about this piece of code? 

```
watch(open, (open) => {
  if (open) {
    document.body.style.pointerEvents = "none";
  } else {
  	document.body.style.pointerEvents = null;
  }
})
```

I thought that we already have place to put it in AlertContent and I did so.

khairulhaaziq:
> Are you talking about this piece of code?
> 
> ```
> watch(open, (open) => {
>   if (open) {
>     document.body.style.pointerEvents = "none";
>   } else {
>   	document.body.style.pointerEvents = null;
>   }
> })
> ```
> 
> I thought that we already have place to put it in AlertContent and I did so.

you understand we need to lock scroll for accessibility right?

MellKam:
> you understand we need to lock scroll for accessibility right?

Oh, sorry, my bad. I just realized that pointer-events does not stop the scroll). Checked the radix sources and will try to reproduce it a bit later. They actally use different package to do this: "react-remove-scroll".

PR #276, opened 2023-07-29
[Bug]: ScrollArea unable to scroll

zernonia:
Turns out using inline styling doesn't causes some weird bug of not scrolling. Moved back to use Vue SFC's style tag


zernonia:
@khairulhaaziq I trhink you have something configured in the runner that has been causing (previously fine) build to fail 😅

PR #277, opened 2023-07-30
fix: add margin-right to body to prevent content shift

MellKam:
#274 
#273

### DIFF
--- a/packages/radix-vue/src/AlertDialog/AlertDialogRoot.vue
+++ b/packages/radix-vue/src/AlertDialog/AlertDialogRoot.vue
@@ -41,9 +41,22 @@ const open = useVModel(props, "open", emit, {
   passive: true,
 });
 
-const locked = useScrollLock(document.querySelector("body"), open.value);
+const locked = useScrollLock(document.body, open.value);
 
 watchEffect(() => {
+  if (open.value) {
+    const verticalScrollbarWidth =
+      window.innerWidth - document.body.clientWidth;
+
+    if (verticalScrollbarWidth) {
+      document.body.style.marginRight = verticalScrollbarWidth + "px";
+    }
+
+    document.body.style.pointerEvents = "none";
+  } else {
+    document.body.style.marginRight = "";
+    document.body.style.pointerEvents = "";
+  }
   locked.value = open.value;
 });
 
--- a/packages/radix-vue/src/AlertDialog/AlertDialogRoot.vue
+++ b/packages/radix-vue/src/AlertDialog/AlertDialogRoot.vue
@@ -43,21 +43,28 @@ const open = useVModel(props, "open", emit, {
 
 const locked = useScrollLock(document.body, open.value);
 
-watchEffect(() => {
+const cleanupBodyLock = () => {
+  document.body.style.marginRight = "";
+  document.body.style.pointerEvents = "";
+  locked.value = false;
+};
+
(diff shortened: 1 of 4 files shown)


packages/radix-vue/src/AlertDialog/AlertDialogRoot.vue
```diff
@@ -41,10 +41,30 @@ const open = useVModel(props, "open", emit, {
...
+    }
+
+    document.body.style.pointerEvents = "none";
+    locked.value = true;
+  } else {
+    cleanupBodyLock();
+  }
+
+  onCleanup(cleanupBodyLock);
 });
```
sadeghbarati, line 68: https://github.com/twbs/bootstrap/blob/main/js/src/util/scrollbar.js#L33C1-L34C55

I think bootstrap approach is better, also use `padding` instead of `margin`

MellKam: I used margin to match original radix-ui. I dont think that it makes much difference.

zernonia (commented):
lgtm! however can we add this to `DialogRoot` as well?

MellKam:
Done. If you want, we can move this logic to some composable.

zernonia:
> Done. If you want, we can move this logic to some composable.

Im ok if you want to wrap it in a composable 😁

zernonia (approved):

zernonia:
@MellKam  I just saw that using margin-right doesn't solve the content shift on https://www.radix-vue.com/components/alert-dialog.html. However, using padding-right does solves it.

MellKam:
> @MellKam I just saw that using `margin-right` doesn't solve the content shift on [radix-vue.com/components/alert-dialog.html](https://www.radix-vue.com/components/alert-dialog.html). However, using `padding-right` does solves it.
> 
> Could you investigate, and perhaps we can use `padding-right` instead?

Interesting. The original radix-ui is also done via margin and it worked fine for me. I'll take a look at it. 

Maybe the problem is additional styles that I haven't added. (position and scroll-behavior) 
image

zernonia:
Sorry I did a quick test using `paddingRight` and deployed, yet it seems to have another weird content shifting on Radix-Vue website 🙈

MellKam:
Personally, I don't know what to do about this problem. Some libraries do hacks and add padding to fixed elements to prevent this offset, but that doesn't seem right to me. I think it's easier to fix the layout styles than to fix this problem.

sadeghbarati:
@ zernonia 

> Tested on Windows 11

I have no problem with the main content in `vitepress` but the header and sidebar have layout shifts,
which also happens on header in [radix ui](https://www.radix-ui.com/primitives/docs/components/alert-dialog) too

I prefer padding for this, cause the CSS Box Model order is 

1. content
2. padding
3. border
4. margin

---

We can add another CSS variable with js (width of scrollbar), so the user can prevent layout shift on fixed elements

Don't take this one seriously 😄 👇 
Or recommend users to use `scrollbar-gutter` and check `scrollbar-gutter` support with js `CSS.supports("scrollbar-gutter:stable")` to prevent adding padding to body in useBodyScrollLock composable

MellKam:
@sadeghbarati I've tested `scrollbar-gutter` and I don't think it's what we need. It will always keep the scrollbar even if the content doesn't have one. But yes, it will solve the problem of content shifting.

Issue #269, opened 2023-07-27
Reduce Primitive components to one with tag name props


MellKam:
I would like to propose a different option for creating primitive components. Now the code literally creates 16 components in a loop using reduce and additionally spread operator to create a new object. 

https://github.com/radix-vue/radix-vue/blob/73c3303d0eaa5c1aea11632d150770920b64f26a/packages/radix-vue/src/Primitive/Primitive.vue#L119-L123

I don't see any reason for this. Wouldn't it be easier to just create one component with a prop for tag name? We can do something similar to headless-ui with a "as" props. If you still want to maintain a similar api to radix, you could treat `:asChild="true"` as `as="template"`. 

```vue
<template>
  <PrimitivButton>
    ...
  </Primitive>
</template>

<template>
  <Primitive as="button">
    ...
  </Primitive>
</template>
```

```vue
<template>
  <PrimitivButton asChild>
    ...
  </Primitive>
</template>

<template>
  <Primitive as="template">
    ...
  </Primitive>
</template>
```

Here's an example of how I would do it. 
```ts
import {
  defineComponent,
  h,
  mergeProps,
  cloneVNode,
  type PropType,
  Fragment,
  type VNode,
} from "vue";
import { renderSlotFragments, isValidVNodeElement } from "@/shared";

type AsTag =
  | "a"
  | "button"
  | "div"
  | "form"
  | "h2"
  | "h3"
  | "img"
  | "input"
  | "label"
  | "li"
  | "nav"
  | "ol"
  | "p"
  | "span"
  | "svg"
  | "ul"
  | "template";

const throwError = (componentName: string) => {
  throw new Error(
    [
      `Detected an invalid children for \`${componentName}\` with \`asChild\` prop.`,
      "",
      "Note: All components accepting `asChild` expect only one direct child of valid VNode type.",
      "You can apply a few solutions:",
      [
        "Provide a single child element so that we can forward the props onto that element.",
        "Ensure the first child is an actual element instead of a raw text node or comment node.",
      ]
        .map((line) => `  - ${line}`)
        .join("\n"),
    ].join("\n")
  );
};

export const Primitive = defineComponent({
  inheritAttrs: false,
  props: {
    asChild: {
      type: Boolean,
      default: false,
    },
    as: {
      type: String as PropType<AsTag>,
      default: "div",
    },
  },
  setup(props, { attrs, slots }) {
    const asTag = props.asChild ? "template" : props.as;

    if (asTag !== "template") {
[...]

PR #278, opened 2023-07-30
Refactor Primitive

MellKam:
#269


packages/radix-vue/src/ScrollArea/ScrollAreaScrollbarImpl.vue
```diff
@@ -128,5 +130,5 @@ useResizeObserver(rootContext?.content, handleSizeChange);
     @pointerup="handlePointerUp"
```
zernonia, line 130: we need to add `as` as injectedValue in `SCROLL_AREA_SCROLLBAR_INJECTION_KEY`.

then we can pass in the correct `:as="scrollbarContext?.as"` here

zernonia (commented):
Teriffic work @MellKam !


Could you pull in latest main? I've added a few internal component, which is using `PrimitiveDiv` still, which is causing error.

MellKam:
> Teriffic work @MellKam !
> 
> Could you pull in latest main? I've added a few internal component, which is using `PrimitiveDiv` still, which is causing error.

Thanks. I'll try to do it soon.

packages/radix-vue/src/ScrollArea/ScrollAreaScrollbarImpl.vue
MellKam: [178e90ab59f8a4c659e2a600e0e31e09f06369f7](https://github.com/radix-vue/radix-vue/commit/178e90ab59f8a4c659e2a600e0e31e09f06369f7)

zernonia (approved):
nicely done @MellKam ! 🤩

Issue #266, opened 2023-07-27
Throw error when missing injected value


MellKam:
I would like to propose to create some small utility to handle provided values. Because now there is no logic in components that would throw errors when injectedValue is missing. Here is an example of how radix handles this https://github.com/radix-ui/primitives/blob/main/packages/react/context/src/createContext.tsx . I think it's pretty easy to port to vue concepts.

MellKam:
Here is example from the arc repository: https://github.com/chakra-ui/ark/blob/main/packages/vue/src/context.ts

MellKam:
Here's an example of how I would write it

```ts
import { type InjectionKey, inject, provide } from "vue";

export const createContext = <ContextValue>(rootComponentName: string) => {
  const injectionKey: InjectionKey<ContextValue> = Symbol(
    rootComponentName + "Context"
  );

  /**
   * @throws When failed to inject context.
   * This happens when the component injecting the context is not a child of the root component providing the context.
   */
  const injectContext = (fallback?: ContextValue) => {
    const context = inject(injectionKey, fallback);
    if (!context) {
      throw new Error(`Component must be used within ${rootComponentName}`);
    }
    return context;
  };

  const provideContext = (contextValue: ContextValue) => {
    provide(injectionKey, contextValue);
  };

  return [injectContext, provideContext] as const;
};
```

```ts
export const [injectAlertDialogContext, provideAlertDialogContext] =
  createContext<AlertDialogContext>("AlertDialogRoot");
```

khairulhaaziq:
yeah we were going to do that someday later, you can create a pr if you want to handle it, thanks!

PR #279, opened 2023-07-30
Fix for: injection "Symbol()" not found

smastrom:
This is a way to fix and close the issue #229 and any further injection warning.

I have created two main utils which should be used in any other component from now on to avoid warnings and precisely inject stuff:

1. Conditionally inject values provided by parent components (only if actually provided)
2. Dynamically inject the collection the consumer needs. 

Both also return an empty object on the server.

I left comments in the code, which I think are better than writing here the logic and all the rest. In short:

#### provide

`provide` can be called as usual, the only new thing introduced is that in the component that calls `provide`, this should be added as well in the setup() function:

```ts
defineExpose({
  providerKey: SOME_INJECTION_KEY,
});
```

#### inject

In the component that may or may not inject the value, this should be used in place of Vue's `inject`:

```ts
const injectedValue = injectSafely(
  SOME_INJECTION_KEY
);
```
This basically will traverse the parent chain until it finds a component that exposes the injection key. If it is found it means that the value is 100% provided, and it will inject the key passed as first parameter, if not it returns an empty object, so no warning is returned.

### 2. Dynamically injecting collections

I have created two new utils that should be used in place of `useCollection`:

#### provideCollection

This should be used in place of `provide` along with `defineExpose`:

```ts
provideCollection(sourceElement)

defineExpose({
  providesCollection: true
})
```

It also returns a reactive copy of the collection, like `useCollection`.

```ts
const collection = provideCollection(sourceElement)
```

This will provide to the children the focusable elements, computing an injection key based on the component uid instead of using a symbol.

This means that when a component that calls `provideCollection` is mounted, it will provide the collection with a distinct key according to its location in the tree, avoiding overrides and overlaps.

In any Root component (like `DropdownMenuRoot`), if any child renders a shared component that is supposed to inject the collection (like `BaseMenuItem`) this value should also be exposed to help to break earlier the loop that looks the parent chain in case no collection is provided.

```ts
defineExpose({ isComponentRoot: true })
```

#### injectCollection

This should be called in place of `inject`. It will traverse the parent tree until it finds the parent that exposes `providesCollection: true`. Once it is found, it injects the key correspondent to the collection. The key is computed using a fixed prefix + the component uid that exposes `providesCollection: true`.

Here we just cannot use a Symbol because `BaseMenuItem` can be child of any radix-vue component, so we can't import all the possible Symbols from any possible parent, and it must be computed.

```ts
const collection = injectCollection()
```

### Other notes

I've also fixed the way the radio/checkbox injections were handled, now each component provides a separate value, and we let `DropdownMenuIndicator` understand if a radio or checkbox is provided by looking at the return value.  

### Next steps

I worked on the Dropdown menu and I can say it is totally fixed/refactored. I preferred to work on the playground which I commented out any other component, you can restore the previous App.vue, it won't break any functionality as `useCollection` is still there.

The components which inject the previous-way will still throw warnings. So I think we should refactor the injections inside those components and then test everything.

### Nuxt Notes
When rendering on the server, if for example a template node has an injected value in the attributes like this:

`:data-state="injectedValue?.someRef?.value`

Make sure to optionally chain everything because inject returns an empty object on the server, and it can break things if trying to access `.value` for example. This can be avoided by returning an object with the same shape instead of an empty one on the server, like this:

```ts
return { someRef: ref() }
```

Let me know what you think and if that's fine for you, I'll refactor the injections in any other component that throws warnings.


BayBreezy:
Commenting here again to get an update when it is merged :). Thank you for your work @smastrom.

khairulhaaziq:
quite mindblown with open-source, hahah thanks! I will let @zernonia review this

zernonia:
That was fast @smastrom ! While checking on your PR, I found this issue when navigating dropdown menu using keyboard, which causes the component to break

```
DropdownMenuRoot.vue:60 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'focus')
    at Object.changeSelected (DropdownMenuRoot.vue:60:28)
    at handleKeydown (DropdownMenuTrigger.vue:48:20)
```

smastrom:
> That was fast @smastrom ! While checking on your PR, I found this issue when navigating dropdown menu using keyboard, which causes the component to break
> 
> ```
> DropdownMenuRoot.vue:60 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'focus')
>     at Object.changeSelected (DropdownMenuRoot.vue:60:28)
>     at handleKeydown (DropdownMenuTrigger.vue:48:20)
> ```

There you go, I forgot to comment out something I was testing 😵‍💫

smastrom:
There you go, sorry I forgot to comment in something I commented out during dev 😵‍💫

packages/radix-vue/src/DropdownMenu/DropdownMenuSubContent.vue
```diff
@@ -41,13 +36,28 @@ const props = withDefaults(defineProps<DropdownMenuSubContentProps>(), {
...
 
+defineExpose({
+  providesCollection: true,
+});
+
+const rootInjectedValue = injectSafely<DropdownMenuProvideValue>(
+  DROPDOWN_MENU_INJECTION_KEY,
+  Components.ROOT
+);
+
```
zernonia, line 47: I like this `inhjectSafely` utility. However I think we can improve it.


Im thinking since we already getting the `DROPDOWN_MENU_INJECTION_KEY` from RootComponent, can we someone use that key as an identifier instead of exposing all the `componentName`, and importing it from`constants`?

This way, we don't need to modify much for other components that are having this `Injection` warning, making migration to this `injectedSafely` so much easier.

smastrom: Brilliant! I have committed the changes.

There's a small change that must be noted. Instead of the root component name I have defined the expose `isComponentRoot` to break earlier the loop and can be shared with `injectCollection` (which before was using `isRoot`).

```ts
defineExpose({
  isComponentRoot: true,
  providerKey: SOME_ROOT_PROVIDER_KEY,
})
```

zernonia: hmm.. that's not quite what I was looking for.

I've tested with the following, by checking if the `injectionKey` was provided in the `getCurrentInstance`, and it works quite well.


```ts
export function injectSafely<T>(injectionKey: InjectionKey<T>) {
  if (isSSR) return {} as T; // Suppress Nuxt SSR warning

  const instance = getCurrentInstance();
  // @ts-ignore provides type was remove from type `ComponentInternalInstance` due to internal feature
  // It's fine for us to use as a injection checker
  if (instance?.provides?.[injectionKey]) return inject(injectionKey);
  return {} as T;
}
```

smastrom: I know what you mean, but as per Vue docs, relying on internal properties in production is not safe nor advised because those properties can change between Vue versions without further notice.

There's a reason why the Vue team didn't include them in the types and don't want us to rely on them, and that's exactly why I decided to go with the `defineExpose` approach instead of the above one: it's a bit more verbose I know that, but at least is safe.

I believe that when a package is meant to be used by thousands of people, we just cannot suppress an explicit core warning with `@ts-ignore` or to say: "warnings are fine since they're not displayed in production". People's is simply going to have problems during the first minutes or development and choose something else.

I did my best to help you out, but it doesn't seem we're on the same page on this matter. And that's perfectly fine. 

I personally won't keep working on this PR unless thing are done safely. If you're not ok with that, feel free to close it or to adapt the code as you see fit.

zernonia: Thanks for the thorough explanation and reasoning with this approach. Your finding was super valuable! 

Sorry if my suggestion or comment made you fed up, I'm trying to understand, and find the best solution for this issue here.

BayBreezy: Soooo.. umm.. anything on this? @zernonia

zernonia:
Thank you so much for this PR @smastrom ! I've refactor the DropdownMenu component, and ported some of your suggestion here to https://github.com/radix-vue/radix-vue/pull/291. 😁

Also, I've come to a decision not to use the `injectSafely` you proposed here, reason being:

1. If `injection "Symbol()" not found`, we should fix it in a core, and think twice our logic when using provide/inject. (which I have done for DropdownMenu above)
2. If someone using the library, and forgot to use the `Root` component for example, it should throw error (instead of hiding away the warning). This would be track via https://github.com/radix-vue/radix-vue/issues/266.

PR #287, opened 2023-08-06
fix(useBodyScrollLock): improve document SSR

sadeghbarati:
To prevent SSR error (`document is not defined`) in Nuxt playground,
Used **existing** utils in `@vueuse/core` and `@vueuse/shared` modules


zernonia (approved):
lgtm! thanks @sadeghbarati

PR #290, opened 2023-08-08
feat: add nuxt module

### LINKED ISSUE #123: feat: Add Nuxt module for Component autoImport
We should have a module that would autoImport all component in `radix-vue`.

owlnai:
This PR adds a Nuxt module for auto-imports. The Nuxt playground is updated too to make use of this module.

## Description
- This adds a Nuxt module built with unbuild as a package to be released alongside `radix-vue`. 
- Every exported component by `radix-vue` will be consumed by the Nuxt autoimports - no need to keep a separate list ✨ 
- Updated the Nuxt playground to avoid explicitly importing Radix components and Vue `ref`.
## Related Issue
Resolves #123 

## How Has This Been Tested?
Tested the Nuxt playground and all primitives that were present work out-of-the-box, no import needed. 

Screenshot from Nuxt Devtools showing components auto-imported by the library

### DIFF
--- a/packages/radix-nuxt/build.config.ts
+++ b/packages/radix-nuxt/build.config.ts
@@ -2,6 +2,7 @@ import { defineBuildConfig } from 'unbuild'
 
 export default defineBuildConfig({
   failOnWarn: false,
+  outDir: '../radix-vue/dist/nuxt',
   externals: [
     '@nuxt/schema'
   ],
--- a/packages/radix-nuxt/build.config.ts
+++ b/packages/radix-nuxt/build.config.ts
@@ -1,0 +1,8 @@
+import { defineBuildConfig } from 'unbuild'
+
+export default defineBuildConfig({
+  failOnWarn: false,
+  externals: [
+    '@nuxt/schema'
+  ],
+})
--- a/packages/radix-nuxt/package.json
+++ b/packages/radix-nuxt/package.json
@@ -1,0 +1,42 @@
+{
+  "name": "@radix-vue/nuxt",
+  "version": "0.0.1",
+  "type": "module",
+  "description": "Nuxt module for UnoCSS",
+  "license": "MIT",
+  "keywords": [
+    "radix",
+    "radix-vue",
+    "nuxt-module",
+    "nuxt",
+    "nuxt3"
+  ],
+  "sideEffects": false,
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "require": "./dist/index.cjs",
+      "import": "./dist/index.mjs"
+    }
+  },
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.mjs",
+  "types": "./dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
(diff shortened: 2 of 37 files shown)


zernonia:
Awesome job @owlnai!!!

@khairulhaaziq  How do you think we should release this package ya? As another package `@radix-vue/nuxt`, or we bundle all together into `radix-vue`, and access Nuxt module from `radix-vue/nuxt`, similar to [Unocss](https://unocss.dev/integrations/nuxt)

khairulhaaziq:
> Awesome job @owlnai!!!
> 
> 
> 
> @khairulhaaziq  How do you think we should release this package ya? As another package `@radix-vue/nuxt`, or we bundle all together into `radix-vue`, and access Nuxt module from `radix-vue/nuxt`, similar to [Unocss](https://unocss.dev/integrations/nuxt)

Bundling together seems like a decent approach, what do you think?

zernonia:
> Bundling together seems like a decent approach, what do you think?

I think so too. Main benefit I see is that doing so Nuxt will always be up-to-date with the latest version of Radix-vue. And given that we have tree-shakeable capacitbility, need not to worry to much about the bundled size 😁

owlnai:
I agree that since the package is reasonably light (<1kb _if_ bundled) we could ship it with the core package. However, the downside of bundling together is that we wouldn't be able to fix Nuxt issues without bumping the core package

zernonia:
> I agree that since the package is reasonably light (<1kb _if_ bundled) we could ship it with the core package. However, the downside of bundling together is that we wouldn't be able to fix Nuxt issues without bumping the core package

I believe that should be fine. As we may not encountered many issue with the Nuxt package. 😁
Could you help update the `build`  function to include nuxt module into core?

owlnai:
> As we may not encountered many issue with the Nuxt package. 😁

True. Also, SSR issues will be easier to fix since we can now make adjustments to Nuxt itself 😄 
> Could you help update the build function to include nuxt module into core?

The build command already builds all /packages/ so that should be covered. I have included the Nuxt export so `radix-vue/nuxt` should work, in theory, as a bundled package. I've never used multiple exports so I'd be glad if someone else could review it :)

zernonia:
can we build `nuxt` into `radix-vue/dist/nuxt`, and change the import path from `"./../radix-nuxt/dist/index.mjs"` to `./dist/nuxt/index.mjs` ?

reason is `npm publish` will look at the `files` in `package.json`.. using relative path might messed up the pathing I afraid.
```json
 "files": [
    "./dist",
    "./LICENSE"
  ],
```

owlnai:
Right, good catch. Unbuild now outputs to the radix-vue/dist directory.
```
i Building @radix-vue/nuxt                                                          16:51:10
√ Build succeeded for nuxt                                                          16:51:13
  ../radix-vue/dist/nuxt/index.cjs (size: 753 B, exports: default)                  16:51:13
  ../radix-vue/dist/nuxt/index.mjs (size: 475 B, exports: default)  
```

zernonia:
Really a job well done @owlnai !!! Appreciate the contributions! 😁❤️

PR #291, opened 2023-08-08
[feat] Dropdown menu 2/3

zernonia:


khairulhaaziq:
shouldnt the trigger be focused when click outside? has the safepolygon been implemented or will be on next phase?

khairulhaaziq (approved):

zernonia:
@khairulhaaziq  thanks for capturing that! focus when click outside has been fixed. Could you help me check again? 😁

The safepolygon is in. It is a crude way of handling that. using similar function from RadixUI. After looking at FloatingUI's `useHover` (react only) composable I decided to go with Radix approach for now.

https://github.com/radix-vue/radix-vue/assets/59365435/8129b2a0-7c24-4161-86fd-40b48422e68c

zernonia:
Special thanks to @smastrom for suggesting some good provide/inject approach 😁

PR #301, opened 2023-08-11
chore(accordion-item): remove watcher that was logging 'run' in accordion-item

BayBreezy:
## Description

Removed the watcher in the AccordionItem component that was logging the word 'run' when the `dataState` value was truthy.

image


zernonia (approved):
lgtm!

PR #313, opened 2023-08-19
chore: organize codebase

mukundshah:
- move eslint to root of project and update the config to use `@antfu/eslint-config`
- add vscode specific settings and extension recommendation
- update PR and Issues template
- perform some miscellaneous cleanup


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,62 @@
...
+      label: Reproduction steps
+      description: How do you trigger this bug? Please walk us through it step by step.
+      value: |
+        1.
+        2.
+        3.
+        ...
+      render: bash
+    validations:
+      required: true
```
zernonia, line 44: can we add a hint of adding `reproduction link` using CodeSandbox? similar to [vue/core](https://github.com/vuejs/core/blob/main/.github/ISSUE_TEMPLATE/bug_report.yml?plain=1#L27)

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,8 @@
+blank_issues_enabled: true
+contact_links:
+  - name: 📚 Documentation
+    url: https://www.radix-vue.com/
+    about: Check the documentation for usage of Radix Vue.
+  - name: 🗨️ Discord
+    url: https://discord.gg/jZUsrAADe5
```
zernonia, line 7: let's replace this with `https://chat.radix-vue.com/`

zernonia (changes requested):
let's remove other "eslint config" file in `radix-vue`. 
image

some minor suggestion, else lgtm!!!

zernonia:
@mukundshah  Thanks for the PR. However I only realized this changes broke the whole app, especially related to `radix-vue` package. If you look at any `root` component, you can see error `TS1184: Modifiers cannot appear here.`, thus I've reverted it.

Could you look into that please? 😁

mukundshah:
@zernonia `<script setup>` could not contain exports.

Reference: 
[eslint.vuejs.org/rules/no-export-in-script-setup.html](https://eslint.vuejs.org/rules/no-export-in-script-setup.html)
[https://github.com/vuejs/rfcs/blob/master/active-rfcs/0040-script-setup.md](https://github.com/vuejs/rfcs/blob/master/active-rfcs/0040-script-setup.md)

zernonia:
> @zernonia `<script setup>` could not contain exports.
> 
> Reference: [eslint.vuejs.org/rules/no-export-in-script-setup.html](https://eslint.vuejs.org/rules/no-export-in-script-setup.html) https://github.com/vuejs/rfcs/blob/master/active-rfcs/0040-script-setup.md


hmm.. but it's not only in `script setup`, also in normal `script` tag. This changes kinda messed up the codebase, and had to refactor everything (which is not a good idea).

Blinks44:
Is not exports in ```script-setup``` has been fixed in vue 3.3+?
https://github.com/vuejs/core/issues/7794

mukundshah:
@Blinks44 thank you for bringing that rfc tto attenttion.

@zernonia i studied the issue, and conducted some experiments as well.

The conclusion is:
- Only types and interface could be exported from `<script setup>`, which also suggests, it only works for TS. 
- Named export of anything JS (variables, constants, or functoions) [aka ES Module exports] is not allowed.
- Exported interface and types should be first thing in the `<script setup>`.

- vuejs/core#7794
- vuejs/rfcs#227
- https://github.com/vuejs/language-tools/blob/4e20cfbb354671c4621d5bebeded2cbd3b430703/vue-language-tools/vue-language-service/src/plugins/vue.ts#L156
*though is  really old commit, i tested it, and it is still true*

Both the editor and vue-tsc should have thrown error/warnings at the very beginning. I reverted every change I made, and I got red everywhere it's supposed to.  Thus, the final conclusion is a refactor.

zernonia:
> @Blinks44 thank you for bringing that rfc tto attenttion.
> 
> @zernonia i studied the issue, and conducted some experiments as well.
> 
> The conclusion is:
> 
> * Only types and interface could be exported from `<script setup>`, which also suggests, it only works for TS.
> * Named export of anything JS (variables, constants, or functoions) [aka ES Module exports] is not allowed.
> * Exported interface and types should be first thing in the `<script setup>`.
> * [defineProps<PropsType> generates wrong code in typescript when  PropsType is exported vuejs/core#7794](https://github.com/vuejs/core/issues/7794)
> * [`<script setup>` vuejs/rfcs#227](https://github.com/vuejs/rfcs/pull/227)
> * https://github.com/vuejs/language-tools/blob/4e20cfbb354671c4621d5bebeded2cbd3b430703/vue-language-tools/vue-language-service/src/plugins/vue.ts#L156
>   _though is  really old commit, i tested it, and it is still true_
> 
> Both the editor and vue-tsc should have thrown error/warnings at the very beginning. I reverted every change I made, and I got red everywhere it's supposed to. Thus, the final conclusion is a refactor.

I think you missed my pointer earlier @mukundshah .
I know `script setup` only export `ineterface`/`types`, but the `script` tag should be able to export `const`, as it acts like a normal `js` file.

I know the syntatic sugar `script setup` has it's limitation, but using `script` itself works pretty well, and it does all the job we want it to do. This definitely not called for a refactor. And I hope that you understand, this refactor is super unncessary if we dont have a clear and heavily favoured benefit that it will bring. 

Furthermore, I found support for `defineComponent`  is quite limitated with TS. You can checkout the TS for `props/emits` defined in each component in [Headless UI](https://github.com/tailwindlabs/headlessui/blob/main/packages/%40headlessui-vue/src/components/combobox/combobox.ts). All these is not not neccessary with the current `.vue` SFC approach that we have.

You might argue that `>3.3` we can have better types for `defineComponent`, yes you are right, but it's still limited as we still need to manually define the `emits` in each component.

zernonia:
Also, I've found a way to use `antfu/eslint`, by having this config. Doing so we prevent all exports being moved to `script setup`, which prevent us from having those export const issue.

 `eslint-plugin-vue` currently have no active way to distinguish `script` & `script setup` due to how it was wired in `vue/core`. So we can just disable this `eslint` hint for our repo.

Reference: https://github.com/vuejs/eslint-plugin-vue/issues/1577


```
module.exports = {
  extends: '@antfu',

  rules: {
    'import/first': 'off',
  },
}
```

mukundshah:
> I know `script setup` only export `ineterface`/`types`, but the `script` tag should be able to export `const`, as it acts like a normal `js` file.

@zernonia we still need to refactor the code base. Those exports need to be arranged. Isn't it?

zernonia:
> > I know `script setup` only export `ineterface`/`types`, but the `script` tag should be able to export `const`, as it acts like a normal `js` file.
> 
> @zernonia we still need to refactor the code base. Those exports need to be arranged. Isn't it?

Nope. Not needed 😁

PR #319, opened 2023-08-21
chore: add antfu/eslint

zernonia:


zernonia:
@mukundshah  thanks for the original PR https://github.com/radix-vue/radix-vue/pull/313). I've picked some commit from yours into this eslint specific PR. 

By modifying some rules, we are able to apply lint to all the core codebase we have without breaking anything.
Thanks for looking into the `export` issue ya! Appreciate it nonetheless 😁

PR #342, opened 2023-08-24
feat: add `prefix` option to Nuxt module

DamianGlowala:
This PR allows for (optionally) prefixing auto-imported Nuxt components' names.

### DIFF
--- a/packages/radix-nuxt/src/index.ts
+++ b/packages/radix-nuxt/src/index.ts
@@ -3,7 +3,11 @@ import { addComponent, defineNuxtModule } from '@nuxt/kit'
 import type { } from '@nuxt/schema' // workaround for TS bug with "phantom" deps
 import * as components from 'radix-vue'
 
-export default defineNuxtModule({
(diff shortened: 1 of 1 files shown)


packages/radix-nuxt/src/index.ts
```diff
@@ -3,18 +3,25 @@ import { addComponent, defineNuxtModule } from '@nuxt/kit'
 import type { } from '@nuxt/schema' // workaround for TS bug with "phantom" deps
 import * as components from 'radix-vue'
 
-export default defineNuxtModule({
+export interface ModuleOptions {
+  prefix: string
```
zernonia, line 7: Let's make this prefix optional. So when user wanna have empty options, it wont throw ts error 😁

DamianGlowala: All module options are optional in Nuxt by default (`Partial<ModuleOptions>`) 😁

zernonia (approved):
lgtm!

PR #347, opened 2023-08-25
[feat]: `createContext` shared composables, updates Tooltip

MellKam:
#266

After updating the package to the latest version (v0.1.26), there were many warnings with a missing injection key in my code. It was because of new TooptipProvider and it's context, although it is not required to be used. Here is my proposal based on the example given in the issue

 
```ts
interface TooltipContextValue { ... }

export const [injectTooltipRootContent, provideTooltipRootContext]
  = createContext<TooltipContextValue>('TooltipRoot')
  
// in parent
provideTooltipRootContext({ ... })

// in child
const context = injectTooltipRootContext()
// the context will be returned with type `TooltipContextValue`, otherwise error like this will be thrown
// Error: Component must be used within TooltipRoot

// If we dont want to throw an error but still try to inject context, we can provide fallback
const context = injectTooltipRootContext(null)
// Now context will be of type `TooltipContextValue | null`
```

This reduces the number of optional chaining operators and just seems right for me.


zernonia:
Thanks for the PR @MellKam .. I wanna to note that this Tooltip Provider seems to be `required` from RadixUI point of view. It wasn't included previously as we haven't fully complete the API.

If the introduction of `v0.1.26`,  the API is now complete and it *should* required to be wrap inside Tooltip Provider.

Your propose changes for `provide/inject` is good, however I think in this case we want set `const providerContext = injectTooltipProviderContext(null)`.

MellKam:
@zernonia Yes, you are right. I have fixed it

packages/radix-vue/src/Tooltip/TooltipContentImpl.vue
```diff
@@ -77,30 +77,20 @@ const ariaLabel = computed(() => {
...
-  collisionBoundary: props.collisionBoundary,
-  collisionPadding: props.collisionPadding,
-  arrowPadding: props.arrowPadding,
-  sticky: props.sticky,
-  hideWhenDetached: props.hideWhenDetached,
-}))
+const popperContentProps = computed(() => {
+  const { ariaLabel: _, ...restProps } = props
+  return restProps
+})
```
MellKam, line 83: I have also corrected this computed. The effect should be the same, but with less code

zernonia:
Absolutely well done @MellKam ! Im quite liking this composable, and the ability to use `inject`  on any child component without inferring the `injectionKey` from parent, and make sure components that required `parentComponent` would throw error indicating **incorrect component structure** to user.

zernonia (approved):
lgtm!!! 🚀

PR #348, opened 2023-08-25
chore: tailwind intelisense for stories 

MellKam:
I don't know how you guys write tailwind without autocomplete and other intelisense features.  

image


zernonia (approved):
Nice catch! Thanks for the fix!

Issue #353, opened 2023-08-28
[Bug]: Missing `as-child` Prop in `PopperArrow` 


onmax:
### Environment

```bash
Developement/Production OS: Windows 10 19043.1110
Node version: 16.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.3.3
Client OS: Windows 10 19043.1110
Browser: Chrome 90.0.4430.212
```

### Link to minimal reproduction

https://www.radix-vue.com/components/popover.html#arrow

### Steps to reproduce

Read below

### Describe the bug

The docs state that there is a `as-child` prop into the `PopoverArrow`, which uses `PopperArrow`. 

`PopperArrow` does not include `as-child`. See [source code](https://github.com/radix-vue/radix-vue/blob/main/packages/radix-vue/src/Popper/PopperArrow.vue)

### Expected behavior

Have a prop in `PopperArrow`.

[Radix UI](https://www.radix-ui.com/primitives/docs/components/popover#arrow)

zernonia:
Im sorry I didnt look at this closer previously.
We did already extends `PrimitiveProps`  in `PopperArrowProps`, so `PopoverArrow` should have `asChild` props already.

Please check again @onmax

onmax:
Yes. The Prop definition has the prop in fact, but the I don't think is using that prop at all. Iread the code, and I didn't see where `as-child` is used. Maybe I missed it 🥲

Nonetheless, when I try to use it, I don't think is working. 

I placed a random SVG to see if I see something different to the `polygon` Radix uses, but I still see the radix. Here the playground:

https://codesandbox.io/p/sandbox/sq6yyd?file=%2Fsrc%2FApp.vue%3A1%2C1-36%2C1

Thanks for the help :)

zernonia:
Yeah you are right @onmax ! There's a missing binding of props in `PopperArrow.vue` 😁

PR #365, opened 2023-08-30
bug: PopperArrow missing props as-child

D-Lite:
#353 

:) first time contributing here! Looking forward to your kind corrections😊

### DIFF
--- a/packages/radix-vue/src/Popper/PopperArrow.vue
+++ b/packages/radix-vue/src/Popper/PopperArrow.vue
@@ -25,7 +25,9 @@ import {
 import Arrow from '@/shared/component/Arrow.vue'
 import { type ArrowProps } from '@/shared/component/Arrow.vue'
 
-defineProps<PopperArrowProps>()
+const props = withDefaults(defineProps<PopperArrowProps>(), {
+  as: 'svg',
+})
 
 const contentContext = inject<PopperContentContextValue>(POPPER_CONTENT_KEY)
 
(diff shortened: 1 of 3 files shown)


onmax:
Legend!!

Thank you!

packages/radix-vue/src/shared/component/Arrow.vue
```diff
@@ -39,6 +39,8 @@ const { primitiveElement } = usePrimitiveElement()
     :height="height"
     viewBox="0 0 30 10"
     preserveAspectRatio="none"
+    :as-child="props.asChild"
+    :as="as"
```
zernonia, line 43: let's remove this props assigment as we have `v-bind` to bind that props already 😁

packages/radix-vue/src/Popper/PopperArrow.vue
```diff
@@ -67,6 +69,8 @@ const baseSide = computed(() =>
       :style="{
         display: 'block',
       }"
+      :as="as"
+      :as-child="props.asChild"
```
zernonia, line 73: Nice.. this is the missing piece here! thanks for that @D-Lite ! Perhaps let replace `props.asChild` with `asChild`?

packages/radix-vue/src/shared/component/Arrow.vue
D-Lite: Yeah, I forgot to clean up properly

packages/radix-vue/src/Popper/PopperArrow.vue
D-Lite: Great! made the fix and updated the pr

zernonia (approved):
lgtm!

PR #370, opened 2023-08-31
chore: update Tooltip to use Portal instead of Teleport

onmax:
As we chat in Discord, we will recomend the use of `Teleport`.

Therefore, I just updated the demos for Tooltips


zernonia (approved):
lgtm!

Issue #373, opened 2023-09-02
[Bug]: 「Select」Cannot destructure property 'shouldExpandOnScrollRef' of 'o' as it is undefined.


libondev:
### Environment

```bash
Developement/Production OS: MacOS 13.5.1(22G90)
Node version: 18.12.1
Package manager: pnpm@8.7.0
Radix Vue version: 0.1.29
Vue version: 3.3.4
Nuxt version: null
Nuxt mode: null
Nuxt target: null
CSS framework: unocss@0.55.3
Client OS: MacOS 13.5.1(22G90)
Browser: Chrome 116.0.5845.96
```

### Link to minimal reproduction

https://stackblitz.com/edit/vite-vue3-radix-vue-unocss?file=src%2Fcomponents%2FSelect.vue

### Steps to reproduce

1. run install 
2. click `Select` link

### Describe the bug

Mouse over 'Select Scrollupbutton' and 'Select Scrolldownbutton' when there are too many options

### Expected behavior

No mistakes

### Conext & Screenshots (if applicable)

![iScreen Shoter - 20230902182706795](https://github.com/radix-vue/radix-vue/assets/47918504/859c9535-ecf2-4f73-9bc0-ea3741c34f40)

Issue #372, opened 2023-09-02
[Bug]: Switch `disabled` prop don't work


libondev:
### Environment

```bash
Developement/Production OS: MacOS 13.5.1(22G90)
Node version: 18.12.1
Package manager: pnpm@8.7.0
Radix Vue version: 0.1.29
Vue version: 3.3.4
Nuxt version: null
Nuxt mode: null
Nuxt target: null
CSS framework: unocss@0.55.3
Client OS: MacOS 13.5.1(22G90)
Browser: Chrome 116.0.5845.96
```

### Link to minimal reproduction

https://stackblitz.com/edit/vite-vue3-radix-vue-unocss
1. npm run dev
2. access `/switch` route

### Steps to reproduce

`<SwitchRoot />` component set `disabled` prop `true`

### Describe the bug

As the title says, the disabled property is set but does not work properly.

---
When I looked at the source code, I found that the `as` property defaults to `button` , but the `as` property can be set to some other value, so I thought I'd change it to:

[source code](https://stackblitz.com/edit/vitejs-vite-fhhsmc?file=src%2Fcomponents%2FSwitch.vue,src%2Fcomponents%2FSwitch-fixed.vue)

```diff
 function toggleCheck() {
+  if (disabled.value)
+    return

   checked.value = !checked.value
 }
```

```diff
  <Primitive
    v-bind="$attrs"
    :id="id"
    ref="primitiveElement"
    role="switch"
    :type="as === 'button' ? 'button' : undefined"
    :value="value"
    :aria-label="$attrs['aria-label'] || ariaLabel"
    :aria-checked="checked"
    :aria-required="required"
    :data-state="checked ? 'checked' : 'unchecked'"
    :data-disabled="disabled ? '' : undefined"
    :as-child="asChild"
    :as="as"
+  :disabled="disabled"
    @click="toggleCheck"
    @keydown.enter.prevent="toggleCheck"
  >
    <slot />
  </Primitive>
```

I'm not sure this change is reasonable, and if it's OK, I'd be happy to submit it to PR.  😋

### Expected behavior

The disabled property does not respond to the click event

PR #374, opened 2023-09-02
fix:  Switch disabled not work (#372),  Select shouldExpandOnScrollRef undefined #373

libondev:


zernonia (approved):
lgtm! Thanks for reporting the bug, and even push a fix to it! Really appreciate it! ❤️

PR #378, opened 2023-09-04
fix(TabsTrigger): resolve `<TabsTrigger />` component `data-orientation` attribute value display error

libondev:
iScreen Shoter - Google Chrome - 230904221949

This PR corrects the issue of the 'data-orientation' attribute not displaying properly, and because the dependencies of the triggerId and contentId attributes are completely consistent, integrating them into a computational attribute can reduce the side effects it produces

### DIFF
--- a/packages/radix-vue/src/Tabs/TabsTrigger.vue
+++ b/packages/radix-vue/src/Tabs/TabsTrigger.vue
@@ -19,10 +19,8 @@ const props = withDefaults(defineProps<TabsTriggerProps>(), {
 })
 const context = inject(TABS_INJECTION_KEY)
 
-const ids = computed(() => ({
-  triggerId: makeTriggerId(context!.baseId, props.value),
-  contentId: makeContentId(context!.baseId, props.value),
-}))
+const triggerId = computed(() => makeTriggerId(context!.baseId, props.value))
+const contentId = computed(() => makeContentId(context!.baseId, props.value))
 
 const isSelected = computed(() => props.value === context?.modelValue.value)
 </script>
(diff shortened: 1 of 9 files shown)


packages/radix-vue/src/Tabs/TabsTrigger.vue
```diff
@@ -19,30 +19,32 @@ const props = withDefaults(defineProps<TabsTriggerProps>(), {
 })
 const context = inject(TABS_INJECTION_KEY)
 
-const triggerId = computed(() => makeTriggerId(context!.baseId, props.value))
-const contentId = computed(() => makeContentId(context!.baseId, props.value))
+const ids = computed(() => ({
+  triggerId: makeTriggerId(context!.baseId, props.value),
+  contentId: makeContentId(context!.baseId, props.value),
+}))
```
zernonia, line 26: Can we maintain 2 computed? Cause in some case we are providing this value, and I would like to keep this consistent as others 😁

packages/radix-vue/src/Tabs/TabsTrigger.vue
```diff
@@ -19,30 +19,32 @@ const props = withDefaults(defineProps<TabsTriggerProps>(), {
...
       :data-disabled="disabled ? '' : undefined"
-      :data-orientation="context?.orientation"
-      @mousedown="(event) => {
+      :data-orientation="context?.orientation.value"
+      @mousedown.left="(event) => {
         // only call handler if it's the left button (mousedown gets triggered by all mouse buttons)
         // but not when the control key is pressed (avoiding MacOS right click)
-        if (!disabled && event.button === 0 && event.ctrlKey === false) {
+        if (!disabled && event.ctrlKey === false) {
           context?.changeModelValue(value);
```
zernonia, line 48: Thanks for the chnages here! looks good! 😁

packages/radix-vue/src/Tabs/TabsTrigger.vue
libondev: I don't know why, I have already submitted the recovered code in my fork repository, but that commit has not been added to this PR. Did I do it incorrectly？

zernonia (approved):
lgtm!

PR #383, opened 2023-09-08
fix: fix default radiogroup orientation

### LINKED ISSUE #382: [Bug]: RovingFocus should have no default `orientation`
@khairulhaaziq :

I found the radio has accessibility issue where the orientation should be vertical instead of horizontal, https://www.shadcn-vue.com/docs/components/radio-group.html, and then I found out default orientation is undefined https://github.com/radix-vue/radix-vue/blob/044407eb44ebdfe525accbb5fb8a6be113d838fa/packages/radix-vue/src/RadioGroup/RadioGroupRoot.vue#L74 here, which then inherit from the horizontal in roving focus. I wonder if we should add default orientation as vertical in radix-vue radiogroup, or just add it in shadcn radiogroup

khairulhaaziq:
fixes #382 

removing default orientation allows bidirectional orientation

### DIFF
--- a/packages/radix-vue/src/RovingFocus/RovingFocusGroup.vue
+++ b/packages/radix-vue/src/RovingFocus/RovingFocusGroup.vue
@@ -58,7 +58,6 @@ import {
 const props = withDefaults(defineProps<RovingFocusGroupProps>(), {
   loop: false,
   dir: 'ltr',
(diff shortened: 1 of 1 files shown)


zernonia (approved):
works pretty well @khairulhaaziq !

PR #421, opened 2023-09-26
fix: allow AvatarImage to be reactive

lepture:
## Use case

There is a list of users, and you want to select one from the list. When a user is selected, we display the selected user's avatar.

## Problem

1. We select A user, and A user has no avatar, it will show the Fallback avatar.
2. Then select B user, and B user has avatar, without reactive, it will still show the Fallback avatar.

### DIFF
--- a/packages/radix-vue/src/Avatar/AvatarImage.vue
+++ b/packages/radix-vue/src/Avatar/AvatarImage.vue
@@ -6,7 +6,7 @@ export interface AvatarImageProps extends PrimitiveProps {}
 </script>
 
 <script setup lang="ts">
-import { inject, useAttrs, watch } from 'vue'
+import { inject, onUpdated, ref, useAttrs, watch } from 'vue'
 import { Primitive, type PrimitiveProps } from '../Primitive'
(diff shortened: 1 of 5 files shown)


packages/radix-vue/src/Avatar/AvatarImage.vue
```diff
@@ -6,34 +6,61 @@ export interface AvatarImageProps extends PrimitiveProps {}
...
+  (newSrc) => {
+    if (!newSrc) {
+      updateStatus('error')()
+    }
+    else {
+      const image = new window.Image()
+      updateStatus('loading')()
+      image.onload = updateStatus('loaded')
+      image.onerror = updateStatus('error')
+      image.src = newSrc
```
zernonia, line 55: can we remain the `useImageLoadingStatus` composables, and make it accept a reactive `src`. Also.. there's some missing status what you changes here.

lepture:
@zernonia updated

zernonia (approved):
Thanks for the PR @lepture ! I've push some code to handle the fallback so it would not render fallback immediately after the `src` changes!

Once again, thanks for your contribution! ❤️

PR #429, opened 2023-09-28
fix: fixes the overflow when a toast is displayed on mobile

BayBreezy:
On mobile, rendering the toast component would cause an overflow on the x axis.
After inspecting the DOM, I realised that adding a width of 0px and display of inline-block fixed the issue.

### DIFF
--- a/packages/radix-vue/src/VisuallyHidden/VisuallyHidden.vue
+++ b/packages/radix-vue/src/VisuallyHidden/VisuallyHidden.vue
@@ -14,7 +14,8 @@ withDefaults(defineProps<VisuallyHiddenProps>(), { as: 'span' })
       // See: https://github.com/twbs/bootstrap/blob/master/scss/mixins/_screen-reader.scss
       position: 'absolute',
       border: 0,
-      width: 1,
+      width: '0px',
+      display: 'inline-block',
       height: 1,
       padding: 0,
       margin: -1,


zernonia (commented):
I dont think we should set as `0px` (based on this https://github.com/twbs/bootstrap/blob/9c1c3d4d048eeb8b685e5442c21489f101c4ad05/scss/mixins/_visually-hidden.scss)

zernonia:
maybe there's something we're missing on mobile.. causing it to renders differently

zernonia:
I just check on mobile.. seems like `text-slate12` is not computing the desired css. It should have nothing to do with `visuallyHidden` tho


image

BayBreezy:
Hey @zernonia , the element in question is actually the span tag that renders the text from the toast
image

zernonia:
Icic.. could you screenshot what is the UI looks like currently with that `element` blocking the UI?

BayBreezy:
When I add the proposed styles to that element in the DOM, the toast gets rendered properly
image

But without the styles, I get this overflow
image

Blinks44:
Hm, even mobile version on radix-vue.com looks like this after showing a toast.

zernonia:
thanks guys!!! I really my stupid mistake lol...

it should be `1px`, and not number `1` when we apply styling. @BayBreezy  instead of setting `0px`, can you help with fixing the `height`, `width` and `margin`  by adding the word `px`? 😁

BayBreezy:
Sure thing. And leave the display as inline-block correct? From what I remember about css, width and won't have an effect on items that are displayed inline

PR #430, opened 2023-09-28
doc: update composition.md

pkrinesh:
It should be `vue`


zernonia (approved):
lgtm! Thanks for catching that @pkrinesh

Issue #452, opened 2023-10-14
[Bug]: dropdown doesn't is not scrollable when there's not enough room to display it in full


remiconnesson:
### Environment

```bash
Developement/Production OS: Windows 10 19043.1110
Node version: 16.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.3.3
Client OS: Windows 10 19043.1110
Browser: Chrome 90.0.4430.212
```

### Link to minimal reproduction

https://www.radix-vue.com/components/dropdown-menu.html#dropdownmenu

### Steps to reproduce

Open https://www.radix-vue.com/components/dropdown-menu.html#dropdownmenu and shorten your window so the dropdown doesn't fit.

then, open https://www.radix-ui.com/primitives/docs/components/dropdown-menu  and shorten your window so the dropdown doesn't fit.

### Describe the bug

# In the current `radix-vue`

In the gif below, we can notice the dropdown is too big for the space remaining, so a part of it is unreachable. As the dropdown remove the scrollability of the page.

![radix-vue-current-dropdown](https://github.com/radix-vue/radix-vue/assets/28846642/79572add-c6c4-4230-8180-8658da2a2226)

# in the original radix

In the original radix, when there's not enough space, the dropdown is trimmed to fit in the available space and use a scroll area inside to display the content
![radix-original-dropdown](https://github.com/radix-vue/radix-vue/assets/28846642/b99711cc-0659-48a2-853f-c0781cadae3c)

### Expected behavior

The expected behavior would be to match the orginal radix-ui's.

romansp:
@remiconnesson it's expected behavior for Radix Vue/Radix UI primitives.

In your gif you're comparing Radix-vue primitive and Radix UI Theme component. If you look at [Radix UI primitive](https://www.radix-ui.com/primitives/docs/components/dropdown-menu) and repeat your test it doesn't become scrollable by default.

Primitives are low-level UI components and should be treated as base layer of your app components. Primitives like `DropdownMenu` only compute available height/width and expose these as CSS variables and will not constrain content by default.

[...]

PR #454, opened 2023-10-15
fix: add overflow to ensure dropdown menu is not cut off on smaller screens

D-Lite:
radix-452


D-Lite:
#452

zernonia:
Thanks for the PR @D-Lite . However as the comment here pointed out, https://github.com/radix-vue/radix-vue/issues/452#issuecomment-1763126913, the `overflow` styling should be apply by user itself as Radix Vue provides only the "neccessary" styling.

You can easily do so:
```css
[data-radix-popper-content-wrapper]  {
  overflow: auto
}
```
 
So I'm closing ya 😁

Issue #377, opened 2023-09-04
[Feature]: Ability for library user to preventDefault event handler provided by radix-vue


MellKam:
### Describe the feature

I have come across a situation where I need to prevent the default behavior set in the `TooltipTrigger` component on click. In the Radix UI, they employ a custom function that assesses event prevention and executes the handler only if it hasn't been prevented. I think we need this feature too.

https://github.com/radix-ui/primitives/blob/ea6376900d54af536dbb7b71b4fefd6ec2ce9dc0/packages/core/primitive/src/primitive.tsx#L1-L15

### Additional information

- [ ] I intend to submit a PR for this feature.
- [ ] I have already implemented and/or tested this feature.

PR #436, opened 2023-10-02
V1

zernonia:
Yes. v1 soon! 

### Coming real soon!

Docs:
- [x] Complete Docs with Custom API
- [x] Showcase packages/websites

Package: 
- [x] Test all form controls
- [x] More testing
- [x] Config provider (dir, bodyScroll)
   - https://github.com/radix-vue/radix-vue/issues/385
   - https://www.radix-ui.com/primitives/docs/utilities/direction-provider

Fixes:
- https://github.com/radix-vue/radix-vue/issues/377
- https://github.com/radix-vue/radix-vue/issues/446
- https://github.com/radix-vue/radix-vue/issues/444
- https://github.com/radix-vue/radix-vue/issues/434
- https://github.com/radix-vue/radix-vue/issues/388
- https://github.com/radix-vue/radix-vue/issues/385
- https://github.com/radix-vue/radix-vue/issues/448
- https://github.com/radix-vue/radix-vue/issues/433

PR #459, opened 2023-10-21
fix: handle `event.preventDefault()` for `TooltipTrigger` events

MellKam:
#377

I had a problem with preventing the default behavior of closing a tooltip on click trigger, because `TooltipTrigger` was not handling the defaultPrevented case.


zernonia:
I've included this in https://github.com/radix-vue/radix-vue/pull/436, added `disableClosingTrigger` prop. Could you check that on `1.0.0-beta.2`?

MellKam:
@zernonia Aaah, this is cool. I haven't seen it. 
Then, I'll close this one.

PR #467, opened 2023-10-25
refactor: use `normal script` alongside the script setup for exposing type and interface

sadeghbarati:
~~Best Practices~~ Happy Practices for `Component Orders`
Best Practices don't really exist 😅 

```ts
<script lang="ts">
// types and interface

// context
</script>

<script setup lang="ts">
// imports

// defineOptions
// defineProps
// defineEmits

// setup
// computed
// watch

// lifecycles
// methods/functions

// defineExpose
</script>
```

### DIFF
--- a/packages/radix-vue/src/Accordion/AccordionContent.vue
+++ b/packages/radix-vue/src/Accordion/AccordionContent.vue
@@ -1,10 +1,13 @@
+<script lang="ts">
+import type { PrimitiveProps } from '@/Primitive'
+
+export interface AccordionContentProps extends PrimitiveProps {}
+</script>
+
 <script setup lang="ts">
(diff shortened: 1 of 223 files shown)


zernonia:
I wonder can we have `eslint` to do this 😂
Let say in future we add new components.. Letting formatter to handle this would be super duper awesome.

sadeghbarati:
it depends on `eslint-plugin-vue`

I saw that `Component Order` in tutorial sites and found that interesting, maybe we should provide a style guide for new components?

Just like Vue

zernonia (approved):
Thanks for the cleanup @sadeghbarati . Hopefully somehow we can tap into `vue-eslint` to parse `.vue` with the "happy format" that we have here 😂

However, nicely done and LGTM!

PR #469, opened 2023-10-26
chore: enable sourcemaps for `radix-vue` package

romansp:
Enable sourcemaps for `radix-vue` so it's easier to understand and debug possible usage issues in development.

After running locally `pnpm --filter radix-vue build` I see that `*.js.map` files are generated in dist. I don't have much experience with publishing to `npm` but I guess these files should be picked up automatically on publish.

### DIFF
--- a/packages/radix-vue/vite.config.ts
+++ b/packages/radix-vue/vite.config.ts
@@ -23,6 +23,7 @@ export default defineConfig({
     },
   },
   build: {
+    sourcemap: true,
     lib: {
       name: 'radix-vue',
       fileName: 'index',


zernonia:
After adding source map.. the total `dist size` increased to **3.02MB** (previously 1.12MB). Not sure if we wanna include that into the package tho 🙈

romansp:
Yeah that's really unfortunate and understand your concern. Though these maps won't be included in the production build of the app. At least not by default.

I'm looking at the generated sourcemap file and it seems to inlcudes sourcemap for several external packages too, e.g. `vueuse`, `defu`, `aria-hidden`. Unfortunately I'm not highly familiar with the process of generating source maps for libraries. Maybe that's expected behavior or maps for external packages could be omitted somehow? 

Would appreciate if anyone could help with figuring this out.

zernonia:
Thanks for pointing that out.. didn't realized it consist of source map for external packages.
I have no idea how to exclude those too 😂

sadeghbarati:
Probably the quick way is to keep `radix-vue` package and also publish **individual** packages just like `radix-ui`

- `radix-vue` **without sourcemap** for those who don't care about sourcemap

- `radix-vue/accordion`, `radix-vue/alert-dialog`, .... **with sourcemap**

zernonia:
Closing this in favor of https://github.com/unovue/radix-vue/pull/1432 😁

PR #478, opened 2023-10-29
refactor(primitive): migrate to SFC

ferferga:
There's nothing wrong with render functions, but given the whole project uses SFC and it's the de-facto choice in the Vue docs, I think it's a good move to be consistent about this. It also let Vue newbies be more confident about rolling their own changes into the components.

Marking as draft because pnpm fails for me when I do ``pnpm i``, so I couldn't run any tests. I also planned to take over the ``Slot`` component, but first I want to be sure that Primitive alone works. I'm not familiar witrh pnpm, so I'm not sure how to solve this issue:
```
 ERR_PNPM_JSON_PARSE  Unexpected end of JSON input while parsing empty string in /workspaces/radix-vue/node_modules/builtin-modules/package.json
```


zernonia:
it's funny we actually moved from `vue` to `ts`, and we previously has logic in `Primitive.ts`, then we extract it out to `Slot`. So actually I dont see any needed changes here other than just consistency. 

So I would say let's not do any changes here.

ferferga:
May I ask why it's no longer a SFC?

I think all of this can be refactored to not need an additional Slot component at all too, but ofc I need to check that thoroughly before PRing.

Any idea why pnpm fails on me?

ferferga:
@zernonia Hey, I think you missed my comment? 😅☺️

zernonia:
Ahh my bad @ferferga ! I indeed missed that out!
The major reason why we swap from `vue` to `ts` back then is due to HMR not working for any slots inside Primitive, which is a bit factor for DX.

PR #491, opened 2023-11-02
fix: TabsContent appearance

MellKam:
Maybe resolves (works for me, but not sure yet): https://discordapp.com/channels/1121805548782878813/1169341352807497779
I have also added ability to access `present` variable from v-slot.


zernonia:
Tested with SSR and indeed this solves the issue. Thanks @MellKam ! 
Nice addition to have the `v-slot` 👍🏻

PR #500, opened 2023-11-07
feat: OTP input

zernonia:


sadeghbarati:
Thanks for the feature 🙏 

Note that one input that lookalike multiple input >>> multiple input

zernonia:
Thanks @sadeghbarati !
what do you mean by "Note that one input that lookalike multiple input >>> multiple input" yaa? 😁

sadeghbarati:
I shared a link in the OTP Input issue called "rci"

https://github.com/leonardodino/rci

The author of that repo finds a better way to create pin inputs instead of creating multiple inputs 

https://user-images.githubusercontent.com/8649362/136673700-bd227d9c-9919-49d6-ae92-bbbef7882365.png

zernonia:
Ahh icic!! that's an awesome suggestion @sadeghbarati ! Yeah I would try out and see how extensible that approach is 🤩

zernonia:
Saw this [Caveats](https://github.com/leonardodino/rci#caveats) from the package, which unfortunately limits the potential of creating different kind of OTP design. Thus I would like to stick with multi-input approach, which is much more flexible in terms of constructing custom design.

WDYT @sadeghbarati

sadeghbarati:
Right ✅️ 

LGTM then ^_^

PR #506, opened 2023-11-09
fix(PopoverArrow): make asChild work as expected

kdnk:
`asChild` in `Popover` component didn't work because `slot` wasn't used in `PopperArrow` component.


zernonia (approved):
Thanks again @kdnk !

PR #514, opened 2023-11-13
[PopoverArrow][PopperArrow] fix: enable to pass viewBox

kdnk:
When `asChild` is set to `true` and an SVG is passed as a slot, the `viewBox` preassigned to the `Primitive` is used unintentionally. This PR allows the viewBox to be passed as props. In the case of `asChild`, the `viewBox` of `Primitive` is set to `undefined`, and the `viewBox` of the `asChild` SVG can be used as is.

## As is

|histoire|html|
|-|-|
|||

## To be

|histoire|html|
|-|-|
|||

### DIFF
--- a/packages/radix-vue/src/shared/component/Arrow.vue
+++ b/packages/radix-vue/src/shared/component/Arrow.vue
@@ -14,6 +14,13 @@ export interface ArrowProps extends PrimitiveProps {
    * @default 5
    */
   height?: number
+
+  /**
+   * The viewBox of the arrow in pixels.
+   *
+   * @default '0 0 30 10'
+   */
+  viewBox?: string
 }
 </script>
 
@@ -27,6 +34,7 @@ const props = withDefaults(defineProps<ArrowProps>(), {
   width: 10,
   height: 5,
   as: 'svg',
+  viewBox: '0 0 30 10',
 })
 
 const { primitiveElement } = usePrimitiveElement()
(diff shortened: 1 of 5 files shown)


zernonia:
when passing `asChild=true`, the attribute in the first child component will be prioritized, you just need to set the viewBox attribute in `svg` to overwrite the one from the `PopoverArrow`, eg:

```html
<svg view-box="0 0 40 20"> // explicitly set the viewBox you want
    <polygon points="0,0 40,0 20,20" fill="red" />
</svg>
```

...will overwrite `viewBox="0 0 30 10"`. Thus not needing any changes here.


Did I missed out something? 😁

kdnk:
@zernonia Thank you for the quick reply.
Sorry, I think the histoire was not enough for this PR.

I updated the histoire, so could you take another look?

If we have svg element as vue component under `PopoverArrow`, we'll see viewBox as `0 0 30 10` not `0 0 40 20`.

packages/radix-vue/src/shared/component/Arrow.vue
```diff
@@ -38,7 +38,7 @@ const { primitiveElement } = usePrimitiveElement()
     v-bind="props"
     :width="width"
     :height="height"
-    viewBox="0 0 30 10"
+    :viewBox="asChild ? undefined : '0 0 30 10'"
     preserveAspectRatio="none"
```
zernonia, line 42: Icic! Let's do the same for `preserveAspectRatio` too. I imagine there's a case where dev might want to change this.

kdnk: > Let's do the same for preserveAspectRatio too. I imagine there's a case where dev might want to change this.

I missed. We should do that!👍

kdnk:
@zernonia I made a change. Please take another look!

zernonia (approved):
LGTM! Thanks for your contribution @kdnk ! 🚀

PR #519, opened 2023-11-15
fix(FocusScope): document.activeElement doesn't reflect the correct focused element

zernonia:


zernonia:
@enkot  do help me check more use cases ya! we still need this MutationObserver to capture if the focused element was being removed from DOM.

packages/radix-vue/src/FocusScope/FocusScope.vue
```diff
@@ -113,13 +113,18 @@ watchEffect((cleanupFn) => {
...
   function handleMutations(mutations: MutationRecord[]) {
-    const focusedElement = document.activeElement as HTMLElement | null
-    if (focusedElement !== document.body)
-      return
-    for (const mutation of mutations) {
-      if (mutation.removedNodes.length > 0)
-        focus(container)
+    const isLastFocusedElementExist = container.contains(lastFocusedElementRef.value)
+    if (!isLastFocusedElementExist) {
+      for (const mutation of mutations) {
```
enkot, line 124: Do we need to loop through the mutations and check for removed nodes? 
`lastFocusedElementRef` it is always up to date anyway.

Just tested this: 
```ts
function handleMutations(mutations: MutationRecord[]) {
  const isLastFocusedElementExist = container.contains(lastFocusedElementRef.value)
  if (!isLastFocusedElementExist) focus(container)
}
```
and it handles focus properly when: 
- [x] focused element gets removed
- [x] some element added on button click
- [x] some element removed on button click
- [x] some element added on input change
- [x] some element removed on input change

zernonia: Ah make sense! Thanks for checking the use cases. I've tested too and it seems like working as expected!

PR #546, opened 2023-12-05
fix(Dialog): remove `aria-label="Close"` from DialogClose

### LINKED ISSUE #541: [Bug]: DialogClose overrides accessible name given by child content
### Environment

```bash
Developement/Production OS: macOS 11.6.2
Node version: (whatever Stackblitz uses)
Package manager: (whatever Stackblitz uses)
Radix Vue version: 1.2.3
Vue version: latest
Nuxt version: -
Nuxt mode: -
Nuxt target: -
CSS framework: -
Client OS: macOS 11.6.2
Browser: Chrome 119.0.6045.159
```

https://www.radix-vue.com/components/dialog.html

### Steps to reproduce

1. Use DialogClose to wrap any button that closes the modal, whether `as-child` or `<MyCustomButton :as="DialogClose">`

### Describe the bug

The resulting HTML is `<button aria-label="Close" ...>{{ the button's content }}</button>`.

This means that the button's accessible name is overridden by the `aria-label="Close"`, and "Close" will be what is announced by screen readers, for example.

### Expected behavior

I'm not really sure what the best solution is. Personally, I interpret DialogClose as the preferred way of closing a Radix Vue Dialog. It doesn't necessarily have to be a "Close" button, it could be a "Cancel" or an "Okay" button (which otherwise behave the same as a "Close" button). I would like to be able to use DialogClose in such cases too, instead of needing to manually control the state myself.

Maybe it should be the consumer's responsibility to give the 'Close' button an accessible name, like in Radix UI?

### Conext & Screenshots (if applicable)

It's also a problem for AlertDialogAction and AlertDialogCancel.

hoonweiting:
To be in line with Radix UI. Authors can still set `aria-label="Close"` themselves, or use another method to provide an accessible name.

Resolves #541

### DIFF
--- a/docs/content/components/dialog.md
+++ b/docs/content/components/dialog.md
@@ -468,6 +468,26 @@ const container = ref(null)
 
 Adheres to the [Dialog WAI-ARIA design pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialogmodal).
 
+### Close icon button
+
+When providing an icon (or font icon), remember to label it correctly for screen reader users.
+
+```html line=8-9
+<DialogRoot>
+  <DialogTrigger />
+  <DialogPortal>
+    <DialogOverlay />
+    <DialogContent>
+      <DialogTitle />
+      <DialogDescription />
+      <DialogClose aria-label="Close">
+        <span aria-hidden>×</span>
+      </DialogClose>
+    </DialogContent>
+  </DialogPortal>
+</DialogRoot>
+```
+
 ### Keyboard Interactions
 
 <KeyboardTable
--- a/packages/radix-vue/src/Dialog/DialogClose.vue
+++ b/packages/radix-vue/src/Dialog/DialogClose.vue
@@ -19,7 +19,6 @@ const rootContext = injectDialogRootContext()
   <Primitive
     v-bind="props"
     :type="as === 'button' ? 'button' : undefined"
-    aria-label="Close"
     @click="rootContext.onOpenChange(false)"
   >
     <slot />


docs/content/components/dialog.md
```diff
@@ -468,6 +468,26 @@ const container = ref(null)
...
+      <DialogTitle />
+      <DialogDescription />
+      <DialogClose aria-label="Close">
+        <span aria-hidden>×</span>
+      </DialogClose>
+    </DialogContent>
+  </DialogPortal>
+</DialogRoot>
+```
+
```
hoonweiting, line 490: I added a note in the docs to remind authors to give their 'Close' icon button an accessible name as well (copied and pasted from Toast)! Let me know if it should be in a separate PR.

I'd do it for AlertDialog and Popover too, but I'm not sure if that'd be too out of scope. (Anyway, I could not get `pnpm docs:dev` to work 🙈 But it's late and I'll try again another time.)

zernonia: Nic addition! Let's handle AlertDialog and Popover in another PR ya 😁
Have you tried `pnpm i` then `pnpm docs:dev`?

zernonia (approved):
Thanks for the PR @hoonweiting ! LGTM!

docs/content/components/dialog.md
hoonweiting: Yep, I ran `pnpm i` and then `pnpm docs:dev`, both in the root folder and the docs folder. The error/warning was that the entrypoint of 'radix-vue' couldn't be resolved, or something along those lines.

I will say that I've never used pnpm before, I installed it yesterday for the first time ever 🙈 Maybe I should have tried restarting my machine at some point. Anyway, not to worry now, I will ask for help in the Discord server if/when I do try again!

PR #550, opened 2023-12-07
fix(PinInput): modelValue elements mutated directly

### LINKED ISSUE #549: [Bug]: modelValue elements in `PinInput` are being mutated directly
### Environment

```bash
Developement/Production OS: Windows 10 19045.3693
Node version: 20.10.0
Package manager: pnpm@8.11.0
Radix Vue version: 1.2.4
Vue version: 3.3.10
CSS framework: tailwindcss@3.3.5
Client OS: Windows 10 19045.3693
Browser: Edge 119.0.2151.97, Firefox 120.0.1
```

https://stackblitz.com/edit/3bjcu7?file=src%2FApp.vue

### Steps to reproduce

1. Instantiate a `ref<string[]>`
2. Associate the created ref with a PinInputRoot using the `:model-value` prop.
3. Observe that the ref is modified directly, even in the absence of an `@update:model-value` handler.

### Describe the bug

The individual elements of the `model-value` prop in a PinInputRoot component are [directly mutated](https://github.com/radix-vue/radix-vue/blob/main/packages/radix-vue/src/PinInput/PinInputInput.vue#L36). Consequently, this results in the corresponding ref being updated even when no `@update:model-value` handler is used.

I will create a PR for this shortly.

### Expected behavior

If no `@update:model-value` handler is provided, the ref assigned to the `model-value` prop of a PinInputRoot should remain unaltered.

### Conext & Screenshots (if applicable)

_No response_

Scalamando:
Resolves #549.

Addresses the direct mutation of `modelValue` elements in `PinInputInput`s by using a temporary copy.

This workaround seems a little clunky, but currently, I couldn't think of a better solution. Feedback is appreciated!

### DIFF
--- a/packages/radix-vue/src/PinInput/PinInputInput.vue
+++ b/packages/radix-vue/src/PinInput/PinInputInput.vue
@@ -33,7 +33,7 @@ function handleInput(event: Event) {
   }
 
   target.value = target.value.slice(-1)
-  context.modelValue.value[props.index] = target.value
+  updateModelValueAt(props.index, target.value)
 
   const nextEl = inputElements.value[props.index + 1]
   if (nextEl)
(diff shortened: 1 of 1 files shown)


packages/radix-vue/src/PinInput/PinInputInput.vue
```diff
@@ -104,12 +104,18 @@ function handlePaste(event: ClipboardEvent) {
...
     input.value = value
-    context.modelValue.value[i] = value
+    updateModelValueAt(i, value)
     input.focus()
   }
   inputElements.value[lastIndex]?.focus()
 }
 
+function updateModelValueAt(index: number, value: string) {
+  const tempModelValue = context.modelValue.value
```
zernonia, line 114: Because `Array` is a "Reference value" in JS,  we need to assign the array differently, else the same issue still occur.

```ts
  const tempModelValue = [...context.modelValue.value]
```

zernonia (approved):
LGTM! Thanks for the contribution @Scalamando !

PR #556, opened 2023-12-12
chore: change docs theme-color 

MellKam:
The current `theme-color` simply doesn't match the logo and og:image color scheme.

image


zernonia (approved):
LGTM! Thanks for catching that 😁

PR #563, opened 2023-12-16
Typo in Presence.ts error message

oaleynik:
A tiny typo in Presence.ts error message.


zernonia:
Thanks for catching that and the PR @oaleynik ! 🚀

PR #569, opened 2023-12-21
fix(typo): update bug-report.yml

sanscheese:
Spotted a typo: Conext -> Context


zernonia (approved):
Thanks for catching that! 😁

PR #570, opened 2023-12-24
Update slider.md: correct parameter type

rsoltanzadeh:
`v` is of type `number[] | undefined`:

image


zernonia (approved):
Thanks for catching that! and the PR! 😁

Issue #578, opened 2024-01-02
[Bug]: [Select] Selecting an option triggers a touch event on elements positioned behind


lennartzellmer:
### Environment

```bash
Developement/Production OS: MacOS Catalina
Node version: 20
Package manager: pnpm@8.6.0
Radix Vue version: 1.2.5
Vue version: 3.0.0
CSS framework: tailwindcss@3.3.3
Browser: Chrome 90.0.4430.212
```

### Link to minimal reproduction

https://codesandbox.io/p/devbox/radix-vue-select-forked-ddlzyz

### Steps to reproduce

Open minimal reproduction. 

Open DevTools and switch to responsive view. 

This way you force your browser to also send touch events. 

Select an item from the list and see that also a click on the background is recorded.

### Describe the bug

When on mobile and selecting something from the select list, not only the item selected is clicked, but there is a second click event that hits whatever is behind this item. Well, it's not actually a click event, it's more of a touch-end event. This event propagates through to whatever container is there and triggers the click.

### Expected behavior

Only the item in the select list is clicked

### Context & Screenshots (if applicable)

There's already an open discussion on the original Radix components because they have the exact same bug in the current version. (Not fixed since a year :( )
https://github.com/radix-ui/primitives/issues/1658

The bug can be reproduced in the docs even. Video:
https://github.com/radix-vue/radix-vue/assets/2241624/d91d7c56-5e38-4663-93cd-fdf1a11d1be1

The most likely solution is to prevent the additionally imitated touch event from propagating.

PR #579, opened 2024-01-02
Prevent touch event propagation on select

lennartzellmer:
Solving this issue: https://github.com/radix-vue/radix-vue/issues/578


zernonia (approved):
Thanks for the issue and PR @lennartzellmer 💚 Works pretty well!

PR #592, opened 2024-01-08
fix: modal false not closing when clicking on other trigger

zernonia:


peter-emad99:
why this fix isn't published yet ?

zernonia:
publishing in 1-2 day @peter-emad99 . More fixes will be coming along

Issue #600, opened 2024-01-11
[Bug]: DropdownMenuItem contains disabled="false"


JacksonBowe:
### Environment

```bash
Package manager: npm@10.0.0
Radix Vue version: ^1.2.6
Vue version: 3.0.0
CSS framework: tailwindcss@3.4.0
Client OS: Windows 10 19043.1110
Browser: Chrome 90.0.4430.212
```

### Link to minimal reproduction

https://stackblitz.com/github/JacksonBowe/shadcn-quasar?file=src%2Fcomponents%2Fui%2Fdropdown-menu%2FSDropdownMenuItem.vue

### Steps to reproduce

Appears to only happen with certain frameworks. I am using Quasar Framework and get the issue. If I use base Vite then the issue doesn't occur.
Note: even with base Vite the superflous `disabled="false"` is still present

### Describe the bug

Due to the `disabled="false"` prop something is happening in my setup that ignores the `false` and disables the element. No idea why the `false` isn't being respected.

### Expected behavior

`disabled` prop should not be present if `false`

### Context & Screenshots (if applicable)

Attached is a comparison between Radix and Radix-Vue (both snippets taken from docs webpage). 

Note that the Radix component does not contain `disabled="false"`

![image](https://github.com/radix-vue/radix-vue/assets/26427343/bb0365d2-6c3c-43c9-9797-2acce2f1f519)

zernonia:
Make sense @MellKam . We can remove the `disabled` attributes in `MenuItemImpl.vue`!

sadeghbarati:
> Note that even if any `MenuItem` component in the original radix-ui has `disabled={true}`, it does not display the disabled attribute at all. Only `aria-disabled` and `data-disabled`. I think this is mainly because menu items are by default just a `div` and it doesn't have a disabled state. So, I think we should do the same thing. Just remove disabled attribute at all

@MellKam 

What If I used (just for example) 

```html
<DropdownMenuItem as-child>
    <Button>asd</Button>
</DropdownMenuItem>
```

If we remove disabled attribute, only data-disabled and data-disabled with pass down to Button Component so button appearance will look disabled but not completely disabled

I don't know why it's disabled act like this `disabled="false"` is should be removed when it's value is false or undefined

MellKam:
@sadeghbarati Yeah, in that case you will need to pass disabled twice. First to MenuItem and then to your button. I agree that it may not be obvious. 

This can get even more tricky when you use ’as’ props

[...]

PR #608, opened 2024-01-13
fix: don't set disabled attribute to `MenuItem` component

MellKam:
resolves: #600

Original radix-ui doesn't set disabled attribute to MenuItem component. Only `aria-disabled` and `data-disabled` https://github.com/radix-ui/primitives/blob/c31c97274ff357aea99afe6c01c1c8c58b6356e0/packages/react/menu/src/Menu.tsx#L708-L709

If you want to use MenuItem as button - use `asChild` props 

```vue
<MenuItem as-child disabled>
	<button disabled>
		Hello world
	</button>
</MenuItem>
```

### DIFF
--- a/packages/radix-vue/src/Menu/MenuItemImpl.vue
+++ b/packages/radix-vue/src/Menu/MenuItemImpl.vue
@@ -66,8 +66,8 @@ defineExpose({
     data-radix-vue-collection-item
     :aria-disabled="disabled || undefined"
     :data-disabled="disabled ? '' : undefined"
+    :disabled="disabled && as === 'button' ? true : undefined"
     :data-highlighted="isFocused ? '' : undefined"
-    :disabled="as === 'button' && disabled ? true : undefined"
     @pointermove="handlePointerMove"
     @pointerleave="handlePointerLeave"
     @focus="
--- a/packages/radix-vue/src/Menu/MenuItemImpl.vue
+++ b/packages/radix-vue/src/Menu/MenuItemImpl.vue
@@ -66,8 +66,8 @@ defineExpose({
     data-radix-vue-collection-item
     :aria-disabled="disabled || undefined"
     :data-disabled="disabled ? '' : undefined"
-    :disabled="disabled"
     :data-highlighted="isFocused ? '' : undefined"
+    :disabled="as === 'button' && disabled ? true : undefined"
     @pointermove="handlePointerMove"
     @pointerleave="handlePointerLeave"
     @focus="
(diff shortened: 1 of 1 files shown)


sadeghbarati:
But `MenuItemImpl` has `disabled` prop which you are removing it 🙅 

https://github.com/radix-ui/primitives/blob/c31c97274ff357aea99afe6c01c1c8c58b6356e0/packages/react/menu/src/Menu.tsx#L639

---

We have to investigate why Vue act like this, it should act like this 👇 

`isDisabled: true` -> disabled attributes appears

`isDisabled: false` or `undefined` ---> disabled attribute should remove completely

[`Vue SFC Playground`](https://play.vuejs.org/#eNp9UUtPAyEQ/isjFzVp2kNvzdb46kEParRHLlt2dqVlAWFomzT73wXWvhLThAMz34NvmB17sHa4DsgmrPDCSUvgkYK941q21jiCHTisoYPamRauI/Waa66F0Z5A+mfpy4XCCqaJdlOXyuMt18WoN4s2sSBsrSoJYwVQLAKR0TCp/qRTzo4+nGUSwM8G48mCUa/IXkf9vVBSrM7EMcTVmdXcNI3CE4NidJKFDRj5OEgtm+HSGx3/YJde4EyY1kqF7t2SjINyNoGMJKxUymxec49cwMG+L75RrP7pL/029Tj7cOjRrZGzA0ala5B6ePb1htt4P4CtqYKK7AvgJ3qjQsrY0x6DrmLsE15O+5I3KXUz97Mtofb7oVLQxOwyn7O43acLox/jjofjrOO6Y90vO53C/A==)

sadeghbarati:
I think we should add `withDefault` to the MenuItemImpl `defineProps`

https://github.com/radix-vue/radix-vue/blob/main/packages/radix-vue/src/Menu/MenuItemImpl.vue#L19


```ts
const props = withDefaults(defineProps<MenuItemImplProps>(), {
   disabled: false
})
```

sadeghbarati:
@MellKam I'm checking Radix UI, you are right

- MenuItem -> should have disabled prop

- MenuItemImpl -> disabled attribute must be removed

MellKam:
> @MellKam I'm checking Radix UI, you are right
> 
> * MenuItem -> should have disabled prop
> * MenuItemImpl -> disabled attribute must be removed

Yep, it's a bit weird at first, but If you think about radix more as the set of logic that can be used on other components, it makes sense. We shouldn't use element specific attributes. It's just confusing that it calls "disabled" but doesn't actually set disabled on the element. 

```vue
                   ↓ handle logic for disabled menu item
<MenuItem as-child disabled>
            ↓ set disabled attribute
    <button disabled>  
        Hello world
    </button>
</MenuItem>
```

Issue #602, opened 2024-01-11
[Bug]: Tabs TabsIndicator doesn't update position or size when the width changes or window is resized


Jordan-Ellis:
### Environment

```bash
n/a
```

### Link to minimal reproduction

https://codesandbox.io/p/devbox/radix-vue-tabs-klnqdj?file=/src/App.vue

### Steps to reproduce

Open the code sandbox and resize the preview pane so the tabs change width.

### Describe the bug

When the tabs width changes, the tab indicator position and size doesn't update, breaking it's positioning. This happens in the latest version (1.3.0). It seems like there needs to be a ResizeObserver or something similar that listens for when the tabs container changes size.

### Context & Screenshots (if applicable)

![Screenshot 2024-01-11 at 1 01 54 PM](https://github.com/radix-vue/radix-vue/assets/18269476/28a0cdb0-abe1-4dde-8c7f-825e2e27a3cf)

zernonia:
Thanks @Jordan-Ellis ! Yup seems like we just need to add a ResizeObserver to `Indicator`. Should be relatively simple, interested for a PR? 😁

PR #609, opened 2024-01-13
fix(TabsIndicator): add ResizeObserver (#602)

Jordan-Ellis:
Fixes the issue where the tab indicator doesn't update its position when the container changes size. #602


zernonia (approved):
Thanks for the PR @Jordan-Ellis !! LGTM!

PR #630, opened 2024-01-19
feat(ScrollArea): expose viewport

teleskop150750:


zernonia (approved):
Thanks @teleskop150750 ! There will be a coming up PR that would exposes all element.. But thanks nonetheless!

Issue #636, opened 2024-01-23
[Feature]: Expose useBodyScrollLock


CapitaineToinon:
### Describe the feature

Expose the useBodyScrollLock composable for use in custom components:

https://github.com/radix-vue/radix-vue/blob/main/packages/radix-vue/src/index.ts#L37

I'm currently using radix-vue in a project but also created my own drawer component. As such, I need to also lock the body when the drawer is opened. I tried implementing my own scroll body lock logic but this conflict with radix-vue's logic and I ended up with weird bugs and the body staying locked.

I would love to be able to just import useBodyScrollLock and have the same logic through my whole app.

Thanks ✌️

### Additional information

- [X] I intend to submit a PR for this feature.
- [X] I have already implemented and/or tested this feature.

CapitaineToinon:
After more exploring, `useBodyScrollLock` doesn't seems to be working the way I was expecting. I assume it would be working the same way vueuse's useScrollLock would, as a composable I can assign to to change the state. https://vueuse.org/core/useScrollLock/#usescrolllock

However it seems to be meant for use by components only as the stack counting the instances of useBodyScrollLock is only decreased on unMounted. What confused me is that useBodyScrollLock returns a ref that can be assigned to and doing so indeed changes the state of the body.

However, when looking at the code, it's never used like that internally by radix. So I think this hook is just not right. Why is resetBodyStyle called whenever no components are hooked rather than when all hook calls are false?

In my opinion, this hook should either:
- Be renamed "bodyScrollLock" and not return a ref (the return value is never used by radix)
- Fixed to reflect the body state based on the value of the ref, not based on how many instances that called the hook are mounted

PR #637, opened 2024-01-23
fix: useBodyScrollLock is now using a singleton logic

CapitaineToinon:
Fix and feature from #636

### DIFF
--- a/packages/radix-vue/src/shared/useBodyScrollLock.ts
+++ b/packages/radix-vue/src/shared/useBodyScrollLock.ts
@@ -1,5 +1,5 @@
 import {
-  createGlobalState,
+  createSharedComposable,
   useEventListener,
 } from '@vueuse/core'
 import { type Fn, isClient, isIOS } from '@vueuse/shared'
@@ -7,18 +7,15 @@ import { nextTick, onBeforeUnmount, ref, watch } from 'vue'
 import { defu } from 'defu'
 import { injectConfigProviderContext } from '@/ConfigProvider/ConfigProvider.vue'
 
-const useInitialOverflowStyle = createGlobalState(() => ref<string | undefined>())
-
-const useBodyLockStackCount = createGlobalState(() => {
+const useBodyLockStackCount = createSharedComposable(() => {
   const count = ref(0)
+  const initialOverflow = ref<string | undefined>()
 
   const context = injectConfigProviderContext({
     scrollBody: ref(true),
   })
 
   let stopTouchMoveListener: Fn | null = null
-
-  const initialOverflow = useInitialOverflowStyle()
 
   const resetBodyStyle = () => {
     document.body.style.paddingRight = ''
--- a/packages/radix-vue/src/shared/useBodyScrollLock.ts
+++ b/packages/radix-vue/src/shared/useBodyScrollLock.ts
@@ -7,18 +7,13 @@ import { nextTick, onBeforeUnmount, ref, watch } from 'vue'
 import { defu } from 'defu'
 import { injectConfigProviderContext } from '@/ConfigProvider/ConfigProvider.vue'
 
-const useInitialOverflowStyle = createGlobalState(() => ref<string | undefined>())
-const useBodyLockStackCount = createGlobalState(() => ref(0))
+const useBodyLockStackCount = createGlobalState(() => {
+  const count = ref(0)
 
-export function useBodyScrollLock(initialState?: boolean | undefined) {
   const context = injectConfigProviderContext({
     scrollBody: ref(true),
   })
 
-  const stack = useBodyLockStackCount()
-  const initialOverflow = useInitialOverflowStyle()
-
-  const locked = ref(initialState)
   let stopTouchMoveListener: Fn | null = null
 
   const resetBodyStyle = () => {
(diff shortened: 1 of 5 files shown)


CapitaineToinon:
This PR changes the useBodyScrollLock hook:

- moved the body lock logic in the createGlobalState that counts useBodyScrollLock instances;
- the useBodyScrollLock now just increments the count based on the value of the locked variable;
- the count is still decremented on onMounted so the usage in current radix code is unaffected;
- the hook is now exposed to be used in user-code.

For example, I wanted to have my own drawer component so I wrote this custom hook

```ts
import { breakpointsTailwind } from '@vueuse/core'
import { router } from '@inertiajs/core'
import { useBodyScrollLock } from 'radix-vue'

export function useDrawer() {
	const _open = shallowRef(false)

	const { smaller } = useBreakpoints(breakpointsTailwind)
	const isSmallerThanLg = smaller('lg')

	const isOpen = computed(() => isSmallerThanLg.value && _open.value)

	function close() {
		_open.value = false
	}

	function open() {
		_open.value = true
	}

	function toggle() {
		_open.value = !_open.value
	}

	const isLocked = useBodyScrollLock(isOpen.value)

	watch(isOpen, (value) => {
		isLocked.value = value
	})

	onMounted(() => {
		router.on('navigate', close)
		useEventListener(window, 'resize', close)
	})

	onBeforeUnmount(() => {
		_open.value = false
	})

	return {
		isOpen,
		open,
		close,
		toggle,
	}
}
```

And I can now show and hide the drawer to have the same logic applied than if radix ui did. For example, my navbar uses the `--scrollbar-width` css variable to add some padding when radix-ui components are opened.

zernonia:
Thanks @CapitaineToinon . That's interesting. We never expose the composable previously as it was designed to be internal use. But seeing your use case, I think it would be good to export it.

Will review the code real soon ya 😁

CapitaineToinon:
Tests are passing locally 👍

packages/radix-vue/src/shared/useBodyScrollLock.ts
```diff
@@ -7,93 +7,98 @@ import { nextTick, onBeforeUnmount, ref, watch } from 'vue'
 import { defu } from 'defu'
 import { injectConfigProviderContext } from '@/ConfigProvider/ConfigProvider.vue'
 
-const useInitialOverflowStyle = createGlobalState(() => ref<string | undefined>())
-const useBodyLockStackCount = createGlobalState(() => ref(0))
+const useBodyLockStackCount = createGlobalState(() => {
```
zernonia, line 10: one reason we save the initial overflow style is because we wanna preserve the user defined overflow styling. 

eg:
user set "overflow: hidden", if we didnt preserve this after unlocking body lock, it would revmoe the styling, removing the default behavior

packages/radix-vue/src/shared/useBodyScrollLock.ts
```diff
@@ -7,93 +7,98 @@ import { nextTick, onBeforeUnmount, ref, watch } from 'vue'
...
+  const count = ref(0)
```
zernonia, line 11: also, I would like to keep `createGlobalState` as a state only reactive value, and not incurring side-effect (watcher) inside the function. 

Let move the function/watcher out into the `useBodyScrollLock` composable

packages/radix-vue/src/shared/useBodyScrollLock.ts
CapitaineToinon: Oh make sense. My bad I'll put it back.

packages/radix-vue/src/shared/useBodyScrollLock.ts
CapitaineToinon: Okay I got it working with `createSharedComposable`, I think it's working well now. However while writing the tests and having to mock createSharedComposable made me realize that the implementation could lead to bugs if multiple ConfigProviders are used through the same app. Because the `injectConfigProviderContext` call is inside the `createSharedComposable`, it will only be called once and therefore, if multiple Config Providers exists only the first value will be used.

This is why in the tests I mocked `createSharedComposable` to disable caching and just return the function. So here is the question, is using multiple Config Providers an expected use case or should it be considered an anti-pattern?

Switching to a shared composable style was something neat I wanted to add but isn't strictly needed for this PR which its goal was to open useBodyScrollLock to the outside, which could be done by keeping the code closer to how it used to be.

zernonia: `createSharedComposable` make sense!

packages/radix-vue/src/shared/useBodyScrollLock.ts
```diff
@@ -32,68 +28,83 @@ export function useBodyScrollLock(initialState?: boolean | undefined) {
...
+      stack.value++
+    else
       stack.value--
+  })
 
-    if (stack.value === 0)
-      resetBodyStyle()
+  onBeforeUnmount(() => {
+    if (locked.value)
+      stack.value--
```
zernonia, line 107: @CapitaineToinon  seems like reducing the stack size doesn't trigger the watcher, thus not reseting the body style.
Can you look at it?

CapitaineToinon: Okay I think I got it working. I'm trying to write some tests to avoid the back and forth but I'm facing some caching issues.

If I disable the cache for `createSharedComposable` then I can't do tests such as ensure once 2 components are unmounted, the body unlocks since the count won't be shared. However if I don't cache `createSharedComposable` then my first tests will affect the following tests. I can just unmount components to ensure it cleans up but if the test fails for some reason and unmount isn't called then the next test will be wrongly affected and fail.

Is there a way with vitest to create a new mock module or something in a beforeEach in order to have the `createSharedComposable` being only shared between vue instances inside a same test and not all of them?

CapitaineToinon: Okay let me know how this one looks. Tests are working and the dropdown and popover components work fine on histoire.

zernonia: LGTM @CapitaineToinon !

zernonia (approved):
Thanks for the issue and the PR @CapitaineToinon ! Impressive work here! 😁

PR #647, opened 2024-01-31
feat(Pagination): ellipsis should separate at least 2 pages

andreww2012:
Hello! 

Rendering an ellipsis in the Pagination component when it only put instead of a single page item is not that useful I think:

This PR changes this behavior by requiring the difference between pages that ellipsis separate be at least 2.

### DIFF
--- a/packages/radix-vue/src/Pagination/Pagination.test.ts
+++ b/packages/radix-vue/src/Pagination/Pagination.test.ts
@@ -105,16 +105,19 @@ describe('given show-edges Pagination', () => {
       await wrapper.find('[aria-label="Page 5"]').trigger('click')
     })
 
-    it('should have default 2 siblings on each side', () => {
+    it('should have page 2', () => {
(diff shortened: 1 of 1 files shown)


zernonia:
Hey @andreww2012 .. Not sure what you are looking for.. have you tried changing `siblingCount` to `5`?

andreww2012:
Hey @zernonia!

I don't think this is related to `siblingCount` (and I need to use `1` anyway).

On the screenshot (taken directly from the radix-vue site), "..." is put between pages 1 and 3, but it can just be replaced with "2". Same happens on the right side, in the end of the list. This PR changes exactly that 🙂

Other UI libs handle this case, for example:
- Buefy: https://buefy.org/documentation/pagination
- Quasar: https://quasar.dev/vue-components/pagination#handling-boundary

zernonia (approved):
LGTM! Thanks for the suggestion and changes @andreww2012 !

andreww2012:
I was about to make necessary changes in tests this evening, but you beat me. Thank you for that & merging! :)

PR #649, opened 2024-02-01
V1.4.0

zernonia:


packages/radix-vue/src/Combobox/ComboboxInput.vue
```diff
@@ -63,16 +64,17 @@ function handleInput(event: Event) {
 
   rootContext.isUserInputted.value = true
 }
+
+const pickedProps = reactiveOmit(props, 'autoFocus')
```
iamandrewluca, line 68: @zernonia, we would need, I think, also to omit the `as` attribute and pass it manually otherwise, we get this warning

> [Vue warn]: Vue received a Component that was made a reactive object. This can lead to unnecessary performance overhead and should be avoided by marking the component with `markRaw` or using `shallowRef` instead of `ref`.

```vue
const pickedProps = reactiveOmit(props, 'as', 'autoFocus')

// ...

<Primitive
  v-bind="pickedProps"
  :as="props.as"  
```

zernonia: Make sense @iamandrewluca ! Thanks for that! Forgot to consider that `as` could be a component 😅

Instead of converting to reactive object we can forward `as` as is in the composable. 😁

iamandrewluca: The current workaround to get rid of the warning is to use `markRaw`

image

iamandrewluca: I agree with you. Thanks!

PR #654, opened 2024-02-04
fix(Accordion, Collapsible): use separate id for content

dougwollison:
- aria-controls needs to point to the ID of AccordionContent, not that of AccordionToggle
- added contentId prop on AccordionItem context, update content/toggle components
- move v-bind on CollapsibleContent to end, allows custom ID passed from AccordionContent to override the rootContext one


zernonia:
Thanks for catching that @dougwollison ! However I realized since we are using `Collapsible` component underneath, we don't need to explicitly set the id for content.

Issue #669, opened 2024-02-07
[Bug]: Slider - Form submission is prevented when step is a decimal number


pohnean:
### Environment

```bash
Developement/Production OS: Mac OS 13.4
Node version: 18.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.3.3
Client OS: Windows 10 19043.1110
Browser: Chrome 90.0.4430.212
```

### Link to minimal reproduction

see screenshot

### Steps to reproduce

1. Put the slider in a <form> element.
2. Set step = 0.01
3. Submit the form

### Describe the bug

Slider is prevent the submission of form when step is a decimal number. An error is shown in the console: 'An invalid form control with name='slider' is not focusable.'

<img width="1148" alt="image" src="https://github.com/radix-vue/radix-vue/assets/846343/0482e681-7c7d-4450-9e7a-ca140157baf6">

### Expected behavior

The slider component should not prevent form submissions when step < 1

PR #670, opened 2024-02-07
fix(Slider): form submission prevented when step < 1 (#669)

pohnean:
Fixes issue #669.


zernonia (approved):
Thanks for reporting the issue and the relevant fix!!! 😁

PR #682, opened 2024-02-13
fix(Checkbox): check indeterminate state in click event handler

DamianGlowala:
When checkbox is in indeterminate state, we should account for it in the click handler and set the state to `true`.

Reference:
https://github.com/radix-ui/primitives/blob/c31c97274ff357aea99afe6c01c1c8c58b6356e0/packages/react/checkbox/src/Checkbox.tsx#L93C42-L93C57


zernonia (approved):
Thanks for the changes @DamianGlowala !

Issue #679, opened 2024-02-13
[Bug]: Ui jumping/shifting effect because the "ToastViewport" component doesn't have a zero height if a toast is shown inside.


DennisMaass:
### Environment

```bash
Node version: 20.11.0
Package manager: pnpm@8.14.3

Radix Vue version: 1.4.2
Vue version: 3.4.18
Nuxt version: 3.10.1
Nuxt mode: universal
Nuxt target: server

CSS framework: @nuxtjs/tailwindcss@6.11.3

Browser: Edge 121
```

### Link to minimal reproduction

https://stackblitz.com/run?file=src%2FApp.vue

### Steps to reproduce

Trigger the toast and inspect the DOM with the help of the dev tools. 
Select the element below the "add to calender" button:
`<div role="region" aria-label="Notifications (F8)" tabindex="-1" style="pointer-events: none;"<ol tabindex="-1" class="ToastViewport"></ol></div>`

Without a toast is has a 0 height, but with a visible toast the height is 21px (at least > 0px).
I my UI that leads to a jumping/shifting effect.

The reason is the bug in the "FocusProxy" component in line 16:
`style="position: 'fixed'"`

it should be
`style="position: fixed"`

or the object syntax
`:style="{position: 'fixed'}"`

### Describe the bug

 Ui jumping/shifting effect because the "ToastViewport" component doesn't have a zero height if an toast is shown.

### Expected behavior

It should have a zero height, like in react implementation:

https://www.radix-ui.com/primitives/docs/components/toast

zernonia:
You are right @DennisMaass ! This should be relatively quick and striaghtforward, you do want to contribute by creating a PR? 😁

PR #684, opened 2024-02-14
fix(Toast): correct style attribute syntax

DamianGlowala:
closes https://github.com/radix-vue/radix-vue/issues/679


zernonia (approved):
Thanks for another quick fix @DamianGlowala !

PR #686, opened 2024-02-14
fix: check isEmpty even without any searchterm

zernonia:


meirroth:
@zernonia Hi, hope you're well! Just came across this after noticing this issue, and found that you reverted this update later here https://github.com/unovue/reka-ui/commit/27fb34e0759f012e0db7fd4365d229f468ee10cb Do you know why? I would want the empty state to show when there are no options, even when there is no search term.

PR #692, opened 2024-02-17
fix: Missing types for namespaced and resolver

zernonia:


aleksey-hoffman:
Thanks mate

ferferga:
@zernonia After this PR, the types of the resolver doesn't work for me when using a prefix. For example, this pass typechecks:

```
RadixVueResolver()
```

But this does not:

```
RadixVueResolver({
    prefix: 'R'
})
```

Deleting ``node_modules/radix-vue/resolver.d.ts`` fixes it for me, with or without parameters.

Issue #633, opened 2024-01-21
[Feature]: Accordion should auto compute type based on provided `modelValue`


jd-solanki:
### Describe the feature

Hi 👋🏻 

This is my first issue in radix-vue, I was working on wrapper lib as I promised and started with Accordion and found that `type` prop is required and it's not default set to `single`. Moreover, We can infer it from the provided `modelValue` like below:
```ts
const accordionType = computed(() => Array.isArray(modelValue.value) ? 'multiple' : 'single')
```

_P.S. I guess props table for AccordionRoot is not complete because `modelValue` is missing but events has `@update:modelValue`_

### Additional information

- [ ] I intend to submit a PR for this feature.
- [ ] I have already implemented and/or tested this feature.

jd-solanki:
Another thing is shouldn't `defaultValue` prop be removed because if `v-model` initially (e.g. const val = ref('account')) has value then that should be the default value

PR #693, opened 2024-02-18
Refactor `useSingleOrMultiple`

onmax:
# What does it fix?

Sets the `type` prop automatically inferring the value from `v-model` or `defaultValue`. 

For example:

```html
<script setup lang="ts">
const str = ref('my-item')
</script>

<template>
<AccordionRoot v-model="str" /> <!-- `type` is single. If you use add `:type=multiple`, TS won't like it
</template>
```

This improves DX making easier to use.

I also take the oppportunity to create a new `SingleOrMultipleProps` which is the set of props that are used in `Accordion` and `RadioGroup`, making a unique codebase for both.

## Screen captures

This is an example of what the user will see:

| Type | DX |
|--------|--------|
| Setting the type as `single`| image|
| Setting the type as `multiple`| image| 

# Deprecated `Type`

I think the type `Type` is too broad and might create confusion. I checked the whole project, and it seems that this type is only used in the `Accordion` and `GroupRadio`. I think is better to be explicit in this case.

I marked as `@deprecated`, in the next major release, it could be removed.

## Solves #633

# TODO

- [ ] Update the docs removing the `required` prop (?). Not sure if this is the right thing. The `type` props is still required if `v-model` and `defaultValue` are not being used


onmax:
Any feedback is welcomed 😃

onmax:
I am using `generic` from Vue. Not sure if this will cause a problem, since it is a feature introduced in Vue 3.3:

https://github.com/radix-vue/radix-vue/pull/693/files#diff-3a4dd1dcd0ec33556b17ff15436567d35d42e4c602e47df80af5a3b915522a8bR61

zernonia (commented):
Once again nicely done @onmax ! will onhold this for 2.0 as it has refactoring and deprecation. 😁

At the mean time can you help update the JSDOC for `modelValue` and `defaultValue` so that it would cater for all 3 `AccordionRoot`, `ToggleGroupRoot`, `ToolbarToogleGroup`?

onmax:
> Once again nicely done @onmax ! will onhold this for 2.0 as it has refactoring and deprecation. 😁

But we could mark it as `@deprecated` in the `1.5.0` version? Then in `2.0.0`, we could remove it?

onmax:
> At the mean time can you help update the JSDOC for modelValue and defaultValue so that it would cater for all 3 AccordionRoot, ToggleGroupRoot, ToolbarToogleGroup?

I pushed #a910fce with new comments. I am having issues when I try to run `pnpm run build`, so I cannot preview the documentation page. 

Is there anything else missing to do?

Issue #690, opened 2024-02-16
[Bug]: DismissableLayer in Select component not receiving click when dismissing in a scrolled down viewport


olemarius:
### Environment

```bash
Developement/Production OS: Stackblitz
Node version: 18.18.0
Package manager: npm 10.2.3
Shadcn Vue version: 1.4.6
Vue version: 3.4.18
CSS framework: tailwindcss@3.4.1
Browser: Edge Version 121.0.2277.112 (Official build) (64-bit)
```

### Link to minimal reproduction

https://stackblitz.com/~/github.com/olemarius/shadcn-vue-select-bug-repro

### Steps to reproduce

Open stackblitz

First you will see the preview window probably without scroll
![image](https://github.com/radix-vue/radix-vue/assets/184536/dce5b153-b70d-48de-acd6-a35203b40389)

Now resize the browser so you get scroll like this, and scroll all the way to the bottom of the page 
![image](https://github.com/radix-vue/radix-vue/assets/184536/2f76b1dd-e946-4d71-8077-7252e047102e)

Now you'll see that clicking outside the Select wont happen unless you click in the upper section of the preview window. Note that scroll is disabled when it's open.
![image](https://github.com/radix-vue/radix-vue/assets/184536/b38f7bdf-59cf-4013-ba35-93212969c5ac)

### Describe the bug

Clicking outside select trigger mouseup, mousedown, click events but does not close the select. Clicking far enough up in the document will hit the dismissableLayer. So it seems that the DismissableLayer's position is off at scroll. 
![image](https://github.com/radix-vue/radix-vue/assets/184536/5620b6f7-a93d-41ba-81c2-a476e8099ba7)

### Expected behavior

Click anywhere outside select should close the dropdown.

MellKam:
The problem is inside those lines:

https://github.com/radix-vue/radix-vue/blob/2baaefd178c0413e4fdb5312c8c86a04dd148125/packages/radix-vue/src/DismissableLayer/utils.ts#L65-L66

We need to use `clientX` and `clientY` instead of `offsetX` and `offsetY`.

```ts
if (event.clientX > target.clientWidth || event.clientY > target.clientHeight)
	return
```

MellKam:
@romanhrynevych Hi! I checked and that seem to be your code. Can you please explain why do we need it? Is it possible, that we can we get an event on target with coords that are outside the target? I haven't found it in original radix-ui, so I'm just curious)

romanhrynevych:
[...]

PR #695, opened 2024-02-18
fix: invalid coords check in `usePointerDownOutside` composable

MellKam:
resolves: #690


zernonia (approved):
Thanks for the quick fix @MellKam !!

PR #709, opened 2024-02-23
fix(Combobox): Computed warning when opening items

### LINKED ISSUE #708: [Bug]: Combobox computed warning when opening
### Environment

```bash
Developement/Production OS: MacOS Sonoma 14.3.1
Node version: 18.17.1
Package manager: pnpm@8.15.3
Radix Vue version: 1.4.7
Nuxt version: 3.10.3
```

The example from the docs: https://www.radix-vue.com/components/combobox.html

### Steps to reproduce

Click the trigger of the combobox or start typing on the input

### Describe the bug

Whenever the `ComboboxContent` gets displayed, we get a console warning saying that there's a computed property that is dirty:

```
[Vue warn] Computed is still dirty after getter evaluation, likely because a computed is mutating its own dependency in its getter. State mutations in computed getters should be avoided.
```

I know it's just a warning, but maybe there is something wrong or that can be improved internally to not get the warning. I tried to find the issue myself but I couldn't find anything obvious.

This can be replicated even on the example from the docs in Stackblitz, but it also happens when using it on other apps locally.

### Expected behavior

There should be no warning at all

https://github.com/radix-vue/radix-vue/assets/47041342/611548be-e26a-4bde-91bc-9b58e86a9936

ChrisGV04:
### 🔗 Linked issue

Resolves #708 

### ❓ Type of change

### 📚 Description

This PR aims to remove the warning caused when opening the `ComboboxContent`, which says the following:

```
[Vue warn] Computed is still dirty after getter evaluation, likely because a computed is mutating its own dependency in its getter. State mutations in computed getters should be avoided.
```

This warning seems to occur at the `ComboboxRoot` component when reading the value of the `options` property inside of a regular `computed`. The `options` property uses the VueUse utility `computedWithControl` to only change the value when `itemMapSize` changes. 

However, as mentioned in the following issue from VueUse (https://github.com/vueuse/vueuse/issues/3794#issuecomment-1951369704) it is recommended to instead use a `watch` to perform the side effects, instead of `computedWithControl`.

This PR replaces `computedWithControl` with a simple `watch` and `ref` implementation to achieve the same goal. I compared the behavior of both implementations and they where practically identical, but now the warning is gone.

I will leave the PR open in case you have a better suggestion!

### 📝 Checklist

### DIFF
--- a/packages/radix-vue/src/Combobox/ComboboxRoot.vue
+++ b/packages/radix-vue/src/Combobox/ComboboxRoot.vue
@@ -72,7 +72,7 @@ export interface ComboboxRootProps<T = AcceptableValue> extends PrimitiveProps {
 import { computed, nextTick, ref, toRefs, watch } from 'vue'
 import { PopperRoot } from '@/Popper'
 import { Primitive } from '@/Primitive'
-import { computedWithControl, useVModel } from '@vueuse/core'
+import { useVModel } from '@vueuse/core'
 import { VisuallyHiddenInput } from '@/VisuallyHidden'
 import isEqual from 'fast-deep-equal'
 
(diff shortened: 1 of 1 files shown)


ChrisGV04:
I also made a small change to the default filter, which was unsafe.

The default filter should only compare strings, but the current implementation only checked if the first option was a string and trusted the rest would also be strings.

Now, the filter checks each item so that it only searches string values.

PR #711, opened 2024-02-24
feat: Splitter

zernonia:


packages/radix-vue/src/Splitter/SplitterResizeHandle.vue
```diff
@@ -0,0 +1,156 @@
...
+    :data-resize-handle-state="state"
+    :data-panel-resize-handle-enabled="!disabled"
+    :data-panel-resize-handle-id=" resizeHandleId"
+    @blur="isFocused = false"
+    @focus="isFocused = false"
+  >
+    <slot />
+  </Primitive>
+</template>
+./utils/registry./utils/registry
```
sadeghbarati, line 156: this string is outside of template

```suggestion

```

docs/content/components/splitter.md
```diff
@@ -0,0 +1,284 @@
...
+<template>
+  <SplitterGroup auto-save-id="any-id">
+    …
+  </SplitterGroup>
+</template>
+```
+ 
+### Persist layout with SSR
+
+By default, Splitter uses `localStorage` to persist layouts. With server rendering, this can cause a flicker when the default layout (rendered on the server) is replaced with the persisted layout (in localStorage). The way to avoid this flicker is to also persist the layout with a cookie like so:
```
DamianGlowala, line 151:
```suggestion
By default, Splitter uses `localStorage` to persist layouts. With server rendering, this can cause a flicker when the default layout (rendered on the server) is replaced with the persisted layout (in `localStorage`). The way to avoid this flicker is to also persist the layout with a cookie like so:
```

docs/content/components/splitter.md
```diff
@@ -0,0 +1,284 @@
...
+    <SplitterResizeHandle />
+    <SplitterPanel :default-size="layout[1]">
+      …
+    </SplitterPanel>
+  </SplitterGroup>
+</template>
+```
+ 
+
+### Collapse/Expand programatically
```
DamianGlowala, line 173:
```suggestion
### Collapse/Expand programmatically
```

docs/content/components/splitter.md
```diff
@@ -0,0 +1,284 @@
...
+       …
+     </SplitterPanel>
+   </SplitterGroup>
+</template>
+```
+ 
+
+### SSR
+
+Splitter component heavily rely on unique `id`, however for Vue<3.4 we don't have a reliable way of generating [ssr-friendly id](https://github.com/vuejs/rfcs/discussions/557).
```
DamianGlowala, line 229:
```suggestion
Splitter component heavily relies on unique `id`, however for Vue<3.4 we don't have a reliable way of generating [SSR-friendly `id`](https://github.com/vuejs/rfcs/discussions/557).
```

docs/content/components/splitter.md
```diff
@@ -0,0 +1,284 @@
...
+
+Thus, if you are using Nuxt or other SSR framework, you are required to manually add the `id` for all Splitter component. Alternatively, you can wrap the component with `<ClientOnly>`.
```
DamianGlowala, line 231:
```suggestion
Thus, if you are using Nuxt or other SSR framework, you are required to manually add the `id` for all Splitter components. Alternatively, you can wrap the component with `<ClientOnly>`.
```

docs/content/components/splitter.md
```diff
@@ -0,0 +1,284 @@
...
+      …
+    </SplitterPanel>
+  </SplitterGroup>
+</template>
+```
+
+
+## Accessibility
+
+Adheres to the [Window Splitter WAI-ARIA design pattern](https://www.w3.org/WAI/ARIA/apg/patterns/windowsplitter/).
```
DamianGlowala, line 252:
```suggestion
Adheres to the [Window Splitter WAI-ARIA design pattern](https://www.w3.org/WAI/ARIA/apg/patterns/windowsplitter).
```

DamianGlowala (changes requested):

PR #721, opened 2024-03-04
feat: CSP nonce prop

sadeghbarati:


zernonia:
let's apply the same for `ComboboxViewport` too 😁

PR #723, opened 2024-03-04
fix(Popper): disable pointer events when hidden

sadeghbarati:


zernonia:
@sadeghbarati  can you share the relevant PR from Radix UI? Interested to see the problem user having

sadeghbarati:
https://github.com/radix-ui/primitives/pull/2745

Issue #733, opened 2024-03-07
[Bug]: Tooltip's `aria-describedby` is empty on first hover.


remonke:
### Environment

```bash
Developement/Production OS: Windows 10 19043.1110
Node version: 16.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.3.3
Client OS: Windows 10 19043.1110
Browser: Chrome 90.0.4430.212
```

### Link to minimal reproduction

https://stackblitz.com/edit/vitejs-vite-mfk4up?file=src%2FApp.vue%2Cpackage.json

### Steps to reproduce

1. Use a tooltip.
2. Open devtools and find a button
3. Hover over the button
4. No `aria-describedby`

### Describe the bug

Initial hover

![image](https://github.com/radix-vue/radix-vue/assets/102010617/2537855d-fd78-4bf9-8c9b-8a64d2bd0790)

Every subsequent hover

![image](https://github.com/radix-vue/radix-vue/assets/102010617/103cdf83-af85-4c78-bd55-9d912b4beb5d)

remonke:
Okay sorry, I forgot tooltips can be persistent, and are not just shown on user interaction

PR #734, opened 2024-03-07
fix(tooltip): fix aria-describedby

remonke:
https://github.com/radix-vue/radix-vue/issues/733


remonke:
Also, I wonder if it would be worth refactoring the code, because it looks all the components mutate their context ids. I think it would be better to initialize such things at the root, as it would avoid such behaviors.

zernonia:
Thanks for the PR @remonke . `useId` was previously generated from `Root`. However it is not compatible with the current `useId` from Nuxt, thus I moved it to component.

Let me help ammend abit ya!

PR #740, opened 2024-03-08
docs: i18n docs, resolves #719

### LINKED ISSUE #719: [Feature]: Select, Calendar, DropDown by default set dir to ltr even when html dir is rtl
### Reproduction

https://codesandbox.io/p/devbox/shadcn-vue-selectdemo-forked-gxptxq

### Describe the bug

I set global dir for html to be `dir='rtl'`
Select, calendar, dropdown seems to have a default `dir='ltr'` so everytime i should force them to be `dir='auto'` so they can have parent HTML dir
it seems like auto is not a valid value in radix components 
so we need to support it in shadcn components as default value

### System Info

```bash
System:
    OS: Windows 11 10.0.22631
    CPU: (32) x64 AMD Ryzen 9 7950X 16-Core Processor
    Memory: 6.19 GB / 31.16 GB
  Binaries:
    Node: 18.18.2 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 9.8.1 - C:\Program Files\nodejs\npm.CMD
    pnpm: 8.14.3 - ~\AppData\Local\pnpm\pnpm.CMD
  Browsers:
    Edge: Chromium (121.0.2277.128)
    Internet Explorer: 11.0.22621.1
  npmPackages:
    @vueuse/core: ^10.7.2 => 10.7.2
    radix-vue: ^1.4.7 => 1.4.7
    vue: ^3.4.19 => 3.4.19
```

### Contributes

- [X] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests

Saeid-Za:


zernonia (approved):
Thanks for the docs @Saeid-Za !! Really appreciate it! 💚

PR #759, opened 2024-03-16
fix: making toggle pressed  logic a computed property

### LINKED ISSUE #754: [Bug]: ToggleGroupItem data-state attribute don't change when type="multiple"
### Environment

```bash
Node version: 16.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Vue version: 3.4.21
Nuxt version: 3.10.3
Nuxt mode: universal
Nuxt target: server
CSS framework: unocss@0.58.3
Browser: Chrome 90.0.4430.212
```

https://stackblitz.com/edit/vitejs-vite-enahac?file=src%2FApp.vue

### Steps to reproduce

1. Import `ToggleGroupRoot`, `ToggleGroupItem` 
2. Set `type="multiple"` on the Root
3. Check how the `[data-state]` attribute behave on the Items

### Describe the bug

The `[data-state]` attribute of the `ToggleGroupItem` component is always in the `off` state when using `type="multiple"`.

I encountered this error after upgrading from `v1.4.6` to `v1.5.1`.

### Expected behavior

The `[data-state]` attribute on `ToggleGroupItem` components toggles between "on" and "off" on interaction.

_No response_

Saeid-Za:
resolves #754
But another problem remains. providing a default value in `useSingleOrMultipleValue` for `modelValue` never let's the model to be undefined in `single` mode. so in that mode, we can never uncheck all `toggle` items.
I really don't know if this is intentional or not. 

Try to comment line `74` of `useSingleOrMultipleValue.ts` file and then try to uncheck all toggle values in the reproduction provided in #754

### DIFF
--- a/packages/radix-vue/src/ToggleGroup/ToggleGroup.test.ts
+++ b/packages/radix-vue/src/ToggleGroup/ToggleGroup.test.ts
@@ -3,13 +3,19 @@ import ToggleGroup from './story/_ToggleGroup.vue'
 import type { DOMWrapper, VueWrapper } from '@vue/test-utils'
 import { mount } from '@vue/test-utils'
 import { axe } from 'vitest-axe'
+import { nextTick, ref } from 'vue'
 
 describe('given default Toggle Group', () => {
   let wrapper: VueWrapper<InstanceType<typeof ToggleGroup>>
   let triggers: DOMWrapper<HTMLButtonElement>[]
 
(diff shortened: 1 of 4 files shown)


zernonia:
Thanks for the PR @Saeid-Za , and picking up on the `single` mode issue. Looking into it 😁

PR #773, opened 2024-03-19
chore: useForwardPropsEmits doc fix

dermitzos:
This PR fixes a the wrong TS types in the `useForwardPropsEmits` docs example


zernonia:
Thanks for the PR @dermitzos !  😁

Issue #774, opened 2024-03-19
[Feature]: Add Tooltip flag to use only focus from the keyboard.


remonke:
### Describe the feature

VueUse uses the following trick to match only `:focus-visible` when focusing on an element:

```vue
useEventListener(targetElement, 'focus', (event) => {
   if (!focusVisible || (event.target as HTMLElement).matches?.(':focus-visible'))
     innerFocused.value = true
 })
 ```
 
I suggest making this an optional flag, and using regular focus by default (this would be a breaking change I believe, and it has only 94% support according to caniuse.com). This would slightly improve the UX, as the tooltip currently shows up when say I focused on a button by mouse and focused back from a different tab, or closed a dialog opened by a button with a tooltip.

[Here's an example](https://stackblitz.com/edit/nuxt-starter-wgak7r?file=nuxt.config.ts%2Capp.vue%2Ccomponents%2Ftooltip-button.vue) - try to focus with a mouse on each of the button and then switch browser tabs. The one without `focus-visible` gets triggered. I personally find this not ideal...

I think I can implement this on my own, however I don't know if I should implement this behavior myself or use the `useFocus` hook from VueUse, as VueUse is already included in this library. 

Please let me know what do you think about this. 😊

### Additional information

- [X] I intend to submit a PR for this feature.
- [ ] I have already implemented and/or tested this feature.

zernonia:
Thanks for the issue and PR @remonke ! I think this would be a good addition to the Tooltip! Some dev might want to have `keyboard` only tooltip! 👍

PR #775, opened 2024-03-19
feat: added keyboard-focus-only prop to Tooltip

remonke:
#774

### DIFF
--- a/packages/radix-vue/src/Tooltip/TooltipProvider.vue
+++ b/packages/radix-vue/src/Tooltip/TooltipProvider.vue
@@ -39,9 +39,10 @@ export interface TooltipProviderProps {
    */
   disableClosingTrigger?: boolean
   /**
-   * Uses the `:focus-visible` selector to prevent the tooltip from opening
-   * if the focus did not come from the keyboard. This is useful if you want
-   * to avoid opening it when switching browser tabs or closing a dialog.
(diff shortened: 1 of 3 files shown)


packages/radix-vue/src/Tooltip/TooltipProvider.vue
```diff
@@ -37,6 +38,14 @@ export interface TooltipProviderProps {
...
    */
   disableClosingTrigger?: boolean
+  /**
+   * Prevent the tooltip from opening if the focus did not come from
+   * the keyboard by using the `:focus-visible` selector.
+   * This is useful if you want to avoid opening it when switching
+   * browser tabs or closing a dialog.
+   * @defaultValue false
+   */
+  keyboardFocusOnly?: boolean
```
zernonia, line 48: `keyboardFocusOnly` naming might confuse user with changing the tooltip behavior to only "keyboard focus".
Maybe something like `disableNonKeyboardFocus`?

remonke: Updated the prop name to `ignoreNonKeyboardFocus`.

zernonia: Lovely!! Thanks @remonke ! 😁

PR #778, opened 2024-03-20
feat(navigation-menu): multiple triggers on navigation menu

### LINKED ISSUE #565: [Feature]: Navigation Menu only trigger when "clicked"
### Describe the feature

Allow the [navigation menu component](https://www.radix-vue.com/components/navigation-menu) to be triggered on click

### Additional information

Chrtyaka:
Hi 👋🏻 Closes #565 

### NavigationMenuRoot.vue
- add props `openOnClick`, `openOnHover`

Both props a optional, and `true` by default. It's not a breaking change.

Actually, I had 3 solutions in my mind:
1. Create prop `triggers` which can be array of strings, i.e `['hover', 'click']`. But it's too complicated. 
2. Create props trigger, with values `click` or `hover` like [element-plus popover](https://element-plus.org/en-US/component/popover.html#attributes).  But it's not flexible. We can use only one trigger at the time. It would be a breaking change
3. Create two props for disable hover o click. I choose this one 😄 Thanks to [vuetify](https://vuetifyjs.com/en/api/v-menu/#links)

### Tests
add test for both props. Includes trigger on hover
### Docs
update docs for both props

### Default

https://github.com/radix-vue/radix-vue/assets/20302368/c04c207e-7d94-4ee2-9bb6-3b5964a12392

### openOnHover -- open only by hover

https://github.com/radix-vue/radix-vue/assets/20302368/34a63919-761e-49af-be5a-144624d7d5dd

### openOnClick -- open only by click

https://github.com/radix-vue/radix-vue/assets/20302368/069a546f-580c-44cd-9c4a-3296b0e31bff

### DIFF
--- a/docs/content/components/navigation-menu.md
+++ b/docs/content/components/navigation-menu.md
@@ -439,7 +439,7 @@ import { NavigationMenuItem, NavigationMenuList, NavigationMenuRoot } from 'radi
             Home
           </RouterLink>
           <NavigationMenuLink />
-        </navigationmenulink>
+        </NavigationMenuLink>
       </NavigationMenuItem>
       <NavigationMenuItem>
         <NavigationMenuLink :as="RouterLink" to="/about">
--- a/docs/content/meta/NavigationMenuRoot.md
+++ b/docs/content/meta/NavigationMenuRoot.md
@@ -52,6 +52,20 @@
     'type': 'number',
     'required': false,
     'default': '300'
+  },
+  {
+    'name': 'openOnClick',
+    'description': '<p> Should open content by click on trigger </p>\n',
+    'type': 'boolean',
+    'required': false,
+    'default': true
+  },
+  {
+    'name': 'openOnHover',
+    'description': '<p> Should open content by hover on trigger </p>\n',
+    'type': 'boolean',
+    'required': false,
+    'default': true
   }
 ]" />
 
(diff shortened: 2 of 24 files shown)


zernonia:
Yoo @Chrtyaka ! Thanks for the PR!!!

May I ask why the first solution is complicated ya? I kinda prefer that tho 😂

Chrtyaka:
Yeah, at first I wrote solution with arrays. But I looked in the code, and I didn't like it 😅 

It's just my personal experience. Two boolean props are easier for me to operate than an array. 
But if you think that array is the better solution, I can rewrite it. 

No problem👍🏻

Chrtyaka:
Yes, this is a good point. I think something like `disableClickTrigger` or `disableHoverTrigger` would work.

Because `disableOpenOnClick` and `disableOpenOnHover` are too many symbols 😅

Would that be okay?

zernonia:
That sounds perfect! 🤩

Chrtyaka:
Done!

Props names changed, disabled by default

zernonia:
Nicely done @Chrtyaka ! I've update the `disableClickTrigger` abit to make sure we only disable based on `pointerType: 'mouse'`. We want to make sure accessibility for the component 😁

yoni-noma:
It this going to work on the react version as well? :)

PR #761, opened 2024-03-17
feat(DatePicker): update date library package

epr3:
Closes #742.

Hello,

I forked and refactored Adobe's `@internationalized/date` package and converted all ES6 classes into plain objects and converted the original class methods into functions. All functions return new calendar objects and are immutable.

Then, I've refactored the whole implementation of the `Calendar`, `DateField` and `DatePicker` components to use this library and removed some redundant functions, updated the docs and added localised stories for some of the calendars in Histoire. 

It should fix all errors that were caused by Vue changing the `this` context from the `DateValue` class to it's own `Proxy` class (plus any date jumps to the Gregorian Calendar and back causing dates to desync).

This is the library: https://www.npmjs.com/package/flat-internationalized-date

If you guys want, I can relinquish the package ownership to the `radix-vue` org in Github and continue maintenance under `radix-vue`.

I'm looking forward to any feedback and improvements you suggest.

Thanks!

CC-ing @sadeghbarati for the Persian Calendar fixes.

PR #784, opened 2024-03-24
chore: make vue peer dependency

MellKam:
Idk why we haven't done this before, but I think we need to make vue a peer dependency. You basically can't use radix-vue without vue. It's a general thing among outher libraries, for example [headless-ui](https://github.com/tailwindlabs/headlessui/blob/000e0c01927f38a70f2197df6ee9a388f7a72d99/packages/%40headlessui-vue/package.json#L44-L51) or [ark-ui](https://github.com/chakra-ui/ark/blob/4aad0c95cf7e3fdecc85ec083e16fd586cf7c3ee/packages/frameworks/vue/package.json#L98-L102)

I also moved this part of the dependencies from "devDependencies" to just "dependencies" because they are actually used at runtime.

```
"@vueuse/components": "^10.5.0",
"@vueuse/core": "^10.5.0",
"@vueuse/shared": "^10.5.0",
"aria-hidden": "^1.2.3",
"defu": "^6.1.4",
```

### DIFF
--- a/packages/radix-vue/package.json
+++ b/packages/radix-vue/package.json
@@ -78,7 +78,6 @@
   },
   "devDependencies": {
     "@iconify/vue": "^4.1.1",
-    "@internationalized/date": "^3.5.2",
     "@testing-library/dom": "^9.3.3",
     "@testing-library/jest-dom": "^6.2.0",
     "@testing-library/user-event": "^14.5.1",
--- a/packages/radix-vue/package.json
+++ b/packages/radix-vue/package.json
@@ -56,15 +56,32 @@
     "test": "vitest",
     "pub:release": "pnpm publish --access public"
   },
+  "peerDependencies": {
+    "@internationalized/date": "^3.5.2",
+    "vue": "^3.2.0"
+  },
+  "peerDependenciesMeta": {
+    "vue": {
+      "optional": false
+    },
+    "@internationalized/date": {
+      "optional": true
+    }
+  },
   "dependencies": {
     "@floating-ui/dom": "^1.5.4",
     "@floating-ui/vue": "^1.0.4",
-    "@internationalized/date": "^3.5.2",
+    "@vueuse/components": "^10.5.0",
+    "@vueuse/core": "^10.5.0",
+    "@vueuse/shared": "^10.5.0",
+    "aria-hidden": "^1.2.3",
+    "defu": "^6.1.4",
     "fast-deep-equal": "^3.1.3",
     "nanoid": "^5.0.6"
   },
   "devDependencies": {
     "@iconify/vue": "^4.1.1",
+    "@internationalized/date": "^3.5.2",
     "@testing-library/dom": "^9.3.3",
     "@testing-library/jest-dom": "^6.2.0",
     "@testing-library/user-event": "^14.5.1",
@@ -77,11 +94,6 @@
     "@vue/shared": "^3.4.6",
     "@vue/test-utils": "^2.4.1",
     "@vue/tsconfig": "^0.4.0",
-    "@vueuse/components": "^10.5.0",
-    "@vueuse/core": "^10.5.0",
-    "@vueuse/shared": "^10.5.0",
-    "aria-hidden": "^1.2.3",
-    "defu": "^6.1.4",
     "dts-bundle": "^0.7.3",
     "jsdom": "^23.2.0",
     "typescript": "^5.3.3",
--- a/packages/radix-vue/package.json
+++ b/packages/radix-vue/package.json
@@ -57,20 +57,17 @@
     "pub:release": "pnpm publish --access public"
   },
   "peerDependencies": {
-    "@internationalized/date": "^3.5.2",
     "vue": "^3.2.0"
   },
   "peerDependenciesMeta": {
     "vue": {
       "optional": false
-    },
-    "@internationalized/date": {
-      "optional": true
     }
   },
   "dependencies": {
     "@floating-ui/dom": "^1.5.4",
     "@floating-ui/vue": "^1.0.4",
+    "@internationalized/date": "^3.5.2",
     "@vueuse/components": "^10.5.0",
     "@vueuse/core": "^10.5.0",
     "@vueuse/shared": "^10.5.0",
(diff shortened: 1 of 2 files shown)


MellKam:
@zernonia I also have a question. What is the purpose of "dts-bundle" pacakge? Because I couldn't find any usage of it in repository. Seems like we don't use it. 

https://github.com/radix-vue/radix-vue/blob/ed730548c01a7e906d8b695f226f012a9e27a9a2/packages/radix-vue/package.json#L85

packages/radix-vue/package.json
```diff
@@ -57,20 +57,17 @@
...
       "optional": false
-    },
-    "@internationalized/date": {
-      "optional": true
     }
   },
   "dependencies": {
     "@floating-ui/dom": "^1.5.4",
     "@floating-ui/vue": "^1.0.4",
+    "@internationalized/date": "^3.5.2",
```
sadeghbarati, line 70: flat-internationalized-date

sadeghbarati:
Why not move others like `@vueuse` to peers too?

I think I asked this question back then but I don't remember the answer 


---

https://nodejs.org/en/blog/npm/peer-dependencies

> Even for plugins that do have such direct dependencies, probably due to the host package supplying utility APIs, specifying the dependency in the plugin's package.json would result in a dependency tree with multiple copies of the host package

MellKam:
@sadeghbarati Yeah, maybe we should. I was thinking about moving `@internationalized/date` to optional peer deps, because we recommend installing it as a separate dependency in the documentation. But at least let's do it with vue.

image

MellKam:
Just noticed that `@vueuse/components` dependency is never used in codebase XDD

https://github.com/radix-vue/radix-vue/blob/ed730548c01a7e906d8b695f226f012a9e27a9a2/packages/radix-vue/package.json#L80

`defu` and `nanoid` are used only in a single place

https://github.com/radix-vue/radix-vue/blob/ed730548c01a7e906d8b695f226f012a9e27a9a2/packages/radix-vue/src/shared/useBodyScrollLock.ts#L59-L62

↑ this can be easily done without libraries

https://github.com/radix-vue/radix-vue/blob/ed730548c01a7e906d8b695f226f012a9e27a9a2/packages/radix-vue/src/shared/useBodyScrollLock.ts#L100

↑ and this can be changed to something like `Math.random().toString(36).slice(6)` if we don't need it to be truly random

zernonia:
> @zernonia I also have a question. What is the purpose of "dts-bundle" pacakge? Because I couldn't find any usage of it in repository. Seems like we don't use it.
> 
> https://github.com/radix-vue/radix-vue/blob/ed730548c01a7e906d8b695f226f012a9e27a9a2/packages/radix-vue/package.json#L85

You are right. This might be previous attempt to generate `d.ts`, but no longer required. We can remove this

zernonia:
> Just noticed that `@vueuse/components` dependency is never used in codebase XDD
> 
> https://github.com/radix-vue/radix-vue/blob/ed730548c01a7e906d8b695f226f012a9e27a9a2/packages/radix-vue/package.json#L80
> 
> `defu` and `nanoid` are used only in a single place
> 
> https://github.com/radix-vue/radix-vue/blob/ed730548c01a7e906d8b695f226f012a9e27a9a2/packages/radix-vue/src/shared/useBodyScrollLock.ts#L59-L62
> 
> ↑ this can be easily done without libraries
> 
> https://github.com/radix-vue/radix-vue/blob/ed730548c01a7e906d8b695f226f012a9e27a9a2/packages/radix-vue/src/shared/useBodyScrollLock.ts#L100
> 
> ↑ and this can be changed to something like `Math.random().toString(36).slice(6)` if we don't need it to be truly random

Yeah we can remove `@vueuse/components`.


As for `defu` and `nano`, seeing that both were relatively lightweight, and the implementationis super straightforward, I think it's alright to just keep them. WDYT?

zernonia:
> @sadeghbarati Yeah, maybe we should. I was thinking about moving `@internationalized/date` to optional peer deps, because we recommend installing it as a separate dependency in the documentation. But at least let's do it with vue.
 

I agree! Let's do that

packages/radix-vue/package.json
```diff
@@ -56,15 +56,31 @@
     "test": "vitest",
     "pub:release": "pnpm publish --access public"
   },
+  "peerDependencies": {
+    "@internationalized/date": "^3.5.2",
```
sadeghbarati, line 60: since in the next release we rely on `flat-internationalized-date` isn't better to change it to `flat-internationalized-date`?

packages/radix-vue/package.json
```diff
@@ -56,15 +56,31 @@
...
+    "vue": ">= 3.2.0"
+  },
+  "peerDependenciesMeta": {
+    "@internationalized/date": {
+      "optional": true
+    },
+    "vue": {
+      "optional": false
+    }
```
sadeghbarati, line 69: I think adding vue in peerDependencies is enough cause the optional property is false by default

MellKam:
I don't know if we should add @vueuse/core to the peer dependencies, because if we do, the library user will have to install it along with radix-vue. And this may not be the case for every project.

packages/radix-vue/package.json
MellKam: I dont think that this is correct thing to do in this PR. You can do it in https://github.com/radix-vue/radix-vue/pull/761

packages/radix-vue/package.json
MellKam: yep, you are right. I removed it

MellKam:
@zernonia Can you please clarify why in the docs you recommend to install `npm install -D @internationalized/date` as dev dependency? Because I think it should probably be not only for development

sadeghbarati:
> clarify why in the docs you recommend to install npm install -D @internationalized/date as dev dependency?

Cause it's in `radix-vue` **dependencies** and we install in **devDependencies** to get _types and autocomplete_ in development 
But after this PR we install it in dependencies ourselves

MellKam:
@sadeghbarati gotcha. So maybe I should update those snippets to `npm install @internationalized/date`?

zernonia:
Nicely done @MellKam ! LGTM! 💚
Thanks for the initiative ya! 😁

zernonia (approved):

packages/radix-vue/package.json
zernonia: Yup it's better to handle that in relevant MR 😁

PR #788, opened 2024-03-25
chore(docs): remove `ohmyfetch` dependency and use `tsx` instead of `esno`

MellKam:
- `ohmyfethc` is deprecated package and it is used only once in codebase. There is no real reason why we can't use just `fetch`
- `esno` is currently just an alias for `tsx`, so I mean.. why to install one more package


zernonia (approved):
Thanks @MellKam for the cleanup ! This contributors script was copied over from vitest originally 😂

PR #793, opened 2024-03-29
fix(PinInput): remove trailing empty strings from `modelValue`

DamianGlowala:
Currently, trailing empty strings make it more difficult to validate the input with libraries like FormKit with one particular scenario in mind, where all strings are empty and there is a `required` rule applied.

Instead of `['', '', '']` (which the `required` rule would accept), this PR changes it to `[]`, which this rule correctly reports as invalid.


zernonia (approved):
I had the similar solution in mind when you mentioned the problem! Nicely done @DamianGlowala and LGTM!

Issue #797, opened 2024-03-30
[Bug]: Cannot find package '@internationalized/date' radix-vue@1.6.0


MellKam:
### Describe the bug

After the version 1.6.0 when you try to use radix-vue, without `@internationalized/date` package installed, it will throw an error. Although the package is listed as an optional peer dependency, it gets imported anyway when bundled into the single index.js file, causing the error.

We probably need to move the `@internationalized/date` package back to be just dependency.

Looking ahead, it would make sense to split the radix-vue package into a sub-packages for each component (as it does radix-ui). That would resolve this PR https://github.com/radix-vue/radix-vue/pull/469 and allow us to install only dependencies that we actually need. 

### Context & Screenshots (if applicable)

![image](https://github.com/radix-vue/radix-vue/assets/51422045/f83e7087-0ce3-43fc-8ac9-3f4b5b9a2677)

zernonia:
I don't really like splitting components into their own packages.. There's many depedency versioning issue, as you may observe in `radix-ui`. I prefer not to let dev worry about what version of `DismissableLayer` is compatible with which version of `Dialog`. This also ensure that we won't have much compatibility issue.

As for the case of `@internationalized/date`, I think we can future improve on this, maybe a different way to bundle the lib? 
For now yeah we can add it back as deps to unblock `1.6.0`

sadeghbarati:
@MellKam, hi, can you share your project?

having a lot of deps for package is not good imo 

peerDependency means it's user duty to install the package (it's not included in the radix-vue package)

Having it in dependencies means for every release the make we should do it too, which is not good for package maintainers

MellKam:
@sadeghbarati No, sorry. I can't share it. It's my work project. But I think you can easily reproduce it. Since we are bundling all the components into one index.js file, the import for `@internationalized/date` is also included in that file. So if you decide not to use any date components and do not install this dependency, you will see an error.

PR #798, opened 2024-03-30
fix: remove `@internationalized/date` from list of peer dependencies and make it just dependency

MellKam:
resolves: #797

### DIFF
--- a/packages/radix-vue/package.json
+++ b/packages/radix-vue/package.json
@@ -57,17 +57,12 @@
     "pub:release": "pnpm publish --access public"
   },
   "peerDependencies": {
-    "@internationalized/date": "^3.5.2",
     "vue": ">= 3.2.0"
-  },
-  "peerDependenciesMeta": {
-    "@internationalized/date": {
-      "optional": true
-    }
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


packages/radix-vue/package.json
```diff
@@ -57,17 +57,12 @@
...
   },
-  "peerDependenciesMeta": {
-    "@internationalized/date": {
-      "optional": true
-    }
-  },
   "dependencies": {
     "@floating-ui/dom": "^1.5.4",
     "@floating-ui/vue": "^1.0.4",
+    "@internationalized/date": "^3.5.2",
```
zernonia, line 65: I saw @epr3 updated to use `"@internationalized/date": "^3.5.1"` in devDependencies, perhaps we should use the same here?

MellKam: Do you want to change it to 3.5.1? We don't need it to be dev dependency now. What do you suggest? 3.5.2 is the latest version, should be fine

epr3: Yes, use the latest version.  👍

zernonia: Cool! Just wanna make sure the version is the same as in the dev version.

Issue #800, opened 2024-03-31
[Bug]: Out animation doesn't work for `Select` and `DropdownMenu` with `:modal="false"`


MellKam:
### Link to minimal reproduction

https://stackblitz.com/edit/8sgcno?file=src%2FApp.vue

### Steps to reproduce

I used playground example from shadcn-vue and modified it with `:modal="false"` on the root component. For some reason, enabling this prevents out animation from running. 

This problem happened after the 1.6.0 release.

MellKam:
I think the problem is related to this commit - 12965aa5ada3bd224bce66342758316b5cf19538

ChrisGV04:
This issue exists for many components that have overlay behavior. The Vue `Transition` is not being applied

PR #801, opened 2024-03-31
fix: slot prevents out transition for non-modal dropdown and select components

MellKam:
resolves: #800 

The main change that brought the out transition back was:

```ts
if (childrens.length === 1)
	return cloned
```

Basically unwrap the vnodes array if it only has one element. Idk why this is fixed, but it works XD. I'd like others to look at this, because I'm not really sure about it. 

I should probably write a story to showcase this, but I'm lazy right now. I'll get back to it tomorrow.

### DIFF
--- a/.histoire/package.json
+++ b/.histoire/package.json
@@ -20,7 +20,8 @@
     "eslint": "^8.56.0",
     "histoire": "^0.17.14",
     "postcss": "^8.4.33",
-    "tailwindcss": "^3.4.1"
+    "tailwindcss": "^3.4.1",
+    "tailwindcss-animate": "1.0.7"
   },
   "peerDependencies": {
     "@vitejs/plugin-vue": "*",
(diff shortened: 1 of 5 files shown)


zernonia:
Tested out and indeed was the array of vnodes causing the out animation to be missing.
Thanks for the fix and improvement here @MellKam !  That's a brilliant solution! 💚

Not sure what kind of story can we write here, and writing test for animation/transition part is still something I need to figure out lol 😂

zernonia (approved):

MellKam:
I updated the DropdownMenu story with out animation using `tailwindcss-animate` plugin. 

https://github.com/radix-vue/radix-vue/assets/51422045/15e84819-b60a-4514-af7a-f2f3c561fd6a

I have also tried to come up with some test that would check that animation is going as it should, but no success.

MellKam:
@zernonia Why do we have the `require` in the tailwind configs? Is this necessary? If not, I would like to open the separate PR and update all configs to esm only.

zernonia:
Nice @MellKam ! I thought of adding the `tailwind-animate` plugin to histoire too!
We can change `tailwind.config` to `esm` version 😁

Issue #802, opened 2024-03-31
[Bug]: HoverCardTrigger is untappable on mobile


remonke:
### Environment

```bash
Developement/Production OS: Windows 10 19043.1110
Node version: 16.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.3.3
Client OS: Windows 10 19043.1110
Browser: Chrome 90.0.4430.212
```

### Link to minimal reproduction

https://stackblitz.com/edit/nuxt-starter-kklahm?file=app.vue

### Steps to reproduce

1. Create a HoverCard
2. Try to tap on it with a touch pointer

### Describe the bug

A user cannot press on a link with a HoverCard.

### Context & Screenshots (if applicable)

[Screencast_20240401_011302.webm](https://github.com/radix-vue/radix-vue/assets/102010617/05d8dd99-e1b3-443d-84dc-f3dee3a3ffa6)

zernonia:
This was by designed as mentioned [here](https://github.com/radix-ui/primitives/issues/955#issuecomment-960610209)
As this component only cater for "Hover" interaction, maybe you should consider using `Popover` instead.

remonke:
But it's not like I can't open the hover card (which I can), but I can't simply tap on the link on mobile.

remonke:
Okay I did some digging and appearently Vue registers `touchstart` listener as a non-passive listener, while React uses a passive listener. As a result, `e.preventDefault()` on `touchstart` in Vue prevents any actions on the target, while in React it logs an error to the console.

![Screenshot_20240401_123920](https://github.com/radix-vue/radix-vue/assets/102010617/66129f04-9f98-4505-b9dc-fb2be0291686)
![Screenshot_20240401_123942](https://github.com/radix-vue/radix-vue/assets/102010617/a01b266f-a814-49af-b1f9-37da71755651)
![ezgif-6-46588a02e8](https://github.com/radix-vue/radix-vue/assets/102010617/ba8ca5b7-5650-4602-902c-7fd095aeca22)

ChrisGV04:
What @zernonia means is that this component is only meant to be triggered on **hover** as the name implies. This is because there are certain behaviors you only want to work with a mouse on hover, not because of browser limitations.

[...]

PR #803, opened 2024-04-01
fix: make hover card tappable

remonke:
Let's just remove this.

In Radix Primitives for React this doesn't even do anything as the `event` there is some weird React abstraction.

### DIFF
--- a/packages/radix-vue/src/HoverCard/utils.ts
+++ b/packages/radix-vue/src/HoverCard/utils.ts
@@ -1,6 +1,6 @@
-export function excludeTouch(eventHandler: () => void) {
-  return (event: PointerEvent) =>
-    event.pointerType === 'touch' ? undefined : eventHandler()
+export function excludeTouch(event: PointerEvent, eventHandler: () => void) {
+  if (event.pointerType !== 'touch')
+    eventHandler()
 }
 
 /**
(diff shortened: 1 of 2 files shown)


remonke:
#802

ChrisGV04:
Looks great!

I didn't see this PR and opened another one almost identical, but I already closed it. Sorry.

I also suggest replacing the end of docs with the real keyboard interactions: `docs/content/components/hover-card.md`

```md
## Accessibility

The hover card is intended for sighted users only, the content will be inaccessible to keyboard users.

### Keyboard Interactions

<KeyboardTable :data="[
    {
      keys: ['Tab'],
      description: 'Opens/closes the hover card.',
    },
    {
      keys: ['Enter'],
      description: 'Opens the hover card link',
    }]" />
```

This comes from Radix UI.

packages/radix-vue/src/HoverCard/utils.ts
zernonia, line 1: let's revert this changes. Wanna keep the similar callback reference, and don't think this is causing any issue

zernonia (commented):
Thanks for the PR @remonke ! Happy with removing `@touchstart.prevent`, however I'll like to keep the utility function the same.

And let's add the docs regarding the Accessibility as suggested by @ChrisGV04  😁

zernonia (approved):
Thanks for the PR @remonke . Reverted `@pointerleave` as it overwrote the `GraceArea` feature of HoverCard

PR #806, opened 2024-04-01
fix(date): shared calendar year jump on createDecade

epr3:
Hello,

This PR fixes an issue outlined by @sadeghbarati in which the `createDecade` doesn't output the correct year progression, but it just jumps from the current year to 10 years in the future.

Screenshot of the issue on `CalendarPopover.story.vue`:
Screenshot 2024-04-01 at 14 17 52


zernonia (approved):
LGTM! ⭐️

Issue #804, opened 2024-04-01
[Bug]: Select triggers on touch scroll


ChrisGV04:
### Environment

```bash
- Operating System: Darwin
- Node Version:     v18.17.1
- Package Manager:  pnpm@8.15.4
- Radix Vue version: 1.6.1
- Browser: Chrome 123
```

### Link to minimal reproduction

https://www.radix-vue.com/components/select.html

### Steps to reproduce

On touch devices, try to scroll up/down by starting the scroll gesture on top of the Select trigger.

### Describe the bug

If you start a scroll gesture, the `Select` gets triggered and opens the `SelectContent`, which feels wrong. Any other component if you start the gesture it doesn't trigger until an actual click/tap happens.

I believe the issue comes from the `SelectTrigger` which listens to the `@pointerdown` event and on touch devices it immediately triggers the event even on scroll.

I see that even the official Radix UI has the same problem.

### Expected behavior

I would expect that the `Select` only gets triggered on click, because if I'm trying to scroll the page starting from the trigger, it blocks scroll and opens the options, which doesn't feel right.

### Context & Screenshots (if applicable)

https://github.com/radix-vue/radix-vue/assets/47041342/6a6cf580-6192-49d1-a3b5-2516b980fe63

MellKam:
It seems like `radix-ui` also has this issue


https://github.com/radix-vue/radix-vue/assets/51422045/5e1d2ca6-e331-44d4-bab0-2c32916c214b

ChrisGV04:
Indeed. And I'm pretty sure that it's because they both use the `@pointerdown` event to open the content instead of `@click`.

I just don't understand exactly what's the reason behind that choice. Even other components that have similar behaviors don't do it like that.

The `Popover` and `DropdownMenu` don't have the issue because they do trigger `@click`.

ChrisGV04:
I have been playing for a while now trying to get the interaction to work as expected but it seems to be more complicated than I thought... Here are the options I tried:

### Merge `@pointerdown` to `@click`
This fixes the scroll activation, but for some reason this caused selecting an item with the `Enter` key to not close the menu.

### Replace `@pointerdown` with `@pointerup`
This also fixes the scroll activation, but now sometimes pressing on a `SelectItem` with touch devices doesn't select it until a second press. I have no idea why that happens. Reading MDN makes me think that using pointer events requires more control.

ChrisGV04:
I think I have a solution... Will submit PR in a few minutes

PR #807, opened 2024-04-01
fix(Select): Touch scroll triggers the select content

ChrisGV04:
This PR aims to improve the behavior of the `Select` component on touch devices.

Fixes: #804 

### The Problem

On the latest version of Radix UI (v1.6.1) if you try to scroll on a touch screen device and the initial tap to begin the scroll gesture fell on the `SelectTrigger`, it prevents scrolling and immediately opens the `SelectContent`.

This behavior feels wrong to the end user because they expect to be able to scroll by tapping anywhere on the screen. But if they tap on any `SelectTrigger`, not only does it lock scrolling, it also opens the `SelectContent` and feels like a bug.

See #804 for more details.

### The Cause

The issue was caused by the `SelectTrigger` opening the `SelectContent` via the `@pointerdown` event, which feels normal with a mouse, but on touch devices it gets triggered immediately after touching the screen, even if it's for a gesture.

This was done because if we triggered the open event on `@click`, restoring the focus on the `SelectTrigger` after the content closes caused it to get triggered again, so the `SelectContent` did not get closed on `Enter`.

### The Solution

This proposal replaces the `@pointerdown` with `@pointerup` and follows the same behavior as the `DropdownMenuTrigger` to prevent focus competition.

It also removes the `watchEffect` from `SelectContentImpl.vue` that prevented the `@pointerdown` event from selecting an option accidentally. Since we now use `@pointerup`, the pointer is no longer interacting and cannot trigger an item. (This was also the behavior that prevented an option from being selected until a second `pointerup` event was triggered)

I updated the tests to use `@pointerup` as well. I also tested on Safari, Chrome and Firefox on both iOS, MacOS, Windows and Android and the behavior seems to work as expected.

### DIFF
--- a/packages/radix-vue/src/Select/Select.test.ts
+++ b/packages/radix-vue/src/Select/Select.test.ts
@@ -4,6 +4,7 @@ import Select from './story/_SelectTest.vue'
 import type { DOMWrapper, VueWrapper } from '@vue/test-utils'
 import { mount } from '@vue/test-utils'
 import { nextTick } from 'vue'
+import { fireEvent } from '@testing-library/vue'
 import { handleSubmit } from '@/test'
 
 describe('given default Select', () => {
@@ -33,7 +34,7 @@ describe('given default Select', () => {
 
   describe('opening the modal', () => {
     beforeEach(async () => {
-      await wrapper.find('button').trigger('pointerup', {
+      await wrapper.find('button').trigger('pointerdown', {
         button: 0,
         ctrlKey: false,
       })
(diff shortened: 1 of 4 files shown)


ChrisGV04:
Hi @zernonia!

I believe this PR is ready to be reviewed if you're available.

Please let me know if there are any improvements to be made. I'll be glad to help.

zernonia:
Unfortunately this changes causes the change in behaviour. This idea for this `Select` component is to mimic the native `<select>` as close as possible. I have a quick solution.

in `pointerdown` event, we add 
```
      if (event.pointerType === 'touch')
            return 
```

in `pointerup` event, we add
```
     if (event.pointerType === 'touch')
          handleOpen()
```


this way, it fixes the touch scrolling issue, and persist the original behavior.

ChrisGV04:
Oh, I didn't know those behaviors of the native `select`. Always learning something new 😅

I reverted the changes and implemented what you suggested.

I also had to add the touch prevention to the `handlePointerUp` inside the `SelectContentImpl.vue` because it was causing it to sometimes need to tap on an option twice to be selected.

zernonia (approved):
Thanks for the PR @ChrisGV04 ! 💚🚀

PR #813, opened 2024-04-03
Fix size prop should be width (number)

nicooprat:
Looks like the `size` (string) prop should actually be `width` (number) in examples according to the docs.

Thanks for this excellent library!


zernonia (approved):
LGTM! Thanks for catching that @nicooprat  💚

PR #827, opened 2024-04-06
fix(useCalendar): grid update condition in watcher

epr3:
Hello,

This PR fixes an issue with the `Calendar/RangeCalendar` always updating the grid view (with the `numberOfMonths` prop set to a number greater than 1) when selecting a date on a month other than the first in the grid view, even if the date is part of the `visibleView`.

EDIT: This PR also makes all props inside `useCalendar` reactive and adds a watcher to update the `CalendarGrid` on props update.

cc @sadeghbarati for reporting this 🚀


zernonia (approved):
LGTM!  🤩

PR #830, opened 2024-04-06
fix(docs): correct TooltipContent data-state values

remonke:


zernonia:
Thanks for the PR @remonke !

PR #831, opened 2024-04-07
feat: expose internal date utils functions

epr3:
Hello,

This PR exposes some of the functions used internally in the `Calendar`/`Date Field` components for creating months/years/decades, determining the types of `DateValue` objects and comparison functions.

These functions might prove useful for users who might want to build more complicated `Calendar`-type components and having these utilities might help with the manipulation of the `placeholder`/`modelValue` and/or building more complex `Calendar` grids.


zernonia:
Thanks @epr3 ! I like the idea of exposing these utility. However I'm not sure we should expose it as a composable (useXXX). As the term `use` mostly ties with composable that utilize Vue reactivity, like how it does in VueUse.

I think for clarity sake we can rename `useDateUtils` to `dateUtility`.
or maybe we expose this utility as such `radix-vue/date`? WDYT?

epr3:
Hello @zernonia,

Both variants sound very good for me. 

Let me know what updates should I do to this PR.

zernonia:
@epr3  I'm leaning toward `radix-vue/date`. This way all `Dates` related utility, types, function & composable can be moved there. Though it might have some breaking change as the import path will be different now, but since it is still in `alpha` tag, it should be quick and easy update for folks.

epr3:
@zernonia Will add a new folder under `packages` and move everything `Date` related there.

PR #846, opened 2024-04-11
docs: reference hydration section in SSR page

Saeid-Za:
Thanks for supporting custom `useId` implementation ! It seems it went under my radar.
Since this is an SSR issue, I've added some information on SSR page from `ConfigProvider` section.


Saeid-Za:
A quick question, is the `note` on this page still relevant?

zernonia:
Good addition here. I totally missed out the SSR page 😂

Yeah let's remove the `info` part here

Saeid-Za:
Done.

PR #852, opened 2024-04-16
feat: add combobox multiple select with delete item icon

no2ehi:
I added a new example to the combobox component in story, which displays selected items above the input and allows each item to be removed. I thought this example was necessary to better showcase other uses of this component


zernonia:
Thanks @no2ehi . However we have a better example [here](https://www.radix-vue.com/components/tags-input.html#with-combobox). It caters for the acccessibility as well

no2ehi:
Thanks for reminding me of this.

PR #857, opened 2024-04-18
fix(pagination): set pageCount minimal value is 1

preflower:
this PR to fix Pagination component abnormal behaviours when total is 0:
1. `PaginationList` items is [] when total is 0
2. `PaginationNext` is clickable when total is 0

### Demo
image


zernonia:
Nice catch and PR @preflower !!! 
Can you help add a simple test for the case above? 😁

zernonia (approved):
Nicely done @preflower ! Appreciate the fix and the test 💚🔥

PR #860, opened 2024-04-18
docs: improve Tabs docs

ArtemTropanets:
Adds explanation for `activationMode` prop on `TabsRoot`.
Adds outline shadow on `TabsTrigger` focus for Tabs demo. I removed the animation on `TabsIndicator` because it doesn't make sense with the shadow.


zernonia (approved):
Thanks for the PR @ArtemTropanets . I've modified the demo component to show the border only on keyboard focus (focus-visible). I want to keep the `Indicator` component to showcase it.

PR #863, opened 2024-04-20
feat(tagsInput): added support for object in tags input

onmax:
## What is in this PR

Adds support for object in `TagsInput`. See the new [`Object story`](https://github.com/radix-vue/radix-vue/blob/onmax/tags-input-objects/packages/radix-vue/src/TagsInput/story/TagsInputObject.story.vue) to see the code

> This PR is non-breaking

### DIFF
--- a/packages/radix-vue/src/TagsInput/TagsInput.test.ts
+++ b/packages/radix-vue/src/TagsInput/TagsInput.test.ts
@@ -1,4 +1,4 @@
-import { beforeEach, describe, expect, it } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { axe } from 'vitest-axe'
 import TagsInput from './story/_TagsInput.vue'
 import TagsInputObject from './story/_TagsInputObject.vue'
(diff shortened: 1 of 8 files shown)


onmax:
Any feedback is welcomed!

zernonia:
Awesome job @onmax ! Thanks for taking the initiative to add this feature! I will check and test around this soon! 😁

zernonia:
Really well done @onmax . Love that the props is quite straightforward, and the neccessary error log.

Let's add some test for this ok? 😁

onmax:
What do you think about the new type `AcceptableInputValue`, do you see it useful in other components in the future?

That way I can put it in the shared types file along with `AcceptableValue`

onmax:
done! I don't know if we want to add more tests, I feel like they might be overengineered... So I keep it only very basic ones. Let me know if you prefer more cases

packages/radix-vue/src/TagsInput/TagsInput.test.ts
```diff
@@ -165,3 +167,40 @@ describe('given default TagsInput', () => {
...
+    await input.trigger('keydown.enter')
+    await nextTick()
+    tags = wrapper.findAll('[data-radix-vue-collection-item]')
+  }
+
+  it('should update the tags', async () => {
+    await addTag('Moriah Stanton')
+    expect(tags.at(-1)?.text()).toBe('Person: Moriah Stanton')
+    expect(wrapper.vm.people).toEqual(expect.arrayContaining([expect.objectContaining({ id: expect.any(Number), name: expect.any(String) })]))
+  })
```
zernonia, line 205: let's add another case where `convertValue` is not provided, and the modelValue is `object` type. We should expect error log to be logged

zernonia:
> What do you think about the new type `AcceptableInputValue`, do you see it useful in other components in the future?
> 
> That way I can put it in the shared types file along with `AcceptableValue`

I think we don't need to put them at `shared` for now. Not sure if there's any other component would utilize it

zernonia (approved):
Well done @onmax !!! I've fixed the test you were having issue with. Also run some lint and codegen.

Appreciate the PR! It will be coming on `v1.8` 😁

PR #866, opened 2024-04-21
fix(Tabs): do not infer type of `modelValue` from `defaultValue` prop

DamianGlowala:
This PR:
- prevents `defaultValue` from being inferred by the `modelValue` prop
- replaces `StringOrNumber` with locally declared `ModelValue` type (and, most importantly, replaces it in the docs with `string | number` - its simplest and easiest form)
- renames `T` generic param to `TModelValue` for the sake of clarity
- bumps TypeScript version so that `NoInfer` util can be used

(Other errors seem to be revealed in CI which need to be addressed.)


zernonia:
Thanks for the PR @DamianGlowala . Seems like the Typescript bump causing the error in CI. Is it neccessary to have `NoInfer` here?

DamianGlowala:
`NoInfer` prevents a bug when you use a 'modelValue` combined with `defaultValue`. Currently, generic param will merge these two, which means that `modelValue` isn't the only source of truth and can be extended, which shouldn't be the case.

As far as I remember, this utility type has been introduced in TS v5.4, hence a need for a bump.

zernonia (changes requested):
@DamianGlowala  can you check why the pipeline is failing? Perhaps there's some changes in 5.4 that causes some infinite loop

DamianGlowala:
@zernonia I have resolved 3 out of 4 issues (improved/restructured generics a bit!). If you're fine with the remaining issue being silenced with `@ts-expect-error`, shall we proceed with the merge? 😄

packages/radix-vue/src/shared/types.ts
```diff
@@ -45,6 +45,5 @@ type ScrollBodyOption = {
 
 type AcceptableValue = string | number | boolean | Record<string, any>
 type ArrayOrWrapped<T> = T extends any[] ? T : Array<T>
- type StringOrNumber = string | number
```
zernonia, line 48: I think having this type named as `StringOrNumber` is much better than `ModelValue`. `ModelValue` naming can be ambiguous and if more component using this name it's gonna cause some confusion. 

Let's stick it back to `StringOrNumber` please 😁

zernonia:
> @zernonia I have resolved 3 out of 4 issues (improved/restructured generics a bit!). If you're fine with the remaining issue being silenced with `@ts-expect-error`, shall we proceed with the merge? 😄

I see  `@ts-expect-error` only applied to `Collection` right? 😁

packages/radix-vue/src/shared/types.ts
DamianGlowala: Why would it be confusing to name something more closely resembling its purpose, especially knowing the `ModelValue` type is placed inside of the component it is closely related to (unlike generic `StringOrNumber`)? Two reasons I dislike this abstraction:

- reduces cognitive load when mentioned in the docs (as a user, I'd much rather see a prop is of type `string | number` than `StringOrNumber`)
- why should we abstract something so short and obvious? If the type ever changes, so does the name `StringOrNumber`. I can't see the purpose of having it at all.

Nevertheless, let me know if the above makes sense at all - I'm happy to pick whichever option we decide on 😄

DamianGlowala:
> > @zernonia I have resolved 3 out of 4 issues (improved/restructured generics a bit!). If you're fine with the remaining issue being silenced with `@ts-expect-error`, shall we proceed with the merge? 😄
> 
> I see `@ts-expect-error` only applied to `Collection` right? 😁

@zernonia Yes, indeed! 😄It's only in one place. Other three errors were to do with destructuring variables which TypeScript didn't have enough 'confidence' they were objects. I slightly adjusted the generics and now it works great.

PR #874, opened 2024-04-23
feat(Select): body lock property

epr3:
Hello,

This PR adds a `bodyLock` property to the `SelectContent` component behaving in the same way as the `ComboboxContent` component.

This PR should also fix #873.

Let me know if there's anything else that would need to be added.

Thanks!

### DIFF
--- a/docs/content/meta/SelectContent.md
+++ b/docs/content/meta/SelectContent.md
@@ -39,6 +39,12 @@
     'required': false
   },
   {
+    'name': 'bodyLock',
+    'description': '<p>The document.body will be lock, and scrolling will be disabled.</p>\n',
+    'type': 'boolean',
+    'required': false
+  },
+  {
     'name': 'collisionBoundary',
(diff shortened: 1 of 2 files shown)


zernonia:
Thanks for the quick PR @epr3 !

MellKam:
It seems like I'm late, but I think it should be named `modal` instead of `bodyLock` to keep prop naming consistent. Wdyt? @epr3

epr3:
Hello @MellKam,

I agree with keeping the prop names consistent, but from what I've gathered from the codebase:

- `modal` is used on `Root` components, whereas `bodyLock` is used on `Content` components
- `modal` is used to signal that the component disables/enables interaction outside itself, `bodyLock` only handles the scroll
- `Combobox` uses the `bodyLock` prop, which is basically a `Select`

MellKam:
@epr3 Okey, it makes sense 👍🏻

zernonia:
Thanks for asking abou this @MellKam ! Hope that the answer above clarifies the decision on the prop naming here 😁

PR #879, opened 2024-04-26
chore: fix lint issues

onmax:
It would be nice to have custom rules in the `eslint.config.cjs` for the files `**/**.*.story.vue` and turn off rules like `no-console` and `no-alert`, but I am not sure how to implement it.

So, instead I am using `// disable-next-line no-console` in multiple stories

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -16,4 +16,15 @@ module.exports = {
     'max-statements-per-line': ['error', { max: 2 }],
     'vue/one-component-per-file': 'off',
   },
+
+  overrides: [
+    {
+      files: ['*.story.vue'],
+      rules: {
+        'no-console': 'off',
+        'no-alert': 'off',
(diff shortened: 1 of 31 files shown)


zernonia:
we purposely exclude linting `story` files because we wanna keep some code for easier debugging. We don't really need to add those `// eslint-disable-next-line no-alert` too as it's just dev environment.

zernonia (changes requested):
we purposely exclude linting `story` files because we wanna keep some code for easier debugging. We don't really need to add those `// eslint-disable-next-line no-alert` too as it's just dev environment. 

Let's revert that ya 😁

onmax:
Perfect. I've added some eslint rules to ignore stories.

Can I create another PR updating eslint to v9 and flat config? I will try the changes to the minimum

zernonia:
Awesome!!!
Let's create another PR for v9. I heard some bad experience with v9 🙈

PR #881, opened 2024-04-26
feat(Calendar/RangeCalendar): step calendar increment buttons

epr3:
Hello,

This PR implements a new property on the `CalendarNext`/`CalendarPrev` buttons named `step` which would be used to control how the `placeholder` is incremented/decremented (by year or by month, the last one being the default).

I believe this feature is implemented on most calendar libraries/packages and could prove useful for a lot of users.

I've also added a story and example in `Calendar` and tests for this new feature.


zernonia:
Yet another awesome job @epr3 !!!! 🚀

PR #882, opened 2024-04-28
chore: typos

sadeghbarati:
Based on this PR https://github.com/nuxt/nuxt/pull/26907 💯  found this tool which also has GitHub action too

https://crates.io/crates/typos


zernonia:
Nice @sadeghbarati ! I think we can add the GitHub Action too. Can you help with that? 😁

sadeghbarati:
@zernonia I have zero knowledge in **CI** 😄 

But we need to discuss when it should be run

- on every PR? 
- or with GitHub action manually trigger?

zernonia:
Hahaha no prob @sadeghbarati . I think we can have another PR for adding the CI 💪🏻

PR #888, opened 2024-05-01
Docs(accordion): fix aria-hidden attribute in examples

dwightjack:
This PR is not related to radix-vue itself, but to the usage of `@iconify/vue`. 

@iconify/vue already [adds `aria-hidden` by default](https://github.com/iconify/iconify/blob/59bf4012b5b66fae6b63716fac705fd275da06c2/components/vue/src/render.ts#L24), but if we set the `aria-hidden` attribute without any value, [this check](https://github.com/iconify/iconify/blob/59bf4012b5b66fae6b63716fac705fd275da06c2/components/vue/src/render.ts#L162-L167) will remove it completely, leaving a `role="img"` without name.

If this change is fine and I didn't miss anything, I can submit another to update other examples where we used the attribute on the `Icon` component.

### DIFF
--- a/docs/components/demo/Accordion/css/index.vue
+++ b/docs/components/demo/Accordion/css/index.vue
@@ -34,7 +34,7 @@ const accordionItems = [
         <AccordionHeader class="AccordionHeader">
           <AccordionTrigger class="AccordionTrigger">
             <span>{{ item.title }}</span>
-            <Icon icon="radix-icons:chevron-down" class="AccordionChevron" aria-hidden />
+            <Icon icon="radix-icons:chevron-down" class="AccordionChevron" />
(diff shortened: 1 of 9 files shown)


zernonia:
Thanks for picking this up @dwightjack ! Can you update other similar changes in this PR too? 😁

dwightjack:
> Thanks for picking this up @dwightjack ! Can you update other similar changes in this PR too? 😁

@zernonia Sure! I removed every other similar instance in https://github.com/radix-vue/radix-vue/pull/888/commits/617bea561035f538cf481266a009546f893f52b2.

zernonia:
Really appreciate the PR @dwightjack ! Good job! 💚🚀

PR #889, opened 2024-05-01
Fix(accordion): remove unused open prop from the CollapsibleContent instance

dwightjack:
In `AccordionContent` we are passing a `open` prop to the `CollapsibleContent` instance even if it does not support such prop.

On a related note: I noticed that the `data-disabled` attribute is applied even if the item or the root is not disabled. It might have something to do with the line: 

https://github.com/radix-vue/radix-vue/blob/3a4188c1afbdb7671c0e76a7e9e0253b34defa52/packages/radix-vue/src/Accordion/AccordionItem.vue#L68

but I lack the context to understand if it's a bug or not.

### DIFF
--- a/packages/radix-vue/src/Accordion/AccordionContent.vue
+++ b/packages/radix-vue/src/Accordion/AccordionContent.vue
@@ -21,7 +21,6 @@ useForwardExpose()
 <template>
   <CollapsibleContent
     role="region"
-    :open="itemContext.open.value"
     :hidden="!itemContext.open.value"
     :as-child="props.asChild"
     :aria-labelledby="itemContext.triggerId"


zernonia:
Thanks for another PR @dwightjack ! The `data-disabled` state would present when the `AccordionItem` is not closable.

In the case of `type=single`, `collapsible=false` and `open=true`, we will make it not closable. Thus it's not really a bug ya 😁

dwightjack:
> Thanks for another PR @dwightjack ! The `data-disabled` state would present when the `AccordionItem` is not closable.
> 
> In the case of `type=single`, `collapsible=false` and `open=true`, we will make it not closable. Thus it's not really a bug ya 😁

@zernonia thanks for the clarification. It makes totally sense 👍

Issue #893, opened 2024-05-06
[Docs]: Clicking on Date Range Picker gives 404 


jxjj:
### Environment

```bash
Developement/Production OS: MacOS 14.4.1 (23E224)
Node version: v20.12.2
Package manager: pnpm@8.15.4
Radix Vue version: 1.7.3
Client OS: MacOS 14.4.1 (23E224)
Browser: Brave Version 1.65.126, Chromium: 124.0.6367.118 (Official Build) (arm64)
```

### Link to minimal reproduction

https://www.radix-vue.com/

### Steps to reproduce

1. Navigate to Radix Vue homepage
2. Scroll to "Date Range Picker"  demo container
3. Click on title: "Date Range Picker"

### Describe the bug

Clicking on Date Range Picker brings user to 404 page rather than to the Date Range Picker details page.

### Context & Screenshots (if applicable)

![ScreenShot 2024-05-05 at 23 47 18](https://github.com/radix-vue/radix-vue/assets/980170/db3d61f6-51b4-4753-93dc-d4c5d14d6884)

PR #894, opened 2024-05-06
docs: fix 404 when clicking on "Date Range Field"  DemoContainer

jxjj:
On the home page, when a user clicks on a demo container with a title containing more than two words, they will get a 404.

For example, "Date Range Picker" generates the link: `/components/date-range%20picker.html` rather than the correct link `/components/date-range-picker.html`.

The component responsible for generating this url is `<DemoContainer>`. It uses the `title` prop to generate the link. When a title has a space, it currently substitutes a hyphen for the space using `title.replace(" ", "-")`.

However, [`.replace()` only replaces the first instance when passed a string](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace) as the pattern, so only the first `" "` is replaced with a `"-"`.

This updates to use a regex in `title.replace()` so that each instance of whitespace is replaced with a hyphen.

Resolves https://github.com/radix-vue/radix-vue/issues/893

### DIFF
--- a/docs/components/DemoContainer.vue
+++ b/docs/components/DemoContainer.vue
@@ -11,7 +11,7 @@ defineProps<{
   <div class="w-full" :class="{ 'overflow-x-auto': overflow }">
     <a
       class="capitalize md:text-lg font-semibold mb-2 ml-2 inline-flex items-center group"
-      :href="`/components/${title?.replace(' ', '-')}.html`"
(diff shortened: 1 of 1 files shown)


zernonia:
Thanks for the issue and fix @jxjj ! 😁

PR #901, opened 2024-05-07
feat: allow hooking on Dialog animation end

### LINKED ISSUE #878: [Feature]: Emit event when Dialog's animation are completed
### Describe the feature

Hi, contributor of [@nuxt/ui](https://github.com/nuxt/ui) here. As we are porting features from v2 to v3, I noticed something I am unable to port when transitioning from HeadlessUI to Radix.

When using [modals programmatically](https://ui.nuxt.com/components/modal#control-programmatically), we listen to [`TransitionRoot @after-leave`](https://headlessui.com/react/transition#component-api) so we know we can reset the modal safely.

I understand that `Presence` is used under the hood, but I can't determine if I can use anything to hook into that.

### Additional information

noook:
Resolves #878 

Not sure this is the expected method, I'd love to have feedback on it

### DIFF
--- a/packages/radix-vue/src/Dialog/DialogContent.vue
+++ b/packages/radix-vue/src/Dialog/DialogContent.vue
@@ -34,7 +34,6 @@ const { forwardRef } = useForwardExpose()
 <template>
   <Presence
     :present="forceMount || rootContext.open.value"
-    @animation-end="emits('animationEnd')"
   >
     <DialogContentModal
       v-if="rootContext.modal.value"
--- a/packages/radix-vue/src/Dialog/DialogContent.vue
+++ b/packages/radix-vue/src/Dialog/DialogContent.vue
@@ -32,7 +32,10 @@ const { forwardRef } = useForwardExpose()
 </script>
 
 <template>
-  <Presence :present="forceMount || rootContext.open.value">
+  <Presence
+    :present="forceMount || rootContext.open.value"
+    @animation-end="emits('animationEnd')"
+  >
     <DialogContentModal
       v-if="rootContext.modal.value"
       :ref="forwardRef"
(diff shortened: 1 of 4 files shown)


zernonia:
Thanks for the PR @noook . I'm thinking a more general solution for components that are using `Presence`. I did some test with dispatching custom event, and seems to work pretty well.

something like this 

```ts
  const dispatchCustomEvent = (name: 'enter' | 'after-enter' | 'leave' | 'after-leave') => {
    const customEvent = new CustomEvent(name, { bubbles: true, cancelable: true })
    node.value?.dispatchEvent(customEvent)
  }
```

Then in the `DialogIOverlay` (or any components with Presence) can add listener as such

```vue
<DialogOverlay class="animation" @leave="" @after-leave=""   />
```

noook:
I get it, what do you think of it now @zernonia ?

I'm not sure all events can be fired, for example I haven't seen the `leave` event fired. Also, I am not sure I can detect `after-enter` either

zernonia:
No prob @noook ! I've pushed some `dispatchCustomEvent` to cater for `after-X` event. 

I believe this feature fit your original request right? where you can hook into the animation state 😁

noook:
That looks fine ! And yes, I tested in the stories it fits the request 🔥

zernonia:
Right now it's not explicit that which component has the `hook`, but I'm planning to add them in docs soon 😁

stevebauman:
I'm not seeing anywhere in the docs on how to do this? Was this removed in a later patch? I.e. this does not work (not events fired):

```vue
<DialogOverlay 
    @leave="onLeave"
    @after-leave="onLeave"
    v-on:leave="onLeave"
    v-on:after-leave="onLeave"
>...
```

Am I missing something?

PR #912, opened 2024-05-13
fix: make `null` to showPlaceholder

zernonia:


kilobyte2007:
Shouldn't we then also add null as the type to the`modelValue` prop type here - https://github.com/radix-vue/radix-vue/blob/64d7eb233c9c6009a5069ccb9d21849eff47b0a4/packages/radix-vue/src/Select/SelectRoot.vue#L14 ?

PR #914, opened 2024-05-13
Docs(AlertDialog): Fix default values for the components

dwightjack:
Correct the documentation about the default values for some of the components in `AlertDialog`.

### DIFF
--- a/docs/content/meta/AlertDialogAction.md
+++ b/docs/content/meta/AlertDialogAction.md
@@ -6,7 +6,7 @@
     'description': '<p>The element or component this component should render as. Can be overwrite by <code>asChild</code></p>\n',
     'type': 'AsTag | Component',
     'required': false,
-    'default': '\'div\''
+    'default': '\'button\''
   },
   {
     'name': 'asChild',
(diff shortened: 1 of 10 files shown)


zernonia:
the `meta/*.md` files were auto-generated. I think `vue-component-meta` couldn't pick up the default props from `Dialog` in this case. We should assign a default `as` in the `AlertDialog` component instead.

dwightjack:
Oh, sorry, I didn't noticed it. I'll try to investigate if we can fix this issue via vue-component-meta.

dwightjack:
@zernonia I investigated this problem, and indeed, `vue-component-meta` is not picking the default value, probably because it's not defined in the derived components (for example `AlertDialogAction`), but in the parent component (`DialogClose`). 

I am not familiar with `vue-component-meta` but I don't think it could infer the default value in any way. One option could be to re-define the default values in the derived components, even if that's a repetition 🤔 .

zernonia:
Yup. The `vue-component-meta` is using static analysis, there's some limitation to it. I'm ok with re-defining the default value for them 😁

dwightjack:
Cool. I will update the PR 👍

dwightjack:
@zernonia sorry for the delay. I updated the PR.

zernonia (approved):
LGTM! Thanks for the PR @dwightjack ! 🚀

PR #916, opened 2024-05-13
feat: export CalendarRoot props

RayGuo-ergou:
### Description

Export typo `MultipleCalendarRootProps` and  `SingleCalendarRootProps` used by `CalendarRootProps`.

### Motivation 

I was working on build an UI library based on radix-vue/shadcn-vue, this error shows up when build:

image

This seems a typescript limitation that interface of type alias should also be exposed.

I have also tried:

```ts
interface CalendarRootProps extends MultipleCalendarRootProps, SingleCalendarRootProps {
}
```

which will also work because it's an interface, but I think they are not functionally equivalent.

### DIFF
--- a/packages/radix-vue/src/Calendar/CalendarRoot.vue
+++ b/packages/radix-vue/src/Calendar/CalendarRoot.vue
@@ -83,14 +83,14 @@ interface BaseCalendarRootProps extends PrimitiveProps {
   dir?: Direction
 }
 
-interface MultipleCalendarRootProps extends BaseCalendarRootProps {
+export interface MultipleCalendarRootProps extends BaseCalendarRootProps {
   /** The controlled checked state of the calendar. Can be bound as `v-model`. */
   modelValue?: DateValue[] | undefined
   /** Whether or not multiple dates can be selected */
   multiple: true
 }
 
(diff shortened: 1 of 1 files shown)


zernonia:
Thanks for the PR @RayGuo-ergou ! Can you share more details of the error you are having? It's weird that such error being thrown, as internally the component should be able to refer those types correctly.

RayGuo-ergou:
It's only happening when the build reference and expose this type for my library.
e.g. I have the props type

```ts
// types.ts
export type MyProps = CalendarRootProps & {
  ...
}
```

```ts
// index.ts

export * type from './types'
```

or even just use in `Calendar.vue`, because the build will create a type declare file for each component, will also see this error.

I will create a reproduction repo later.

RayGuo-ergou:
Hi @zernonia, here's a reproduction of this issue: https://github.com/RayGuo-ergou/radix-reproduction

It's actually caused by th `vite-plugin-dts` because that plugin is to export type declare files.

RayGuo-ergou:
image

zernonia (approved):
Thanks for the reproduction and the images before/after the fix! 
LGTM!

PR #918, opened 2024-05-13
Feature/typo ci

RayGuo-ergou:
### Description

Typo check github action.

Followed from: https://github.com/radix-vue/radix-vue/pull/882

### DIFF
--- a/docs/content/index.md
+++ b/docs/content/index.md
@@ -1,7 +1,7 @@
 ---
 layout: home
 
-titleTemplate: Unstyledd, accessible components for building high‑quality design systems and web apps in Vue
+titleTemplate: Unstyled, accessible components for building high‑quality design systems and web apps in Vue
 
 hero:
   name: Radix Vue
@@ -31,7 +31,7 @@ features:
     details: Building on top Radix components will save you time and money, so you can ship a better product faster.
   - icon: ⌨️
     title: Accessibility out of the box.
-    details: WAI-ARIA compliant. Support Keyboard navigation & Focus management.
+    details: WAI-ARIA compliant. Suport Keyboard navigation & Focus management.
   - icon: 🧑🏻‍💻
     title: Developer Experience First.
     details: Unstyled, easily customizable and great for building design system and web apps.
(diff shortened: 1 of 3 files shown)


zernonia (approved):

zernonia:
Thanks for picking this up @RayGuo-ergou ! I wonder how can you test the above yaa? I see that you have the workflow created but seems to be not running on `pull_request: open` tho?

RayGuo-ergou:
Sorry missed `synchronize`. 

`opend` only triggered when open a pull request, so because I pushed another commit after open the issue so it's not running the workflow.

btw `types: [opened, reopened, synchronize]` is default types for `pull_request` just to bypass eslint check.

docs/content/index.md
```diff
@@ -1,7 +1,7 @@
 ---
 layout: home
 
-titleTemplate: Unstyled, accessible components for building high‑quality design systems and web apps in Vue
+titleTemplate: Unstyledd, accessible components for building high‑quality design systems and web apps in Vue
```
zernonia, line 4: @RayGuo-ergou  don't think it's working tho. I purposely have some typo here but the CI doesn't catch it or mention it.

RayGuo-ergou: This due to this word has not yet in the typos dictionary. See: https://github.com/crate-ci/typos/blob/27ba9761d15d0f85acd8d0b500ac227e414d084d/CONTRIBUTING.md?plain=1#L32

e.g. it can detect 'donw'
image


To have more precise typo check, [cspell ](https://github.com/streetsidesoftware/cspell) is a better choice but it's much slow and would be a bit painful to setup for large project. e.g. People's name repo's name etc.
image

Would you prefer cspell over typos?

RayGuo-ergou: Actually I would prefer typos as it's fast, less false position and good enough for general typo check.

zernonia: Cool. Seems like it works. Yeah let's go with typo for now. 😁

zernonia (approved):
Awesome work! Thanks for the PR @RayGuo-ergou !

PR #929, opened 2024-05-17
Refactor(Calendar): update props management in navigation buttons and grid component 

dwightjack:
This PR refactors/fixes these two small issues:

- f14f28e9bb0d697aa91c5dd9d28a0bd300a56297: in `CalendarNext` and `CalendarPrev` we are passing the `step` prop to `Primitive` even if it doesn't support it.
- 062c6978262a42a6b7e1428f4bbbfa05269d5a37: in `CalendarGrid`, we assign `aria-disabled` and `aria-read-only` by reading from the root context, but those values are refs into an object, and such are not unwrapped by the compiler.

### DIFF
--- a/packages/radix-vue/src/Calendar/CalendarNext.vue
+++ b/packages/radix-vue/src/Calendar/CalendarNext.vue
@@ -20,6 +20,7 @@ const rootContext = injectCalendarRootContext()
 <template>
   <Primitive
     :as="props.as"
+    :as-child="props.asChild"
     aria-label="Next page"
     :type="as === 'button' ? 'button' : undefined"
     :aria-disabled="rootContext.isNextButtonDisabled(props.step) || undefined"
--- a/packages/radix-vue/src/Calendar/CalendarNext.vue
+++ b/packages/radix-vue/src/Calendar/CalendarNext.vue
@@ -19,7 +19,7 @@ const rootContext = injectCalendarRootContext()
 
 <template>
   <Primitive
-    v-bind="props"
+    :as="props.as"
     aria-label="Next page"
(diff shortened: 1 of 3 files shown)


packages/radix-vue/src/Calendar/CalendarNext.vue
```diff
@@ -19,7 +19,7 @@ const rootContext = injectCalendarRootContext()
 
 <template>
   <Primitive
-    v-bind="props"
+    :as="props.as"
```
zernonia, line 22: `PrimitiveProps` has `as`, and `asChild`. Let's add that too

packages/radix-vue/src/Calendar/CalendarPrev.vue
```diff
@@ -20,7 +20,7 @@ const rootContext = injectCalendarRootContext()
 <template>
   <Primitive
     aria-label="Previous page"
-    v-bind="props"
+    :as="props.as"
```
zernonia, line 23: `PrimitiveProps` has `as`, and `asChild`. Let's add that too

zernonia (changes requested):

packages/radix-vue/src/Calendar/CalendarNext.vue
dwightjack: Thanks for the comment. I fixed the PR in https://github.com/radix-vue/radix-vue/pull/929/commits/cc7e33a719c8702853ff3c151ec0e3fb700987b9. 🙇

zernonia (approved):
LGTM!  🚀

PR #935, opened 2024-05-20
docs(date-picker): update

Simon-He95:


zernonia (approved):
Thanks for the fix @Simon-He95  😁

PR #937, opened 2024-05-20
fix(NumberField): added missed export of NumberFieldRootEmits

romanslonov:
Hello 👋

Just a small fix that added a missed export of NumberFieldRootEmits. This is needed to create a components like shadcn-vue library do.


zernonia (approved):
Nice catch @romanslonov ! Thanks for the PR 🚀💚

PR #943, opened 2024-05-21
 fix: add MaybeRefOrGetter type to useForwardPropsEmits param

RayGuo-ergou:
### Description

Having type error after upgrade to `v1.8.1`. 

### Reproduction

`T` here will parse to `ComputedRef<ComputedRef<T>>` if pass in `ComputedRef<T>`.

e.g.
```ts
const a = computed(() => {
  return {
    a: 1,
  }
})

const test = useForwardPropsEmits(a)
```
will result in 
image

This due to `UnwrapRef` generic `U` was removed from `useForwardProps` in `v1.8.1`


zernonia (approved):
Nice catch! Thanks for the fix @RayGuo-ergou  🚀

Issue #953, opened 2024-05-25
[Bug]: Number Field value remains after deleting field value 


sadeghbarati:
### Environment

```bash
radix-vue v1.8.1
```

### Link to minimal reproduction

https://www.radix-vue.com/components/number-field.html

### Steps to reproduce

empty/delete Number Field value then click on **plus** or **minus** buttons

value starts from the current value ❌ 

### Describe the bug

Number Field value remains even after empty/delete the field value

### Expected behavior

value should start from zero or from starter value ✅ 

https://react-spectrum.adobe.com/react-aria/useNumberField.html#example

sadeghbarati:
Same bug is happening in Combobox

When I delete the Combobox value I expect to deselect the option in the list

https://www.radix-vue.com/components/combobox.html

- select apple
- delete/empty combobox value
- close combobox
- value is still apple after closing combobox

zernonia:
@sadeghbarati  @epr3  that combobox behavior would be some breaking change, and might not be what some dev want. We can include a behavior props in v2.

PR #954, opened 2024-05-25
fix(NumberField): always updateModelValue on blur

epr3:
Hello,

This PR fixes the issue described in #953 on the NumberField. 

It should now start from the `min` value or `0` on `increment`/`decrement` if the input value is cleared.


zernonia (approved):
LGTM! Thanks for the PR @epr3 !

PR #955, opened 2024-05-25
fix(Combobox): clear modelValue on empty search term

### LINKED ISSUE #953: [Bug]: Number Field value remains after deleting field value

epr3:
Hello,

Along with #954, fix #953.

This PR adds a watcher for the `searchTerm`, clearing the `modelValue` if the `searchTerm` is ever empty.

### DIFF
--- a/packages/radix-vue/src/Combobox/Combobox.test.ts
+++ b/packages/radix-vue/src/Combobox/Combobox.test.ts
@@ -37,6 +37,13 @@ describe('given default Combobox', () => {
     expect(wrapper.html()).toContain('Placeholder...')
   })
 
+  it('should clear the selection when the searchTerm is empty', async () => {
+    const input = wrapper.find('input')
+    input.element.value = ''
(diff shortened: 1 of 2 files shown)


zernonia:
Will close this for now, as per comment here https://github.com/radix-vue/radix-vue/issues/953#issuecomment-2131308877

chasegiunta:
Semi-related, I would think that setting the `searchTerm`, either through v-model or prop, would initiate a re-compute of `filteredOptions` but it doesn't that it does...

Issue #960, opened 2024-05-26
[Bug]: Extending ToastRootEmits with enabled declaration throws "A type annotation is necessary (2742)"


hrubysi:
### Environment

```bash
Node version: 18.20.3
Package manager: npm@10.2.3

Radix Vue version: 1.8.2
Vue version: 3.4.27
```

### Link to minimal reproduction

https://stackblitz.com/edit/vitejs-vite-ur52p7?file=src%2FToast.vue

### Steps to reproduce

`npm run build`

### Describe the bug

Extending ToastRootEmits throws type error.

`TS2742: The inferred type of 'default' cannot be named without a reference to '../node_modules/radix-vue/dist/Toast/utils.js'. This is likely not portable. A type annotation is necessary.`

### Expected behavior

No error

hrubysi:
Resolved via adding `"baseUrl": "."` to tsconfig.

PR #961, opened 2024-05-26
fix: export toast utils types

hrubysi:
Closes https://github.com/radix-vue/radix-vue/issues/960


zernonia:
I'm good with the solution, however I think the problem would exist for other components too. Can you help check if `library` mode with Dialog component also face the similar issue? 😁

hrubysi:
That's a good point, DialogContentProps have the same issue. 
The only solution I know of is adding export statement of utils to each component index, but I don't even know why that helps. I would expect it to work without it.

zernonia:
@hrubysi  I realized it might be some `tsconfig` issue on your end. I've forked your stackblitz and modified the `tsconfig` and it bundles correctly (without type issue)

https://stackblitz.com/edit/vitejs-vite-hc6og5?file=tsconfig.lib.json

hrubysi:
You are right. I didn't realize, that base vue ts config doesn't include baseUrl, which is why it didn't work.
Thank you, I'm closing this.

Issue #965, opened 2024-05-27
[Feature]: Remove NumberFieldLabel


romanslonov:
### Describe the feature

As per [shadcn-vue-571](https://github.com/radix-vue/shadcn-vue/pull/571) PR @zernonia decided to remove `NumberFieldLabel`. 

Should I remove this part completely?

```ts
const labelId = useId(undefined, 'number-field-label')
const inputId = useId(undefined, 'number-field-input')
```

And user will target label to input similar as regular `Label` and `Input`?

Example:

```vue
<template>
  <Label for="example-number-field">Number field</Label>
  <NumberFieldRoot>
    <NumberFieldDecrement />
    <NumberFieldInput id="example-number-field" />
    <NumberFieldIncrement />
  </NumberFieldRoot>
</template>
```

### Additional information

- [X] I intend to submit a PR for this feature.
- [x] I have already implemented and/or tested this feature.

zernonia:
Thanks @romanslonov ! We need to make sure we can focus the input given an id to `NumberFieldRoot`

PR #969, opened 2024-05-27
refactor(NumberField): remove NumberFieldLabel

romanslonov:
More info here #965 

I removed `NumberFieldLabel`, passed id from `NumberFieldRoot` component down to it context so input component get it and applied.

However, not sure about this, do we still need it?

```vue
:aria-labelledby="rootContext.id"
```


zernonia:
yeah we can remove the `aria-labelledby`, and update the `story/_NumberField` with additional `label` to prevent aXe test error.

romanslonov:
@zernonia Should it be `<Label />` or just native html element?

zernonia:
> @zernonia Should it be `<Label />` or just native html element?

Either 😁  since in the `test` component we are just making sure if the component has `label` and appropriate `for`/`id` it should pass the aXe test

chasegiunta:
Is 2.0 going to follow semver? I'm worried that putting these breaking changes in patch releases is gonna be a huge issue.

zernonia:
Hey @chasegiunta .. we are following semver in general, however for new `Alpha` component, we would try to introduce minimal breaking changes in patches (if it's really necessary), such as this case.

chasegiunta:
@zernonia Ah that makes sense. Apologies, I did not realize this was an Alpha component. I just read the breaking change in the release notes. Thank you for the clarification!

Issue #966, opened 2024-05-27
[Feature]: Calendar step per week


RamaHerbin:
### Describe the feature

Hello,

Thank you for all your hard work on this amazing library.

I wanted to know if it is possible to navigate the calendar week by week? 
I couldn't find this option in this API doc : [doc link](https://www.radix-vue.com/components/calendar#next-button)

The `step` prop currently only supports two types: '`month`' and '`year`'.

![Calendar](https://github.com/radix-vue/radix-vue/assets/41597427/12f18548-e5f3-4550-9ed4-868fe5537c11)
_Figma ref_

Thank you !

### Additional information

- [ ] I intend to submit a PR for this feature.
- [ ] I have already implemented and/or tested this feature.

PR #972, opened 2024-05-27
feat(Calendar/RangeCalendar): pagination function props

epr3:
Hello,

This PR adds new `nextPage`/`prevPage` function properties to both `CalendarRoot` and `CalendarNext`/`CalendarPrev` components (their Range and Picker counterparts included) which enables more control in terms of handling pagination inside the calendar, while also enabling a more flexible usage of the next/prev components which have all the data attributes and keyboard interactions as opposed to the regular HTML `button`.

The `CalendarNext`/`CalendarPrev` function properties overwrite the function properties defined on `CalendarRoot` (this addresses the use case where you would want pagination by year along with pagination by month).

It should help address the use case described in #966 .

It also deprecates the `step` property on `CalendarNext`/`CalendarPrev`, which could be removed in `v2`.

I am looking forward to any feedback and/or improvement suggestions.

Thanks! 🚀


packages/radix-vue/src/Calendar/CalendarPrev.vue
```diff
@@ -1,10 +1,15 @@
...
+import type { DateValue } from '@internationalized/date'
 
 export interface CalendarPrevProps extends PrimitiveProps {
-/** The calendar unit to go back */
+/** The calendar unit to go back 
+* @deprecated Use `prevPage` instead
+*/
   step?: CalendarIncrement
+/** The function to be used for the prev page. Overwrites the `prevPage` function set on the `CalendarRoot`. */
+  prevPage?: (placeholder: DateValue) => DateValue
```
zernonia, line 12: why do we need `prevPage` if we already provide the props in `CalendarRoot`?

packages/radix-vue/src/Calendar/CalendarNext.vue
```diff
@@ -1,10 +1,15 @@
...
+import type { DateValue } from '@internationalized/date'
 
 export interface CalendarNextProps extends PrimitiveProps {
-/** The calendar unit to go forward */
+/** The calendar unit to go forward 
+* @deprecated Use `nextPage` instead
+*/
   step?: CalendarIncrement
+/** The function to be used for the next page. Overwrites the `nextPage` function set on the `CalendarRoot`. */
+  nextPage?: (placeholder: DateValue) => DateValue
```
zernonia, line 12: why do we need `nextPage` if we already provide the props in `CalendarRoot`?

packages/radix-vue/src/Calendar/CalendarRoot.vue
```diff
@@ -81,6 +81,10 @@ interface BaseCalendarRootProps extends PrimitiveProps {
   isDateUnavailable?: Matcher
   /** The reading direction of the calendar when applicable. <br> If omitted, inherits globally from `ConfigProvider` or assumes LTR (left-to-right) reading mode. */
   dir?: Direction
+  /** A function that returns the next page of the calendar. It receives the current placeholder as an argument inside the component. */
+  nextPage?: (placeholder: DateValue) => DateValue
+  /** A function that returns the previous page of the calendar. It receives the current placeholder as an argument inside the component. */
+  prevPage?: (placeholder: DateValue) => DateValue
```
zernonia, line 87: we we have the default function to behave like `step:  'month'`? So dev dones't need to provide the props everytime

packages/radix-vue/src/Calendar/CalendarRoot.vue
```diff
@@ -169,6 +173,8 @@ const {
   isDateUnavailable: propsIsDateUnavailable,
   calendarLabel,
   defaultValue,
+  nextPage: propsNextPage,
+  prevPage: propsPrevPage,
```
zernonia, line 177: why declare the name with `propsXXX` if there's no naming conflict? 🤔

packages/radix-vue/src/Calendar/CalendarPrev.vue
epr3: For the case when you want multiple `CalendarNext`/`CalendarPrev` components inside the `CalendarRoot` (e.g. the default `CalendarNext`/`CalendarPrev` increments by `month`, the others by `year`) like in the `CalendarWithYearIncrements` story.

packages/radix-vue/src/Calendar/CalendarNext.vue
epr3: Same explanation as for `CalendarPrev`.

packages/radix-vue/src/Calendar/CalendarRoot.vue
epr3: Yes, the default is like `step: 'month'`.

packages/radix-vue/src/Calendar/CalendarRoot.vue
epr3: ESLint said I had a naming conflict with this line: https://github.com/radix-vue/radix-vue/blob/6ca465cc92140c142404f42559d41c1a0c6284b2/packages/radix-vue/src/Calendar/CalendarRoot.vue#L213 (cannot redeclare block-scoped variable 'nextPage')

zernonia (approved):
LGTM! Nicely done @epr3 !
Will be release in `1.9`! 😁

PR #975, opened 2024-05-27
fix: Missing Listbox and NumberField namespaced exports

### LINKED ISSUE #974: [Bug]: Listbox not present in namespaced imports
### Environment

```bash
Package manager: pnpm@9.1.2
Radix Vue version: 1.8.2
```

Any project with Radix Vue version 1.8.2 or lower

### Steps to reproduce

Try to import the `Listbox` component from `radix-vue/namespaced`, but it will throw an error saying that it is not an exported member.

### Describe the bug

The `Listbox` is not included in the namespaced imports

### Expected behavior

This should work, but it doesn't

```vue
<script setup lang="ts">
import { Listbox } from 'radix-vue/namespaced';
</script>

<template>
  <Listbox.Root>
    <!-- ... -->
  </Listbox.Root>
</template>
```

_No response_

ChrisGV04:
This PR only adds the values to export the `Listbox` and `NumberField` from the namespaced exports, since they are not currently present.

Fixes #974


zernonia (approved):
Thanks for the PR @ChrisGV04 !

PR #976, opened 2024-05-28
feat: Tree

zernonia:


sadeghbarati:
Can we use Tree with Checkbox? if yes, adding an example to the doc would be nice

zernonia:
> Can we use Tree with Checkbox? if yes, adding an example to the doc would be nice

Technically yes. Yeah still need to add demo for `Checkbox` and `Draggable/Sortable` tree

packages/radix-vue/src/Tree/TreeItem.vue
```diff
@@ -0,0 +1,173 @@
...
+import { useCollection } from '@/Collection'
+import { handleAndDispatchCustomEvent, useForwardRef } from '@/shared'
+import { flatten } from './utils'
+
+defineOptions({
+  inheritAttrs: false,
+})
+
+const props = withDefaults(defineProps<TreeItemProps<T>>(), {
+  as: 'li',
```
d--j, line 36: Maybe not default to `li`? This makes the default markup [invalid](https://github.com/radix-vue/radix-vue/blob/48f38b09828a7a4c3654767b2e6f285a468a8508/packages/radix-vue/src/Tree/__snapshots__/Tree.test.ts.snap#L4) (`TreeRoot` is a `div`, `TreeItem` is a `li`). Setting `TreeRoot` to `ul`/`ol` would not work with `TreeVirtualizer` (it renders a spurios `div` resulting in `ul > div > li`).

Since this implementation flattens the tree in HTML, using a `ul`/`ol` is not that helpful anyway. One would expect a nested list of lists structure to semantically represent the tree.

romansp:
Is there a plan to support async loading of tree items on expand? I guess it should already be possible to achieve by listening to `update:expanded` (or binding to `expanded` v-model) and then derive `items` from expanded state. I think it should "just work" but was curious if you already had thoughts in this area.

I'm certainly going to build something with async loading after Tree component goes alpha.

packages/radix-vue/src/Tree/TreeItem.vue
zernonia: Thanks for the comment @d--j ! 💚

I've changed `TreeRoot` to be `ul`, and if user is using virtualization they can change the rendered element to `div`/`ul`/`ol` based on their needs.

And the examples we have are using flattened tree (as this will supports virtualization and dnd easily), I'll add example to showcase proper nested list  😁

zernonia:
> Is there a plan to support async loading of tree items on expand? I guess it should already be possible to achieve by listening to `update:expanded` (or binding to `expanded` v-model) and then derive `items` from expanded state. I think it should "just work" but was curious if you already had thoughts in this area.
> 
> I'm certainly going to build something with async loading after Tree component goes alpha.

@romansp  I've added Async example [here](https://github.com/radix-vue/radix-vue/pull/976/commits/fdab11651dc4903fc52e6d4c182d43581f7b1867#diff-5ebdfc2552d18246f2cc3b56ab4dbb771e445d275fa08e523ae45204809ea6d7)! See if that's what you need 😉

romansp:
> @romansp I've added Async example [here](https://github.com/radix-vue/radix-vue/pull/976/commits/fdab11651dc4903fc52e6d4c182d43581f7b1867#diff-5ebdfc2552d18246f2cc3b56ab4dbb771e445d275fa08e523ae45204809ea6d7)! See if that's what you need 😉

Yep, I think this should work thanks! 

I actually tried it out locally with tanstack query for data fetching and it turned out quite neat: https://gist.github.com/romansp/df740e864a7ab3c3b3fc5415b823b824. Sorry for the gist sample. It was quite hard to setup sandbox without Tree published to npm.

Looking forward to use it when it goes public 🎉 .

packages/radix-vue/src/Tree/TreeItem.vue
zernonia: Added a story here https://github.com/radix-vue/radix-vue/pull/976/files#diff-375bb0c69ffc5c2745835f594c74785416bc92517dc35f2e9061688e5490c659

Will include a stackblitz in the docs just in case anyone is looking for it 😁

PR #987, opened 2024-06-04
feat: Stepper

### LINKED ISSUE #805: [Feature]: Steps / Stepper
### Describe the feature

Hello Folks ! Just wondering if there is some thoughts of adding a Stepper component in the roadmap please ? 

### Additional information

epr3:
Closes #805.

Stepper implementation composed of: `Root`, `List`, `Item`, `Indicator`, `Title`, `Description`.

`Item` has local context for supplying data to underlying components.

Main features:

- Navigate and select steps using keyboard
- `linear` property ensures linear navigation of stepper
- steps must have a `Title` for accessibility (like the `Dialog`)
- handles both vertical and horizontal steppers using the `orientation` property
- support for both LTR/RTL contexts

The component is fully documented and tested.

More context here: https://github.com/radix-vue/radix-vue/pull/987#issuecomment-2148987039

Any feedback is appreciated. 🚀


zernonia:
@epr3  Thanks for working on this. However I'm not sure if this component is really needed. From what I can see it's closely resembles the `Tab` component, and can be easily build with `Tab`.

epr3:
@zernonia Yes, I agree. I've seen most implementations follow a `Tab`-like structure.

I was thinking of making this component different than the `Tabs` component so: 
- adding `state` data attributes to permit a more granular styling (e.g. for `inactive`, `active` and `completed` states)
- a `linear` property to limit navigation and focus and a `Title` and `Description` component following the implementation of the `Dialog` component for accessibility (aka describing the current step)
- this component would not have had a `Content` component

I referenced [this](https://designsystem.digital.gov/components/step-indicator/#accessibility-guidance) and [this](https://v2.chakra-ui.com/docs/components/stepper) for the features.

I would have used a `nav` element and marked it with `aria-label="progress"` which in turn would have wrapped an `ul` if `linear: false` or an `ol` tag if `true`. 

Let me know your thoughts! You can close the PR if the component is overlapping the `Tabs` component in usage. 
Thank you! 👍🏻

Celibioux:
@epr3 Thanks for your amazing work ☺️
@zernonia I Think that the approche of @epr3 is great for steps than hack Tabs to look like Steps ? Even Nuxt UI 3 wait for steppers to Land in Radix-vue
Thank you all for your great work

zernonia:
Thanks for the explanation @epr3 ! If that's the case let's continue on with this component! 🚀 
Thanks @Celibioux  for the input, yeah I can see some demands for this component 🤩

Celibioux:
@epr3 Great Work 😎

packages/radix-vue/src/Stepper/StepperList.vue
zernonia, line 1: I think this component is not needed tho. We can have `StepperRoot`  inherit the `data-*` attributes here.

packages/radix-vue/src/Stepper/StepperItem.vue
```diff
@@ -0,0 +1,159 @@
...
+  <Primitive
+    :ref="forwardRef"
+    :as="as"
+    :as-child="asChild"
+    :aria-current="itemState === 'active' ? 'true' : undefined"
+    :data-state="itemState"
+    :disabled="disabled || !isFocusable ? '' : undefined"
+    :data-disabled="disabled || !isFocusable ? '' : undefined"
+    :data-orientation="rootContext.orientation.value"
+    :tabindex="isFocusable ? 0 : -1"
```
zernonia, line 153: Having `tabindex` on `StepperItem` looks kinda weird tho.. perhaps we can have a `button` element named `StepperTrigger` to allow user to focus, and trigger the `handleKeyDown`?

image

packages/radix-vue/src/Stepper/StepperSeparator.vue
```diff
@@ -0,0 +1,25 @@
...
+
+useForwardExpose()
+</script>
+
+<template>
+  <Separator
+    v-bind="props"
+    decorative
+    :orientation="rootContext.orientation.value"
+  >
```
zernonia, line 22: Since the `Separator` is a child of `Item`, maybe we can inject the `itemState` from `Item`? This way we can have the `data-state` auto populated as well. Thought? 😁

zernonia (commented):
Thanks for create working on this component @epr3 !! Just suggestion regarding some implementation.

Also, as there's no WAI-ARIA spec for Stepper, I think we need to think about how to inform user properly (Screen reader/Voiceover mode). Currently the a11y is not well covered yet, especially informing how many steps there were.

packages/radix-vue/src/Stepper/StepperItem.vue
epr3: Sure thing! To clarify, the anatomy would be:
```
<StepperItem>
  <StepperTrigger>
      <StepperTitle />
      <StepperDescription />
   </StepperTrigger>
 <StepperSeparator />
 </StepperItem>
 ``` ?

epr3:
> Thanks for create working on this component @epr3 !! Just suggestion regarding some implementation.
> 
> Also, as there's no WAI-ARIA spec for Stepper, I think we need to think about how to inform user properly (Screen reader/Voiceover mode). Currently the a11y is not well covered yet, especially informing how many steps there were.

@zernonia Thank you very much for the feedback! I can think of adding a visibility hidden div with `aria-live="polite"` to the root with the text of `Step 1 of 5` or something like that for announcing the total number of steps and the current step. Could you let me know if there any other elements that should be added?

packages/radix-vue/src/Stepper/StepperList.vue
epr3: What should the root element be as the default: `nav` or `ol`/`ul` ?

packages/radix-vue/src/Stepper/StepperItem.vue
zernonia: Yup I think this looks good 👍🏻

zernonia:
> > Thanks for create working on this component @epr3 !! Just suggestion regarding some implementation.
> > Also, as there's no WAI-ARIA spec for Stepper, I think we need to think about how to inform user properly (Screen reader/Voiceover mode). Currently the a11y is not well covered yet, especially informing how many steps there were.
> 
> @zernonia Thank you very much for the feedback! I can think of adding a visibility hidden div with `aria-live="polite"` to the root with the text of `Step 1 of 5` or something like that for announcing the total number of steps and the current step. Could you let me know if there are any other elements that should be added?

This is not a bad suggestion! I was looking into the having role `tablist`, `tab`, but afraid that without proper `tabpanel` it would causes bad a11y. Let's go with `aria-live` announcer 😁

packages/radix-vue/src/Stepper/StepperList.vue
epr3: Removed the `ol`/`ul` default tag as it caused accessibility issues.

PR #990, opened 2024-06-06
fix(Tooltip): added Presence to TooltipContent to allow close animation

Azurency:
Hello,

This PR adds `Presence` to `TooltipContent`, to allow using a close animation. I also added a story to see it in action (feel free to remove or ask me to remove this story if that's not needed).

### DIFF
--- a/packages/radix-vue/src/Tooltip/TooltipContent.vue
+++ b/packages/radix-vue/src/Tooltip/TooltipContent.vue
@@ -3,7 +3,13 @@ import type { TooltipContentImplEmits, TooltipContentImplProps } from './Tooltip
 
 export type TooltipContentEmits = TooltipContentImplEmits
 
-export interface TooltipContentProps extends TooltipContentImplProps {}
+export interface TooltipContentProps extends TooltipContentImplProps {
+  /**
+   * Used to force mounting when more control is needed. Useful when
+   * controlling animation with Vue animation libraries.
+   */
+   forceMount?: boolean
+}
 </script>
 
 <script setup lang="ts">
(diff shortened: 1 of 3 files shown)


zernonia (approved):
Thanks for the PR @Azurency ! LGTM! 🚀

packages/radix-vue/src/Tooltip/TooltipContent.vue
```diff
@@ -23,12 +24,13 @@ const { forwardRef } = useForwardExpose()
...
 <template>
-  <component
-    :is="rootContext.disableHoverableContent.value ? TooltipContentImpl : TooltipContentHoverable"
-    v-if="rootContext.open.value"
-    :ref="forwardRef"
-    v-bind="forwarded"
-  >
-    <slot />
-  </component>
+  <Presence :present="rootContext.open.value">
```
zernonia, line 27: For `Presence` component usage we should also add a new prop `forceMount`, and set it as such. 

```suggestion
  <Presence :present="forceMount || rootContext.open.value">
```

Azurency: I added forceMount to TooltipContent 👍

jramsgz:
Hi!

I've been experiencing an issue with tooltips not appearing consistently, and after some investigation, I traced it back to this PR.

I'm using tooltips in combination with dropdowns and buttons that trigger dialogs. The problem arises when close animations are enabled for these tooltips. After the first time the tooltip is displayed, it no longer appears on subsequent interactions. It seems that the close animation is causing the tooltip to stop rendering correctly after the initial display for the tooltips in combination with dropdowns and dialogs.

Removing the close animation resolves the issue, but obviously, it removes the smooth closing effect. Could this be related to how the tooltip's animation state is being managed by Presence? I'm happy to provide more details or help test any potential fixes.

Thanks in advance!

Issue #994, opened 2024-06-10
[Bug]: "Hover Card" component is missing is-hovering check


phosmium:
### Reproduction

https://www.shadcn-vue.com/docs/components/hover-card.html

### Describe the bug

Hi. Looks like the "Hover Card" component is missing a is-hovering check. If you hover quickly over a hover card, it will popup the card without actually currently hovering.

1. Go to https://www.shadcn-vue.com/docs/components/hover-card.html
2. Quickly hover over `@vuejs`
3. Observe that the "Hover Card" will pop-up when you're not even hovering on it

This is not the same behavior as on https://ui.shadcn.com/docs/components/hover-card. This also gets messy when you have a lot of Hover Cards, as they all will pop-up

### System Info

```bash
Browsers
```

### Contributes

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests

PR #995, opened 2024-06-13
fix(HoverCard): avoid opening for quick hovers

ianfireman:
Hello!

This pr fix the bug of issue https://github.com/radix-vue/radix-vue/issues/994

A hovering ref was added to check after the open delay time if the mouse is still waiting for the content

New behavior:

https://github.com/radix-vue/radix-vue/assets/10001036/b1c84286-05b6-422b-ab27-439518f76e22

### DIFF
--- a/packages/radix-vue/src/HoverCard/HoverCardRoot.vue
+++ b/packages/radix-vue/src/HoverCard/HoverCardRoot.vue
@@ -26,8 +26,6 @@ export interface HoverCardRootContext {
   hasSelectionRef: Ref<boolean>
   isPointerDownOnContentRef: Ref<boolean>
   isPointerInTransit: Ref<boolean>
-  isHoveringRef: Ref<boolean>
-  isFocusedRef: Ref<boolean>
   triggerElement: Ref<HTMLElement | undefined>
 }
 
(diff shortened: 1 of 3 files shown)


packages/radix-vue/src/HoverCard/HoverCardRoot.vue
```diff
@@ -66,11 +67,15 @@ const closeTimerRef = ref(0)
...
 const isPointerInTransit = ref(false)
+const isHoveringRef = ref(false)
 const triggerElement = ref<HTMLElement>()
 
 function handleOpen() {
   clearTimeout(closeTimerRef.value)
-  openTimerRef.value = window.setTimeout(() => open.value = true, openDelay.value)
+  openTimerRef.value = window.setTimeout(() => {
+    if (isHoveringRef.value)
+      open.value = true
```
zernonia, line 77: Thanks for the PR @ianfireman . Adding the  `isHoveringRef` causes the `focus` from Tab doesn't open the HoverCard anymore.

The tab focus behavior should be as such:

https://github.com/radix-vue/radix-vue/assets/59365435/2e91dc39-baf9-4a61-9490-7bbbb6f3682c

ianfireman: Hmm, gonna fix it

ianfireman: @zernonia Now the behavior should be the same, thanks for the feedback!

zernonia:
Thanks again for the changes ya @ianfireman .. after digging deeper I realized it could've been more direct to use `isPointerInTransit` ref directly instead of adding another 2 refs and condition.

Do you mind if I push some of the changes to this PR? 😁

ianfireman:
Not at all! Please do it

Guiqft:
gj chefs! 🫡

PR #997, opened 2024-06-13
fix: update eslint configs

### LINKED ISSUE #996: [Bug]: ESLint config doesn't have the Vue plugin enabled
### Environment

```bash
Package manager: pnpm@8.15.4
Radix Vue version: 1.8.3
```

N/A

### Steps to reproduce

Steps to reproduce:
1. Run `pnpm lint` on any `.vue` file
2. Observe that linting may not work

### Describe the bug

Found this bug while trying to format `.vue` files on save in the editor.

According to this line: https://github.com/antfu/eslint-config/blob/523bc04549541b27f145eb8ec30fe215fd2a6fa7/src/factory.ts#L98, the eslint config does not load any `Vue` related plugins because the`package.json` file that is the closest to the `eslint.config.mjs` file doesn't have `vue` listed as a dependency, `typescript` gets loaded fine though.

### Expected behavior

Linting should work on `.vue` files.

_No response_

epr3:
Hello,

This PR fixes #996.

I've enabled both the `typescript` and `vue` plugins on the config to make sure they always get loaded.

I've also added a default `.editorconfig` file so that any editor will maintain spacing and carriage return consistency.

I've also enabled the `import` related rules (I recall those being on before ESLint 9) plus added a rule for `vue` `template` HTML attributes that makes them more readable IMO.

Looking forward to any feedback 🚀

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -15,6 +15,8 @@ export default antfu(
       'ts/consistent-type-definitions': 'off',
       'symbol-description': 'off',
       'no-console': 'warn',
+      'import/first': 'off',
+      'import/order': 'off',
       'max-statements-per-line': ['error', { max: 2 }],
       'vue/one-component-per-file': 'off',
       'unicorn/prefer-dom-node-text-content': 'off',
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -2,6 +2,10 @@ import antfu from '@antfu/eslint-config'
 
 export default antfu(
   {
+    vue: true,
+    typescript: true,
+  },
+  {
     ignores: ['*.js'],
   },
   {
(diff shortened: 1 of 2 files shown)


eslint.config.mjs
```diff
@@ -9,8 +13,6 @@ export default antfu(
       'ts/no-non-null-asserted-optional-chain': 'off',
       'ts/ban-ts-comment': 'warn',
       'ts/consistent-type-definitions': 'off',
-      'import/first': 'off',
-      'import/order': 'off',
```
zernonia, line 13: We need this 2 lines else it would mess with the way we export `provide/inject` from `vue` SFC

epr3: Fixed 👍🏻

PR #1002, opened 2024-06-16
fix(Combobox): move focus to options on keyboard navigation

### LINKED ISSUE #1001: [Bug]: Combobox options not announced by screen readers
### Environment

```bash
Node version: 16.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
Browser: Chrome 90.0.4430.212
```

https://www.radix-vue.com/components/combobox

### Steps to reproduce

1. Keep your screen reader on
2. Try navigating through the combobox options using arrow keys or home/end
3. None of the options are announced

### Describe the bug

None of the combobox options are announced by screen readers. This is due to the fact that the combobox component doesn't move focus to the option. It merely provides a visual feedback based on data attribute updates - but navigating with arrow keys doesn't move focus to the option.

### Expected behavior

Focus should be moved to the option that is being navigated through.

_No response_

abh1nash:
This fixes #1001

### DIFF
--- a/packages/radix-vue/src/Combobox/ComboboxRoot.vue
+++ b/packages/radix-vue/src/Combobox/ComboboxRoot.vue
@@ -1,5 +1,5 @@
 <script lang="ts">
-import type { Ref } from 'vue'
+import type { ComputedRef, Ref } from 'vue'
 import type { Direction } from '@/shared/types'
 import type { PrimitiveProps } from '@/Primitive'
 import { createContext, useDirection, useFormControl, useForwardExpose } from '@/shared'
@@ -26,6 +26,7 @@ type ComboboxRootContext<T> = {
   onInputNavigation: (dir: 'up' | 'down' | 'home' | 'end') => void
   onInputEnter: () => void
   selectedValue: Ref<T | undefined>
+  selectedElement: ComputedRef<HTMLElement | undefined>
   onSelectedValueChange: (val: T) => void
   parentElement: Ref<HTMLElement | undefined>
 }
@@ -226,6 +227,13 @@ function scrollSelectedValueIntoView() {
     selectedElement.value.scrollIntoView({ block: 'nearest' })
 }
 
+function focusOnSelectedElement() {
+  // Find the highlighted element and focus
+  // This helps the screen readers to read the selected value
+  if (selectedElement.value instanceof Element && selectedElement.value.focus)
+    selectedElement.value.focus()
+}
+
 provideComboboxRootContext({
   searchTerm,
   modelValue,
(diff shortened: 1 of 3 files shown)


packages/radix-vue/src/Combobox/ComboboxRoot.vue
```diff
@@ -226,6 +227,13 @@ function scrollSelectedValueIntoView() {
     selectedElement.value.scrollIntoView({ block: 'nearest' })
 }
 
+function focusOnSelectedElement() {
+  // Find the highlighted element and focus
+  // This helps the screen readers to read the selected value
+  if (selectedElement.value instanceof Element && selectedElement.value.focus)
+    selectedElement.value.focus()
+}
+
```
zernonia, line 236: We don't need this focus function as [`aria-activedescendant`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-activedescendant) already handle this

packages/radix-vue/src/Combobox/ComboboxRoot.vue
```diff
@@ -256,6 +265,9 @@ provideComboboxRootContext({
       selectedValue.value = filteredOptions.value[val === 'up' ? index - 1 : index + 1]
 
     scrollSelectedValueIntoView()
+    focusOnSelectedElement()
+
+    nextTick(() => inputElement.value?.focus({ preventScroll: true }))
```
zernonia, line 270: similarly this is not required

zernonia (changes requested):
LGTM @abh1nash ! Thanks for the issue and PR!! 💚

Some minor comment and we can get this merge! 🚀

packages/radix-vue/src/Combobox/ComboboxRoot.vue
abh1nash: I thought the same - but that doesn't work with all screen readers apparently.

packages/radix-vue/src/Combobox/ComboboxRoot.vue
abh1nash: > similarly this is not required

Same here - focus announces the option, then this returns the focus back the input field - just as how one would expect a combobox to work.

packages/radix-vue/src/Combobox/ComboboxRoot.vue
zernonia: Icic.. If that's the case then no prob 😁

zernonia (approved):

PR #1005, opened 2024-06-17
fix: infinite counting in `NumberFieldInput`

Saeid-Za:
resolving the infinite counting issue reported [here](https://github.com/radix-vue/radix-vue/issues/1006).


zernonia (approved):
LGTM! Thanks for the fix @Saeid-Za !! 🚀

PR #1009, opened 2024-06-19
fix(combobox): when backspacing searchTerm highlight the first available option

gasparrobi:
I think current highlighting functionality is unintuitive.

When the user is removing characters (eg: by backspacing) the highlighted option should always be the first matching option from the dropdown list. It should provide the same behavior as typing, but backwards.

for reference please check [shadcn/ui's combobox](https://ui.shadcn.com/docs/components/combobox) or [headleassui's combobox](https://headlessui.com/react/combobox)

this PR is fixing that behavior.

before:
combobox-before

after the change:
combobox-after

### DIFF
--- a/packages/radix-vue/src/Combobox/ComboboxRoot.vue
+++ b/packages/radix-vue/src/Combobox/ComboboxRoot.vue
@@ -211,10 +211,10 @@ watch(stringifiedModelValue, async () => {
   immediate: !props.searchTerm,
 })
 
-watch(() => filteredOptions.value.length, async (length) => {
+watch(() => [filteredOptions.value.length, searchTerm.value.length], async ([length, searchTermLength], [oldLength, oldSearchTermLength]) => {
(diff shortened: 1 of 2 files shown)


zernonia (approved):
LGTM! 🚀

packages/radix-vue/src/Combobox/ComboboxRoot.vue
```diff
@@ -211,10 +211,10 @@ watch(stringifiedModelValue, async () => {
   immediate: !props.searchTerm,
 })
 
-watch(() => filteredOptions.value.length, async (length) => {
+watch(() => [filteredOptions.value.length, searchTerm.value.length], async ([length, searchTermLength], [oldLength, oldSearchTermLength]) => {
```
timmaier, line 214: If searchTerm.value is null/undefined this causes a .length error and bugs out the Combobox

Issue #1008, opened 2024-06-19
[Feature]: Allow space keydowns in an `Editable` without submitting


ricketyBridges:
### Describe the feature

https://github.com/radix-vue/radix-vue/blob/52100dff9c33d552efc7db0181707e2d2ad6be6a/packages/radix-vue/src/Editable/EditableInput.vue#L47

Spacebar currently submits when submitMode is `enter`. Enforcing strict usage of shift-space is difficult in terms of user education, so exposing a prop to disable this or modifying to something like below would be appreciated.

``` js
if ((context.submitMode.value === 'enter' && event.key === kbd.ENTER) || (context.submitMode.value === 'both' && (event.key === kbd.SPACE || event.key === kbd.ENTER) && !event.shiftKey)) 
```

Willing to submit a PR if it's an acceptable change.

### Additional information

- [ ] I intend to submit a PR for this feature.
- [ ] I have already implemented and/or tested this feature.

epr3:
@ricketyBridges The spacebar submit implementation was me being a bit overzealous on that feature. Feel free to remove the key altogether from that conditional in L47.

PR #1013, opened 2024-06-20
fix(Editable): allow space inputs when submitMode is enter

ricketyBridges:
Context: https://github.com/radix-vue/radix-vue/issues/1008

Currently editable is very strict and submits on space input, even if submitMode is set to `enter`. This changes it to only enforce shift-space usage when submitMode is `both`.


epr3 (approved):
LGTM 👍🏻

PR #1015, opened 2024-06-20
fix(combobox): when null value passed to VisuallyHiddenInput component

true-andrew:
Faced with an issue, when null value passed to Combobox component. The parsedValue in VisuallyHiddenInput doesn't provide correct check for null value

### DIFF
--- a/packages/radix-vue/src/VisuallyHidden/VisuallyHiddenInput.vue
+++ b/packages/radix-vue/src/VisuallyHidden/VisuallyHiddenInput.vue
@@ -28,7 +28,7 @@ const parsedValue = computed(() => {
   }
 
   // if object value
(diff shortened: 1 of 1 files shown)


packages/radix-vue/src/VisuallyHidden/VisuallyHiddenInput.vue
```diff
@@ -28,7 +28,7 @@ const parsedValue = computed(() => {
   }
 
   // if object value
-  else if (typeof props.value === 'object' && !Array.isArray(props.value)) {
+  else if (typeof props.value === 'object' && props.value !== null) {
```
zernonia, line 31: We need `!Array.isArray(props.value)` to make sure this value is type object and not an array.

zernonia:
Thanks for the PR @true-andrew ! 🚀

PR #1018, opened 2024-06-21
fix(MenuRadioGroup): update:modelValue payload type

tristandubbeld:
When using the `DropdownMenuRadioGroup` using `@update:model-value` instead of binding `v-model` we noticed that the payload was type `boolean` when we expected `string`. This type is also in line with similar components (`RadioGroupRoot`).

With this PR the type is updated to reflect the payload value, along with the docs.


zernonia (approved):
LGTM! Thansk for the fix @tristandubbeld ! 🚀

PR #1024, opened 2024-06-22
fix(Pagination): add conditions for 2 siblings edge case

### LINKED ISSUE #970: [Bug]: Pagination hides a single page behind an ellipsis
### Environment

```bash
Node version: 16.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
Browser: Chrome 90.0.4430.212
```

https://stackblitz.com/edit/xbu9yg?file=src%2FApp.vue

### Steps to reproduce

PaginationRoot settings: `<PaginationRoot :total="90" :sibling-count="2" show-edges>`

Navigate between pages to notice the behavior.

### Describe the bug

In the example, pages 2 and 8 depending on proximity to current page get replaced with ellipsis (...) which could be always shown as page numbers since the ellipsis in this case do not conserve any space.

### Expected behavior

Show the page number instead of ellipsis if it replaces only a single page number.

image

image

image

epr3:
Fixes #970.

### DIFF
--- a/packages/radix-vue/src/Pagination/utils.ts
+++ b/packages/radix-vue/src/Pagination/utils.ts
@@ -42,17 +42,17 @@ export function getRange(currentPage: number, pageCount: number, siblingCount: n
       // default condition
       = leftSiblingIndex > firstPageIndex + 2
       // if the current page is towards the end of the list
-      && lastPageIndex - itemCount - firstPageIndex + 1 > 2
+      && Math.abs(lastPageIndex - itemCount - firstPageIndex + 1) > 2
       // if the current page is towards the middle of the list
-      && leftSiblingIndex - firstPageIndex > 2
+      && Math.abs(leftSiblingIndex - firstPageIndex) > 2
 
     const showRightEllipsis
       // default condition
       = rightSiblingIndex < lastPageIndex - 2
       // if the current page is towards the start of the list
-      && lastPageIndex - itemCount > 2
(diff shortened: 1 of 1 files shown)


packages/radix-vue/src/Pagination/utils.ts
```diff
@@ -41,18 +38,37 @@ export function getRange(currentPage: number, pageCount: number, siblingCount: n
 
     const itemCount = totalPageNumbers - 2 // 2 stands for one ellipsis and either first or last page
 
+    const showLeftEllipsis
+      // default condition
+      = leftSiblingIndex > firstPageIndex + 2
+      // if the current page is towards the end of the list
+      && lastPageIndex - itemCount - firstPageIndex + 1 > 2
+      // if the current page is towards the middle of the list
+      && leftSiblingIndex - firstPageIndex > 2
```
zernonia, line 47: Haven't test this out, but the inequality seems like the same as line 43 `leftSiblingIndex > firstPageIndex + 2`

epr3: Yep, my mistake, I adjusted the condition and added `Math.abs` to all calculations to make sure the conditions are always evaluated properly.

zernonia (approved):
LGTM!! 🚀

PR #1025, opened 2024-06-22
feat(Editable): expose editable functions

epr3:
Hello,

This PR exposes the internal functions on the `EditableRoot` slot to enable programmatic control of the internal state of the `Editable` component.

Should address the use case in #1021.

### DIFF
--- a/packages/radix-vue/src/Editable/EditableRoot.vue
+++ b/packages/radix-vue/src/Editable/EditableRoot.vue
@@ -166,8 +166,11 @@ function handleDismiss() {
 }
 
 defineExpose({
+  /** Function to submit the value of the editable */
   submit,
+  /** Function to cancel the value of the editable */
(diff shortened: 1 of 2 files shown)


packages/radix-vue/src/Editable/EditableRoot.vue
```diff
@@ -199,6 +204,9 @@ provideEditableRootContext({
         :model-value="modelValue"
         :is-editing="isEditing"
         :is-empty="isEmpty"
+        :submit="submit"
+        :cancel="cancel"
+        :edit="edit"
```
zernonia, line 209: let's expose these function in `defineExpose` too 😁

epr3: Functions added to `defineExpose` 👍🏻

PR #1029, opened 2024-06-23
fix(Calendar): sometimes first day of next month is missing

xollaborator:
When fixedWeeks is true and next month starts in another row of Calendar first day was missing:
before

After the fix:
after

Fix is quite simple but I've also extended the test case and added a new story for fixedWeeks. 

I do not know if you want to have all fixes covered by a test case but taking in to account that on my machine Calendar tests took 15 seconds before and now 2 seconds were added I would personally not merge my test.

I think the new story is also not providing that much value but it helped me test the fix so I included it as well.

Thank you for this library and for shadcn/vue :)

### DIFF
--- a/packages/radix-vue/src/Calendar/Calendar.test.ts
+++ b/packages/radix-vue/src/Calendar/Calendar.test.ts
@@ -250,21 +250,19 @@ describe('calendar', async () => {
       return calendar.querySelectorAll('[data-week]').length
     }
 
-    function getFirstDayOfNextMonth() {
-      return calendar.querySelector('[data-week]:nth-child(n+4) [data-outside-view]') as HTMLElement
-    }
-
     const nextButton = getByTestId('next-button')
+
+    expect(getByTestId('date-1-1')).toHaveTextContent('1')
 
     for (let i = 0; i < 12; i++) {
       expect(getNumberOfWeeks()).toBe(6)
-      expect(getFirstDayOfNextMonth()).toHaveTextContent('1')
       await user.click(nextButton)
     }
 
+    expect(getByTestId('date-2-1')).toHaveTextContent('1')
+
     for (let i = 0; i < 24; i++) {
       expect(getNumberOfWeeks()).toBe(6)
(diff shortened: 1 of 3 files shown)


epr3:
Great catch @xollaborator! Thank you for your contribution!👍🏻 

This functionality and the date utilities are ported from [Melt UI](https://github.com/melt-ui/melt-ui), therefore this fix would help them too.

Here's the corresponding line in Melt UI: https://github.com/melt-ui/melt-ui/blob/42e2dd430c3e13f56062c4d610dabd84ea4e5517/src/lib/internal/helpers/date/calendar.ts#L97

epr3 (approved):

packages/radix-vue/src/Calendar/Calendar.test.ts
```diff
@@ -252,14 +252,14 @@ describe('calendar', async () => {
 
     const nextButton = getByTestId('next-button')
 
-    expect(getByTestId('date-1-1')).toHaveTextContent('1')
+    expect(getByTestId('date-8-1')).toHaveTextContent('1')
```
xollaborator, line 255: @epr3 Please notice this test case starts now in August, so this `expect` is checking only that August 1st is visible. The issue was that September 1st was skipped. Another `expect` is checking September, but that is September 2025 (because of `for` loop), which does not start on Sunday so the error would not happen there. I think you could move second `expect` to replace the first one. Probably adding a comment will make it easier for someone in the future to understand what was the intent here.

If you would like to further improve this test speed you could consider leaving only one `for` loop. I've calculated that there will be max 13 months at any point between situations that trigger this issue, so maybe `i < 14` will cover enough months for what was originally tested during those 36 months verified currently.

xollaborator:
> This functionality and the date utilities are ported from [Melt UI](https://github.com/melt-ui/melt-ui), therefore this fix would help them too.

Sure, I might try to help them, especially that there is already bug for this issue: https://github.com/melt-ui/melt-ui/issues/1213.

PR #1033, opened 2024-06-24
chore: optimize start-up command to automatically open browser with story:dev

yiwwhl:
When I launch story, the console messages push the browser access link to the top, which impacts the debugging experience


zernonia (approved):
Thanks for sharing the friction during local dev and the PR fix! 🚀

PR #1035, opened 2024-06-24
feat(combobox): focusEl logic implemented

### LINKED ISSUE #744: [Feature]: Combobox highlight (?) event
### Describe the feature

First of all: Thank you guys for your hard work! The library is absolutely amazing!

To my feature reuqest though:
Currently it's not possible (as far as i can tell) to receive an event if or what ComboboxItem is highlighted. You can of course use mouseover, but then it still does not work with hotkeys. In my case, i'd like to (pre)fetch some data based on the current selection, but without the need to actually select the item (for previews for example).

 I've looked into the code and I think most of what's needed is already present. I think we'd just need to expose the event and/or the highlighted selection.
If you guys need help I could try and submit a PR.

Thanks a million already!

### Additional information

romanhrynevych:
This PR fix #744. Added `focusEl` v-model, if there is some additional questions feel free to ask 🙌 

https://github.com/radix-vue/radix-vue/assets/82209198/c9ad99be-8ded-4224-bff6-4cee09d39a14

### DIFF
--- a/docs/components/demo/Combobox/css/index.vue
+++ b/docs/components/demo/Combobox/css/index.vue
@@ -5,6 +5,7 @@ import { Icon } from '@iconify/vue'
 import './styles.css'
 
 const v = ref('')
+const focusEl = ref('')
 const options = ['Apple', 'Banana', 'Blueberry', 'Grapes', 'Pineapple']
 const vegetables = ['Aubergine', 'Broccoli', 'Carrot', 'Courgette', 'Leek']
 </script>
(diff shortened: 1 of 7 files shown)


romanhrynevych:
@zernonia thanks for your great work on this project ❤️  When you have some free time can you take a look if this will be ok or maybe something to change? I don't know if we need `focusEl` prop, or just `emit`. 

Mark this PR as Draft because it needs some more work but seems good for me)

zernonia:
Thanks for the PR @romanhrynevych .. This will be implemented in `v2` (as labelled in the issue above) when wrapping `Combobox` with `Listbox` component. So I will not be merging this ya 😁

PR #1038, opened 2024-06-25
fix(Editable): do not submit if editable is not in edit mode

### LINKED ISSUE #1036: [Bug]: Editable submits on every single click
### Environment

```bash
Node version: 20.x
Package manager: pnpm@8.6.0
Radix Vue version: 1.8.5
```

https://stackblitz.com/edit/qeo66a-zdntmx

### Steps to reproduce

1. assign event listener to the event `submit` at EditableRoot
2. click anywhere in your app
3. see event being triggered

### Describe the bug

The event is emitted on every click.

### Expected behavior

The event should only be emitted when a new value is submitted.

_No response_

epr3:
Fixes #1036.

### DIFF
--- a/packages/radix-vue/src/Editable/EditableRoot.vue
+++ b/packages/radix-vue/src/Editable/EditableRoot.vue
@@ -159,10 +159,12 @@ function submit() {
 }
 
 function handleDismiss() {
-  if (submitMode.value === 'blur' || submitMode.value === 'both')
-    submit()
-  else
-    cancel()
+  if (isEditing.value) {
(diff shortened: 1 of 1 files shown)


Tinoooo:
Is there a practical way to remove the whole event listener (when `isEditing === true`) instead of ignoring the event? This might be more performant.

epr3:
@Tinoooo which event listener?

Tinoooo:
@epr3 The one in DismissableLayer.

epr3:
Oh, got it. Will look into a solution 👍🏻

PR #1041, opened 2024-06-26
fix(Editable): use exposed edit function to trigger edit

iammola:
So the `update:state` event is triggered with the `edit` state


epr3 (approved):
LGTM :+1:

PR #1042, opened 2024-06-26
Refactor(Editable): remove DismissableLayer dependency

epr3:
Hello,

This idea came through from the discussion here: https://github.com/radix-vue/radix-vue/pull/1038#issuecomment-2191731547.

This PR removes the dependency on `DismissableLayer` since that is more addressed towards, well, dismissable components such as `Popover`, `Tooltip`, etc.

The `Editable` component already has an `update:state` and `submit` event which would make redundant the emits from `DismissableLayer` for monitoring interactions outside the `Editable` component.

Any feedback is appreciated. 🚀


zernonia:
Using `@focusout` has some behavior changes, which can be abit annoying for some user. The video below shows that if I `mousedown` on the `Cancel` button, it immediately closes it, while previously it only closes on `mouseup`

https://github.com/radix-vue/radix-vue/assets/59365435/6e53fc8e-32f7-432d-b7db-39d81e786e62

epr3:
@zernonia I refactored a bit the component. I ended up pulling in the functions from the `DismissableLayer`, namely `useFocusOutside` and `usePointerDownOutside`, to consistently track outside interactions for `Editable`, but without emitting any events.

PR #1048, opened 2024-06-27
feat: expose highlighted/selectedValue in combobox root

HuakunShen:


zernonia:
Thanks for the PR @HuakunShen . We can merge this feature in, but will be deprecated in `v2`. 😁

HuakunShen:
@zernonia Thanks, I just need to use it now.

zernonia:
Yup. There will be `@highlight` event which has the value and html for the current highlighted value. Similar to https://www.radix-vue.com/components/listbox.html#root

HuakunShen:
@zernonia Done

PR #1086, opened 2024-07-01
feat(Calendar): handle default slot of CalendarCellTrigger

xollaborator:
This was started as a fix but correct fix is more like a feature.

CalendarCellTrigger is providing a slot but in case of using nested elements the logic in `handleClick` was failing because `event.target` did not have `data-value` to get value from. Instead od reading from DOM a value from `props.day` might be used which also simplifies logic.

To further improve DX for such scenario `dayValue` prop is added to the default scope. This will match behavior of CalendarHeading where year and month are provided.


epr3 (approved):
LGTM 👍🏻

PR #1088, opened 2024-07-02
fix(Combobox,Listbox): `update:modelValue` with `multiple:true` does not get emitted

vojvodics:
When using `v-model` with `multiple:true`, the `update:modelValue` event doesn't get emitted. Please see the reproduction link: https://codesandbox.io/p/sandbox/radix-vue-combobox-forked-9pqcs3

I noticed this behavior on Listbox and Combobox components. Searching through the code, there doesn't seem to be an issue with other components that accept single or multiple values as vModel. 
Alternatively, Listbox and Combobox components could be refactored to use `useSingleOrMultipleValue` composable (like ToggleGroup or Accordion use), however, the fix was pretty simple, and refactoring would take a lot more time (and I'm also not sure if it is a direction you'd want to go in), so I opted to just fix the problem.

When writing tests for the Listbox with `selectionBehavior: replace`, I found another bug - the `update:modelValue` event get's emitted twice for the same value (I only changed the code for `selectionBehavior: toggle`), but I couldn't find a root cause of the issue.

### DIFF
--- a/packages/radix-vue/src/Combobox/Combobox.test.ts
+++ b/packages/radix-vue/src/Combobox/Combobox.test.ts
@@ -10,6 +10,7 @@ import { handleSubmit } from '@/test'
 describe('given default Combobox', () => {
   let wrapper: VueWrapper<InstanceType<typeof Combobox>>
   let valueBox: DOMWrapper<HTMLInputElement>
+  let items: DOMWrapper<Element>[]
   window.HTMLElement.prototype.releasePointerCapture = vi.fn()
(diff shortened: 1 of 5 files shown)


zernonia:
Thanks for the PR and quick fix @vojvodics ! 
Yes we probably can use `useSingleOrMultipleValue` to handle this, but right now that will be abit complicated because of the types.

And the case for `selectionBehavior` is abit tricky too. I will have a look perhaps in v2 😁

PR #1108, opened 2024-07-08
feat: global nonce value via `ConfigProvider`

### LINKED ISSUE #1097: [Feature]: Global config/provider for Content-Security-Policy (CSP) `nonce` value
### Describe the feature

I'd like radix-vue to offer a way to globally configure `nonce` for Content-Security-Policy (CSP) hardened environments. 

Currently we have `ComboboxViewport`, `ScrollAreaViewport` and `SelectViewport` that inject inline styles and these components now offer `nonce` prop (implemented in #721). 

It's a little bit error prone to always remember to set this prop each time for these components (or make sure to always use wrapper that sets it internally). Also even though it doesn't happen often but existing/new components may begin to inject more styles of their own.

I'm proposing to add `NonceProvider` component (similar to `TooltipProvider`) that allows to share nonce value across the whole component tree. 

### Additional information

romansp:
Resolves #1097.

Implemented via existing `ConfigProvider` and additional `useNonce` composable to be used in primitives. Added test to make sure `nonce` still can be overridden locally in place to maintain backwards compatibility.

Also added description to `nonce` prop in related primitives and regenerated docs.

### DIFF
--- a/docs/content/meta/ComboboxViewport.md
+++ b/docs/content/meta/ComboboxViewport.md
@@ -16,7 +16,7 @@
   },
   {
     'name': 'nonce',
-    'description': '',
+    'description': '<p>Will add <code>nonce</code> attribute to the style tag which can be used by Content Security Policy. &lt;br&gt; If omitted, inherits globally from <code>ConfigProvider</code>.</p>\n',
     'type': 'string',
     'required': false
   }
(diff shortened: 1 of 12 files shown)


zernonia (approved):
Nicely done @romansp ! Thanks for using the same approach as other ConfigProvider's composable!

PR #1109, opened 2024-07-08
docs: update mentions of `DirectionProvider` to `ConfigProvider`

romansp:
Noticed this inconsistency in docs as I was working on #1108.

I believe this is just a remnant of Radix UI documentation. We don't have a dedicated `DirectionProvider` and `dir` is configured globally via `ConfigProvider`.


zernonia (approved):
Thanks for capturing that @romansp !! 🙏🏻

PR #1113, opened 2024-07-09
fix(Splitter): splitter-panel works based on the old layout after props change

### LINKED ISSUE #1112: [Bug]: `SplitterPanel` works based on the old layout after props change
### Environment

```bash
Node version: 18.18.2
Package manager: pnpm@8.15.4
Radix Vue version: 1.9.0
Vue version: 3.4.31
```

https://stackblitz.com/edit/vitejs-vite-u4u7kq?file=src%2FApp.vue

### Steps to reproduce

1. Change `SplitterPanel` props.
2. Collapse `SplitterPanel`.
3. We will find that `SplitterPanel` works based on the old layout.

### Describe the bug

We didn't update `eagerValuesRef` when props changed.
All functions work based on the old props.
https://github.com/radix-vue/radix-vue/blob/3aaa11134dcd73a3afdea24b316b832c35024dcf/packages/radix-vue/src/Splitter/SplitterPanel.vue#L110

### Expected behavior

_No response_

_No response_

AIchanKimi:
fix #1112 
after I update `eagerValuesRef` . `Splitter` will be reset to default size. I'm not sure is this a bug or feature.

### DIFF
--- a/packages/radix-vue/src/Splitter/SplitterGroup.vue
+++ b/packages/radix-vue/src/Splitter/SplitterGroup.vue
@@ -415,7 +415,9 @@ function resizePanel(panelData: PanelData, unsafePanelSize: number) {
 
 function reevaluatePanelConstraints(panelData: PanelData, prevConstraints: PanelConstraints) {
   const { layout, panelDataArray } = eagerValuesRef.value
-
+  const index = findPanelDataIndex(panelDataArray, panelData)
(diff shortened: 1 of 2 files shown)


zernonia:
Thanks for the PR @AIchanKimi . The `Splitter` would return to default size because it has to recompute the position, and given that any Splitter might be added/removed, we use default size to properly compute it.

PR #1117, opened 2024-07-09
Fix unplugin-vue-components installation instruction

holtergram:
Following the instructions here (https://github.com/unplugin/unplugin-vue-components) I think we should use the plural, no?


zernonia:
Althought it's not a big factor, agree that it should be `plural` so that it's consistent with the instructions in `unplugin-vue-components` 😁

PR #1118, opened 2024-07-10
Fix the aria-selected attribute for SelectItem

kilobyte2007:
The bug was that the aria-selected was only true when the option was also highlighted (isFocused). I looked at how this is done in Combobox (where it's working correctly) and adjusted it accordingly.


zernonia (approved):
Thanks @kilobyte2007 ! After inspecting deeper and in fact you are right. The previously condition was adapted from Radix UI and is apparently wrong. The `aria-selected` should only based on the current select item, and not depends on the highlighted state.

Thanks for capturing that and the fix!

PR #1123, opened 2024-07-12
fix(Combobox): workaround for maximum recursion update issues

### LINKED ISSUE #1040: [Bug]: Combobox breaks down when there are more than 100 options
### Environment

```bash
Developement/Production OS: Arch Linux x86_64 6.9.2-arch1-1
Node version: 22.1.0
Package manager: yarn@1.22.22
Radix Vue version: 1.8.4+
Vue version: 3.4.30
Nuxt version: N/A
Nuxt mode: N/A
Nuxt target: N/A
CSS framework: None
Client OS: Arch Linux x86_64 6.9.2-arch1-1
Browser: Chromium Version 125.0.6422.112 (Official Build)
```

https://stackblitz.com/edit/cbpual?file=src%2FApp.vue

### Steps to reproduce

- Create a combobox with 100+ options
- Open the combobox
- Pick an option

### Describe the bug

A warning is show in console, and selecting the option does not make the value show as selected.

### Expected behavior

No warning to be logged, and the selector to work.

This issue existed before and was fixed (#551), but it appears there is a regression in 1.8.4 (works just fine in 1.8.3) causing the issue to occur again.

cyyynthia:
This is a hacky solution that fixes #1040 and works fine with 100+ options in a combobox. As v2 is being developed and the combobox gets reworked to better handle these cases, I assume a better solution will be implemented - the purpose of this fix is to unblock people stuck on 1.8.3 due to this bug.

### DIFF
--- a/packages/radix-vue/src/Combobox/Combobox.test.ts
+++ b/packages/radix-vue/src/Combobox/Combobox.test.ts
@@ -1,6 +1,7 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { axe } from 'vitest-axe'
 import Combobox from './story/_Combobox.vue'
+import Combobox1k from './story/_Combobox1k.vue'
 import ComboboxObject from './story/_ComboboxObject.vue'
 import type { DOMWrapper, VueWrapper } from '@vue/test-utils'
 import { mount } from '@vue/test-utils'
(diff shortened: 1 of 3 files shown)


zernonia (approved):
Thanks for the PR @cyyynthia ! Appreciate your help with this workaround for now 😁

PR #1146, opened 2024-07-19
fix(Dates): disable state not affect tabindex and relevant button

zernonia:
- remove tabindex when disabled
- disable button when the root was disabled
- prevent focusable on cell trigger when disabled


epr3 (approved):
LGTM 🚀

PR #1149, opened 2024-07-21
feat(Tree): allow custom children schema

### LINKED ISSUE #1139: [Feature]: allow custom children list in `<TreeRoot />`
### Describe the feature

Currently, `Tree` component requires you to provide data in a certain format. Specifically, it requires you to specify children as a `children`.

Take the file system schema as an example:

```ts
interface File {
  name: string;
  type: FileType;
}

interface Directory {
  name: string;
  files: File[];
  directories: Directory[];
}
```

I can convert `files` and `directories` into `children` on my end, but I will lose some of the type-safety. Also, converting recursive data structures is not convenient or error-prone.

As a solution, consider adding a `getChildren` prop to the `<Tree.Root />` similar to a `getKey` prop.

### Additional information

Destiner:
Adds a new optional prop to `<TreeRoot />`, `getChildren`, which allows the user to provide a custom list of children. This is helpful when you have a custom data schema.

Fixes #1139.

Also adds new tests and updates the docs to account for the change.

### DIFF
--- a/docs/content/components/tree.md
+++ b/docs/content/components/tree.md
@@ -275,17 +275,16 @@ In `CustomTree.vue`
 
 ### Custom children schema
 
-By default, `<TreeRoot />` expects you to provide the list of node's children by passing a list of `children` for every node. You can overwrite that by providing the `getChildren` prop.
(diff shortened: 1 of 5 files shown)


zernonia (approved):
Nicely done @Destiner ! Appreciate for adding the test too! 💚

PR #1171, opened 2024-07-30
feat(Combobox): narrow modelValue as readonly array type

iamandrewluca:
When a readonly array is passed as modelValue, it will error. This change will arrow a readonly array as modelValue. Internally Combobox modelValue is immutable.

```
Type 'readonly Option<T>[]' is not assignable to type 'Option<T> | Option<T>[] | undefined'.ts-plugin(2322)
```


zernonia:
@iamandrewluca  what's the reason you are passing `readonly` to `modelValue`?

iamandrewluca:
I try to keep my code as immutable and static as possible. I was looking through Combobox internals, and all the code is not doing any mutations to the state itself.

zernonia:
We try to keep the types as close to the `v-model`'s types from Vue, so this addition might not be inline with the idea

iamandrewluca:
I see 🤔 in this case, I will close the PR. Sorry for the inconvenience, and thanks for your attention.

Issue #920, opened 2024-05-13
[Bug]: Collapsible open animation is triggered by splice on underlying data


ricketyBridges:
### Environment

```bash
Development/Production OS: macOS Sonoma 14.4
Node version: 18.18.0
Radix Vue version: ^1.5.3
Vue version: ^3.4.15
CSS framework: Vanilla scss
```

### Link to minimal reproduction

https://codesandbox.io/p/sandbox/collapsible-flicker-x59g7c

### Steps to reproduce

Open a bunch of the collapsibles via the buttons. Then drag and drop to rearrange them.

### Describe the bug

Using splice to rearrange an array of collapsibles causes the open animation to fire or flicker. 

This occurs on drag start and when a drag ends (but only on the collapsible that is higher up). If all collapsibles start open, it does not occur, user must close/open first.

### Expected behavior

If the dragged/spliced collapsible was previously open, it should stay open and not flicker.

### Context & Screenshots (if applicable)

Seems similar to issue https://github.com/radix-ui/primitives/issues/1074#issuecomment-2002451130
![313472106-b7dedbcd-92f3-46dc-bba9-b0aa97f09ad7](https://github.com/radix-vue/radix-vue/assets/117129982/477145ef-9239-4d5f-94b1-a189e2bbe75c)

zernonia:
Does the solution mentioned here fix it? https://github.com/radix-ui/primitives/issues/1074#issuecomment-2067649324

ricketyBridges:
Not sure if that solution is applicable here, as I cannot move the class into an inner element without losing animation completely.

ricketyBridges:
Any movement on this? Conditionally disabling animation via a class merely delays the animation, it does not stop it from triggering once `isDragging` (or similar condition) is false.

miloscript:
```
      keyframes: {
        "accordion-down": {
          from: {
            maxHeight: "0",
          },
          to: {
            maxHeight: "var(--radix-accordion-content-height)",
          },
        },
        "accordion-up": {
          from: {
            maxHeight: "var(--radix-accordion-content-height)",
          },
          to: {
            maxHeight: "0",
          },
        },
```

using maxHeight will solve the issue

neunzehn86:
I was able to reproduce the problem and track it down further.

In my case, I am also using the component inside a drag list.

If I never interact with the collapsible (i.e. I do not close and reopen it), there is no flicker after dragging.

[...]

PR #1174, opened 2024-08-01
feat(Collapsible): add disableAnimation to block animation

kaishuige:
feat: #920


kaishuige:
#920

zernonia:
Thanks for the PR @kaishuige , but I don't see the benefit for adding `disableAnimation` props. The user can disable animation by using `class` binding based on condition as well. 

```html
  :class="{ 'animate-open': !isDragging  }"
```

kaishuige:
I'm the one who missed the class attribute.

kaishuige:
> Thanks for the PR @kaishuige , but I don't see the benefit for adding `disableAnimation` props. The user can disable animation by using `class` binding based on condition as well.
> 
> ```
>   :class="{ 'animate-open': !isDragging  }"
> ```

The disableAnimation parameter can be resolved #920, The class attribute doesn't seem to work.

PR #1180, opened 2024-08-04
fix(Menu): reduce amount of created listeners

### LINKED ISSUE #1176: [Bug/Performance]: MenuRoot creates too many event listeners per keystroke
### Environment

```bash
Development/Production OS: Windows 10 (WSL2)
Node version: 20.13.1
Package manager: pnpm@8.15.5
Radix Vue version: 1.9.2
Vue version: 3.4.35
CSS framework: tailwindcss@3.4.7
Client OS: Windows 10
Browser: Microsoft Edge Version 127.0.2651.74 (Official build) (64-bit)
```

https://stackblitz.com/edit/hnxppu?file=src%2Fmain.ts
(if profiling, check the preview frame, and then you'll find a ton of method calls happening per keydown event)

### Steps to reproduce

Type in the input box and see the listener count in the dev tools profiler go up for no good reason.

### Describe the bug

In my app I have a lot of potential menus that can be opened. We're talking three digits here. Radix-vue adds one-off event listeners on every keystroke which adds a ton of new event listeners and causes a lot of memory usage (that has to be GCed).

### Expected behavior

Radix-Vue not creating that many event listeners. Maybe a shared event pool could help?

From what I can tell the lines that create the event listeners are:
https://github.com/radix-vue/radix-vue/blob/562c896ba609028a7fbd5d343aeb70fcde26f211/packages/radix-vue/src/Menu/MenuRoot.vue#L78-L85

Here are some profiler shots, micro view:
image

and macro view:
image

The images show a range of around 4k listeners suddenly jumping up to 23k listeners within 3 seconds as I am typing in my app.

romansp:
Fixes #1176.

I extracted `isUsingKeyboard` into a separate composable and wrapped it in [`createSharedComposable`](https://vueuse.org/shared/createSharedComposable/#createsharedcomposable). This should ensure only single copy is created and reused between all instances of `MenuRoot`.

`onMounted` makes it so event handlers won't be created when running on the server. Added simple test to ensure value correctly toggles between true/false on keydown/pointermove/pointerdown events.

Tested it manually with devtools perf view: amount of JS handlers stays constant on typing.

### DIFF
--- a/packages/radix-vue/src/Menu/MenuRoot.vue
+++ b/packages/radix-vue/src/Menu/MenuRoot.vue
@@ -2,6 +2,7 @@
 import type { Ref } from 'vue'
 import type { Direction } from './utils'
 import { createContext, useDirection } from '@/shared'
+import { useIsUsingKeyboard } from '@/shared/useIsUsingKeyboard'
 
 export interface MenuContext {
   open: Ref<boolean>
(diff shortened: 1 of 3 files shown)


zernonia (approved):
Simple & Elegant solution @romansp ! Thanks for picking this up! 🔥

spaceemotion:
Thanks a bunch for the quick implementation/resolution, much appreciated! Do you have a time window for when the next version will be out containing this fix?

Issue #1188, opened 2024-08-05
[Bug]: `PopperAnchor` prop `element` isn't really reactive


zumm:
### Environment

```bash

```

### Link to minimal reproduction

-

### Steps to reproduce

-

### Describe the bug

Watcher in `PopperAnchor.vue` wachtes only over `currentElement`. So if you pass falsy value for `element` or change it after `PopperAnchor` mounted it will not be used.

https://github.com/radix-vue/radix-vue/blob/main/packages/radix-vue/src/Popper/PopperAnchor.vue#L24

### Expected behavior

Watcher should watch over both `element` and `currentElement`. Actually better to use `watchEffect` for this.

PR #1189, opened 2024-08-05
fix(Popper): use `watchEffect` for anchor update #1188

zumm:
fix: #1188


zernonia (approved):
Appreciate the ticket and PR @zumm ! LGTM! 🚀

Issue #1190, opened 2024-08-06
[Bug]: Combobox doesn't render the selected item when using objects as values


Destiner:
### Environment

```bash
Developement/Production OS: macOS Sonoma (14.4.1)
Node version: v18.20.3
Package manager: npm@10.2.3
Radix Vue version: 1.9.3
Vue version: 3.4.35
Client OS: macOS Sonoma (14.4.1)
Browser: Chrome 127.0.6533.89
```

### Link to minimal reproduction

https://stackblitz.com/edit/vitejs-vite-bdpgs4

### Steps to reproduce

When binding objects (second combobox):

1. Click on the "chevron" icon
2. Select an option

### Describe the bug

Input placeholder is not replaced with the selected value.

### Expected behavior

Input placeholder is replaced with the selected value. 

Compare this to using string values as a binding (first combobox)

romansp:
I think it's really a documentation issue. Afaik, in case of object binding `display-value` prop must be be passed to `ComboboxRoot`. It's used to get string value of selected item.
https://github.com/radix-vue/radix-vue/blob/9c665d687387830c4047940e2aa847771e8902c4/packages/radix-vue/src/Combobox/ComboboxRoot.vue#L194-L196

Would be nice if Combobox could produce a warning in case of object value but missing `display-value`.

Fixed repro: https://stackblitz.com/edit/vitejs-vite-ihsric?file=src%2Fcomponents%2FComboboxObjects.vue

Destiner:
@romansp thanks! I think you're right. Setting `display-value` fixed it for me.

I'll send a PR to clarify the docs.

PR #1192, opened 2024-08-07
docs(Combobox): update "Binding objects as values" example

Destiner:
Updates an example in the docs  to clarify the need for `displayValue` prop when binding an object.

Helps to avoid confusion e.g. #1190.


zernonia:
Make perfect sense! Thanks for the PR and issue @Destiner !! 💚

PR #1193, opened 2024-08-07
fix(Popper): use `watchPostEffect` for `placed` update

maxmaxme:
**Problem**

When floating-ui returned isPositioned=true, radix-vue immediately emitted an event to focus on the selected item in the Select component. However, at this point, the PopperContent element did not yet have the correct transform CSS property set, resulting in the Popper being positioned at coordinates (0, -200%).

**Solution**

I changed the watchEffect to watchPostEffect to ensure that the event is emitted only after the DOM has been updated and the Popper is correctly positioned on the page (e.g., at coordinates 84px, 231px). This ensures that focusSelectedItem is called only after the Popper has been properly positioned.

**Changes**
- Replaced watchEffect with watchPostEffect to emit the positioning event after DOM updates.

### DIFF
--- a/packages/radix-vue/src/Popper/PopperContent.vue
+++ b/packages/radix-vue/src/Popper/PopperContent.vue
@@ -140,7 +140,7 @@ export const [injectPopperContentContext, providePopperContentContext]
 </script>
 
 <script setup lang="ts">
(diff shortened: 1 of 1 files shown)


zernonia (approved):
LGTM! 🔥

PR #1197, opened 2024-08-07
fix(DateField): emit undefined on segment clear after input is filled

### LINKED ISSUE #1195: [Bug]: DatePicker value stops synchronising when input is invalid/empty
### Environment

```bash
Developement/Production OS: macOS 14.5
Node version: 20.15.1
Package manager: pnpm@9.5.0
Radix Vue version: 1.9.2
Vue version: 3.4.34
Nuxt version: 3.12.4
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.4.7
Client OS: macOS 14.5
Browser: Chrome 127.0.6533.99
```

https://stackblitz.com/edit/8yfcwi?file=src%2FApp.vue

### Steps to reproduce

- Enter a valid Date in the Input
- Start deleting the year value with backspace.
- Notice that when the last digit in the year segment gets deleted the value doesn't update anymore and keeps a date with one digit for the year (if you inputed for Example "01/01/2024" the value will be stuck at "01-01-0002").

### Describe the bug

The v-model value in the DatePicker doesn't update when one of the Segments has all its digits deleted and keeps the last valid date value.

(I would be willing to help and submit a PR for this, but looking at the code I'm lost and would need some guidance)

### Expected behavior

I would expect that the v-model gets set to `undefined` when the user didn't input a complete date value or that I can somehow access the "raw" value that the user is seeing on their end.

We are trying to use the DatePicker in a Form and our validation only sees that the Date is set, even when the User deleted the value. Which leads to confusing behaviour when the Form is submitted and we are using the "stuck" date (e.g. "01-01-0002").

Before deleting the last digit:
before

After deleting the last digit:
after

epr3:
Fixes #1195.

### DIFF
--- a/packages/radix-vue/src/DateField/DateFieldRoot.vue
+++ b/packages/radix-vue/src/DateField/DateFieldRoot.vue
@@ -123,7 +123,6 @@ onMounted(() => {
 
 const modelValue = useVModel(props, 'modelValue', emits, {
   defaultValue: defaultValue.value,
-  passive: (props.modelValue === undefined) as false,
 }) as Ref<DateValue>
 
 const defaultDate = getDefaultDate({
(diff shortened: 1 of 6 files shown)


zernonia:
Removing the `year` reset the value to `undefined` correctly. However for `day`,`month` it doesn't reset properly


https://github.com/user-attachments/assets/900f2465-0e82-46d0-843c-9de5391282e8

zernonia (approved):
LGTM! 🔥

PR #1202, opened 2024-08-08
feat(NumberField): add handleIncrease/handleDecrease for custom multipliers to custom buttons

sadeghbarati:
Close #1194

I might be mistaken, so please do a review 🙌


sadeghbarati:
😅

Issue #1204, opened 2024-08-08
[Bug]: Number Field Doesn't working well  if locale is 'ar'


peter-emad99:
### Environment

```bash
Developement/Production OS: Windows 10 19043.1110
Node version: 16.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.3.3
Client OS: Windows 10 19043.1110
Browser: Chrome 90.0.4430.212
```

### Link to minimal reproduction

https://stackblitz.com/edit/9vt2xh?file=src%2FApp.vue

### Steps to reproduce

use these props on NumberFieldRotte Component
```
:format-options="{
      style: 'unit',
      unit:'day',
      unitDisplay: 'long',
    }"
:locale="ar"
:min="0"
```

### Describe the bug

i'm using this option on number-field  it working well if locale is en or fr but when locale is ar  increment and decrement start the unexpected beheviour if value less than 3 ( that's what i get with trial)

zernonia:
Thanks for the issue @peter-emad99 . I believe it might be upstream issue with `@internationalized/number`. seeing the same issue with [`react-aria-components`](https://stackblitz-starters-bwt79x.stackblitz.io)

zernonia:
upstream issue https://github.com/adobe/react-spectrum/issues/6861

peter-emad99:
[...]

PR #1205, opened 2024-08-09
fix(NumberField): don't working well if locale is ar

kaishuige:
Fixes: #1204

My interim solutions is to handle locale = 'ar' , Better programmes can be proposed


zernonia:
Appreciate the PR @kaishuige , however I will hold-on with this workaround as I believe the upstream with `@internationalized/number` not parsing the value correctly.

zernonia:
Closing for now as verified it's upstream issue

PR #1208, opened 2024-08-09
feat(Stepper): add next and prev step slot function

epr3:
This PR adds next step and prev step functions to the `StepperRoot` slot props.

It also fixes a bug on the `goToStep` function which was trying to get an attribute on an out-of-bounds element.


sadeghbarati:
Here is the repro, https://stackblitz.com/edit/fz6rta?file=src%2FApp.vue

zernonia (approved):
LGTM!

Issue #1242, opened 2024-08-21
[Bug]: Pin Input crashes on Chrome 91


maxmaxme:
### Environment

```bash
Developement/Production OS: Windows 11 19043.1110
Node version: 16.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.3.3
Client OS: Windows 11 19043.1110
Browser: Chrome 91.0.4430.212
```

### Link to minimal reproduction

https://www.radix-vue.com/components/pin-input

### Steps to reproduce

1. Install Chrome 91 or use https://www.browserstack.com/
2. Open `Pin Input` component
3. Check console 

### Describe the bug

On the old version of the Chrome browser (91), the Pin Input component crashes here https://github.com/radix-vue/radix-vue/blob/main/packages/radix-vue/src/PinInput/PinInputInput.vue#L162

### Context & Screenshots (if applicable)

<img width="1579" alt="Screenshot 2024-08-21 at 12 16 09" src="https://github.com/user-attachments/assets/324d2f4b-49cd-4c5d-959f-37856a53c338">

maxmaxme:
https://caniuse.com/mdn-javascript_builtins_array_at

`.at` was added in the 92 version

PR #1243, opened 2024-08-21
fix(PinInputInput): replace `.at` with array index

maxmaxme:
Fix https://github.com/radix-vue/radix-vue/issues/1242


zernonia:
Thanks for the issue and fixes @maxmaxme ! 🎉

Issue #1246, opened 2024-08-24
[Bug]: Toast breaks SSR on Vercel 


rankjie:
### Environment

```bash
Developement/Production OS: Vercel
Node version: 20.X
Package manager: npm
Radix Vue version: 1.9.4
Vue version: 3.4.38
Nuxt version: 3.13.0
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.4.10
Client OS: Any
Browser: Any

I started using the Toast component from shadcn-vue last week, and since then, my build on Vercel intermittently encounters 500 errors. These errors are inconsistent: they may appear for every request during a certain period and then disappear entirely.

This issue does not occur in my development environment or when deploying to my own server. It only appears when deploying on Vercel. I'm guessing suggesting there might be some specific configuration or environment difference in Vercel's runtime causing this.

For the reasons above, I'm very sorry but I can not provide a minimal reproduction.
I tried but I can not recreate the error on any other platforms.

The specific error message is:

`ReferenceError: window is not defined
    at m (file:///var/task/node_modules/radix-vue/dist/index.js:17636:48)
    at te.immediate (file:///var/task/node_modules/radix-vue/dist/index.js:17661:65)
    at callWithErrorHandling (/var/task/node_modules/@vue/runtime-core/dist/runtime-core.cjs.prod.js:85:19)
    at callWithAsyncErrorHandling (/var/task/node_modules/@vue/runtime-core/dist/runtime-core.cjs.prod.js:92:17)
    at doWatch (/var/task/node_modules/@vue/runtime-core/dist/runtime-core.cjs.prod.js:4576:7)
    at watch (/var/task/node_modules/@vue/runtime-core/dist/runtime-core.cjs.prod.js:4497:10)
    at setup (file:///var/task/node_modules/radix-vue/dist/index.js:17660:9)
    at callWithErrorHandling (/var/task/node_modules/@vue/runtime-core/dist/runtime-core.cjs.prod.js:85:19)
    at setupStatefulComponent (/var/task/node_modules/@vue/runtime-core/dist/runtime-core.cjs.prod.js:6122:25)
    at setupComponent (/var/task/node_modules/@vue/runtime-core/dist/runtime-core.cjs.prod.js:6109:36)`

The error originates from radix-vue's ToastRootImpl inside the startTimer function. The reference to window seems to cause the issue. I guess it would be appropriate to add some code to check if it's SSR or CSR before using `window`.


Suggested Solution
[...]

PR #1247, opened 2024-08-25
fix(Toast): check if window exists to run startTimer in CSR only

rankjie:
fix issue #1246


packages/radix-vue/src/Toast/ToastRootImpl.vue
```diff
@@ -80,6 +80,11 @@ const remainingRaf = useRafFn(() => {
 function startTimer(duration: number) {
   if (!duration || duration === Number.POSITIVE_INFINITY)
     return
+  // startTimer is used inside a watch with immediate set to true.
+  // This results in code execution during SSR.
+  // Ensure this code only runs in a browser environment
+  if (typeof window === 'undefined')
+    return
```
zernonia, line 87: Let's use `isClient` instead 😁

zernonia:
Thanks for the issue and PR @rankjie !! Nicely done!

PR #1256, opened 2024-08-28
fix(DismissableLayer): branch could be null

rankjie:
Received error report in production via Sentry:

TypeError
Cannot read properties of null (reading 'contains')
at radix-vue/dist/index.js in m at line 2205:18
-> (S) => S.contains(g.target)


zernonia (approved):
Thanks for the fix @rankjie 🚀

PR #1271, opened 2024-09-05
Expose virtualizer and virtualized items

### LINKED ISSUE #1266: [Feature]: Possibility to calcul item's height dynamically for virtualized items (ListboxVirtualizer, TreeVirtualizer)
### Describe the feature

It would be valuable if ListboxVirtualizer (and TreeVirtualizer) could calculate the height of an element dynamically [https://tanstack.com/virtual/latest/docs/framework/vue/examples/dynamic](https://tanstack.com/virtual/latest/docs/framework/vue/examples/dynamic)

I suggest add two props :

`dynamicSize?: boolean`, if true, call vue hook `onVnodeMounted` :
```ts
const virtualizedItems = computed(() => virtualizer.value.getVirtualItems().map((item) => {
  return {
    item,
    is: cloneVNode(slots.default!({
      option: props.options[item.index],
    })![0], {
      'key': `${item.key}`,
      'data-index': item.index,
      'aria-setsize': props.options.length,
      'aria-posinset': item.index + 1,
      'style': {
        position: 'absolute',
        top: 0,
        left: 0,
        transform: `translateY(${item.start}px)`,
        overflowAnchor: 'none',
      },
      onVnodeMounted: props.dynamicSize ? ({el}) => virtualizer.value.measureElement(el) : undefined, // <===== HERE
    }),
  }
}))
```
and `measureElement?: (element: Element)=>void`, directly forwarded to useVirtualizer:
```ts
const virtualizer = useVirtualizer(
  {
    get scrollPaddingStart() { return padding.value.start },
    get scrollPaddingEnd() { return padding.value.end },
    get count() { return props.options.length },
    get horizontal() { return rootContext.orientation.value === 'horizontal' },
    estimateSize() {
      return props.estimateSize ?? 28
    },
    getScrollElement() { return parentEl.value },
    overscan: 12,
    measureElement: props.measureElement, // <===== HERE
  },
)
```

What do you think?

### Additional information

Dodobibi:
Expose the virtualizer to the default slot of ListboxVirtualizer and TreeVirtualizer.

This allows the height of the virtualized element to be dynamically calculated if necessary or styled.
```vue
<template>
 <ListboxVirtualizer
    v-slot="{ option, virtualizer, virtualItem }"
    ... ...
  >
    <ListboxItem
      :class="virtualItem.index % 2 === 0 && 'bg-accent/50'"
      @vue:mounted="(node: VNode) => virtualizer.measureElement(node.el as Element)"
      ... ...
    >
        <div>... ... ...</div>
    </ListboxItem>
  </ListboxVirtualizer>
</template>
```
resolve #1266

### DIFF
--- a/packages/radix-vue/src/Listbox/ListboxVirtualizer.vue
+++ b/packages/radix-vue/src/Listbox/ListboxVirtualizer.vue
@@ -21,12 +21,15 @@ import { getNextMatch } from '@/shared/useTypeahead'
 import { useParentElement } from '@vueuse/core'
 import { useCollection } from '@/Collection'
 import type { AcceptableValue } from '@/shared/types'
+import type { Virtualizer, VirtualItem } from '@tanstack/vue-virtual'
 
 const props = defineProps<ListboxVirtualizerProps<T>>()
 
 defineSlots<{
   default: (props: {
-    option: T
+    option: T;
+    virtualizer: Virtualizer<Element | Window, Element>;
+    virtualItem: VirtualItem
   }) => any
 }>()
 
@@ -71,6 +74,8 @@ const virtualizedItems = computed(() => virtualizer.value.getVirtualItems().map(
     item,
     is: cloneVNode(slots.default!({
       option: props.options[item.index],
+      virtualizer: virtualizer.value,
+      virtualItem: item
     })![0], {
       'key': `${item.key}`,
       'data-index': item.index,
--- a/packages/radix-vue/src/Listbox/ListboxVirtualizer.vue
+++ b/packages/radix-vue/src/Listbox/ListboxVirtualizer.vue
@@ -10,7 +10,7 @@ export interface ListboxVirtualizerProps<T extends AcceptableValue = AcceptableV
 </script>
 
 <script setup lang="ts" generic="T extends AcceptableValue = AcceptableValue">
-import { useVirtualizer } from '@tanstack/vue-virtual'
+import { type VirtualItem, type Virtualizer, useVirtualizer } from '@tanstack/vue-virtual'
 import { type Ref, cloneVNode, computed, useSlots } from 'vue'
 import { injectListboxRootContext } from './ListboxRoot.vue'
 import { compare, queryCheckedElement } from './utils'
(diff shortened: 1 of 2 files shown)


packages/radix-vue/src/Listbox/ListboxVirtualizer.vue
```diff
@@ -21,12 +21,15 @@ import { getNextMatch } from '@/shared/useTypeahead'
...
+import type { Virtualizer, VirtualItem } from '@tanstack/vue-virtual'
 
 const props = defineProps<ListboxVirtualizerProps<T>>()
 
 defineSlots<{
   default: (props: {
-    option: T
+    option: T;
+    virtualizer: Virtualizer<Element | Window, Element>;
+    virtualItem: VirtualItem
```
zernonia, line 32: seems like it's failing due to `VirtualItem` requires 1 type argument, which is `Element`. Let's add this argument ya

https://github.com/radix-vue/radix-vue/actions/runs/10722138349/job/29756726077#step:8:32

packages/radix-vue/src/Tree/TreeVirtualizer.vue
```diff
@@ -20,7 +21,9 @@ const props = defineProps<TreeVirtualizerProps>()
 
 defineSlots<{
   default: (props: {
-    item: FlattenedItem<Record<string, any>>
+    item: FlattenedItem<Record<string, any>>;
+    virtualizer: Virtualizer<Element | Window, Element>;
+    virtualItem: VirtualItem;
```
zernonia, line 26: similarly 😁

zernonia (changes requested):
Great work! Some quick suggestion

packages/radix-vue/src/Listbox/ListboxVirtualizer.vue
Dodobibi: fixed (with lint in the imports too...)

packages/radix-vue/src/Tree/TreeVirtualizer.vue
Dodobibi: fixed too 😁

zernonia (approved):
Appreciate the quick changes and PR @Dodobibi ! 🚀

PR #1303, opened 2024-09-19
Update: Fixed WAI Aria Link in popover.md

stripedpurple:
The WAI Aria documentation link points to a non existent page, so I have updated it to the correct one.


zernonia:
Thanks for the PR @stripedpurple ! 🚀

Issue #1305, opened 2024-09-19
[Bug]: error: Assertion failed when hovering link in popover overlapping SplitterResizeHandle


pedromelo222:
### Environment

```bash
Developement/Production OS: Windows 10
Node version: v18.20.3
Package manager: pnpm@8.15.6
Radix Vue version: 1.9.5
Vue version: 3.4.0
CSS framework: tailwindcss@3.3.3
Client OS: Windows 10 19043.1110
Browser: Chrome 128.0.6613.137 (Official Build) (64-bit)
```

### Link to minimal reproduction

https://stackblitz.com/edit/ztc1e6?file=src%2FApp.vue

### Steps to reproduce

https://github.com/user-attachments/assets/0599819c-4e14-4c83-bbbe-3f31ee9c76e1

1. Add the radix-vue Splitter component to a Vue application.
2. Open a popover that contains a link (<a> element with div inside) so that it overlaps the SplitterResizeHandle.
3. Hover over the link in the popover.
4. Observe that the error "radix-vue: Assertion failed!" is triggered.

### Describe the bug

Uncaught Error: Assertion failed!

### Expected behavior

Hovering over a link in a popover that overlaps the SplitterResizeHandle should not trigger an assertion failure.

PR #1307, opened 2024-09-20
fix(Splitter): fix getParent()

sunnylost:
Fixes https://github.com/radix-vue/radix-vue/issues/1305.

`a` and `area` element can have `host` property.


zernonia (approved):
Yo @sunnylost ! Thanks for the fix! 💚

PR #1309, opened 2024-09-20
fix(Combobox): infer modelValue emit based on multiple

iamandrewluca:


sunnylost:
Turns out we can infer the value of `v-model` without using `multiple`.

```ts
// This is incorrect, T can never be an array
export type ComboboxRootEmits<T = AcceptableValue> = {
  'update:modelValue': [value: T]
```

So we should update `T` to make it can be an array.

```ts
type ModelValue = AcceptableValue | Array<AcceptableValue>

// change context
export const [injectComboboxRootContext, provideComboboxRootContext]
  = createContext<ComboboxRootContext<ModelValue>>('ComboboxRoot')

// change emits
export type ComboboxRootEmits<T = ModelValue> = {
  'update:modelValue': [value: T]
  // ...
}

// also change props
export interface ComboboxRootProps<T = ModelValue> extends PrimitiveProps {
  modelValue?: T
  defaultValue?: T
  // ...
```

Finally, we should update Vue's generic:

```html
<script setup lang="ts" generic="T extends ModelValue = AcceptableValue">
```

zernonia:
Thanks for the PR @iamandrewluca ! The solution suggested by @sunnylost  is make sense and straightforward. Can you give it a try? 😁

iamandrewluca:
@zernonia I started looking into it locally, but then I thought, can it be the user does not want to use `multiple` but the `modelValue` is an `Array`? 🤔

iamandrewluca:
It seems to be a little complicated. TS will infer the type based on different usage, and we might need to use [`NoInfer`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html#the-noinfer-utility-type), which might be a breaking change 🤔

iamandrewluca:
I will close this for now. If anyone is interested in continuing it, feel free to do so. Thanks

PR #1318, opened 2024-09-28
Allow default input behavior for text selection on TagsInput and Combobox

michgeek:
When typing inside the input of a `combobox` or `tagsinput` it should be allowed to select that text with `shift` + either  `left`, `right`, `home` or `end`.

The current behavior is preventing the combination of `shift+end` and `shift+home`.

### DIFF
--- a/packages/radix-vue/src/DateField/DateFieldRoot.vue
+++ b/packages/radix-vue/src/DateField/DateFieldRoot.vue
@@ -192,7 +192,7 @@ watch(locale, (value) => {
 })
 
 watch(modelValue, (_modelValue) => {
-  if (!isNullish(_modelValue) || placeholder.value.compare(_modelValue) !== 0) {
+  if (!isNullish(_modelValue) && placeholder.value.compare(_modelValue) !== 0) {
     placeholder.value = _modelValue.copy()
   }
 })
--- a/packages/radix-vue/src/DatePicker/DatePicker.test.ts
+++ b/packages/radix-vue/src/DatePicker/DatePicker.test.ts
@@ -236,4 +236,17 @@ describe('datePicker', async () => {
     expect(month).not.toHaveAttribute('tabindex')
     expect(year).not.toHaveAttribute('tabindex')
   })
+
+  it('should select and deselect a date', async () => {
+    const { user, trigger, getByTestId } = setup()
+
+    await user.click(trigger)
+    const calendar = getByTestId('calendar')
+    const targetCell = calendar.querySelector('div[data-radix-vue-calendar-cell-trigger]:not([data-outside-view])')!
+
+    await user.click(targetCell)
+    expect(calendar.querySelector('[data-selected]')).toBeInTheDocument()
+    await user.click(targetCell)
+    expect(calendar.querySelector('[data-selected]')).not.toBeInTheDocument()
+  })
 })


packages/radix-vue/src/TagsInput/TagsInputRoot.vue
```diff
@@ -172,6 +172,10 @@ provideTagsInputRootContext({
       case 'End':
       case 'ArrowRight':
       case 'ArrowLeft': {
+        // allow text selection while maintaining `shift` key
+        if(event.shiftKey)
+          break
+
```
zernonia, line 178: I think we don't need it for `Tags Input`. The current implementation works well, where we only focus on `TagsInputItem` when we are not highlighting any text

https://github.com/user-attachments/assets/da69a838-8aaa-4df6-936f-4f652e7a3c97

michgeek: I believe you used arrows to select char one by one. This one enables text selection in its entirety or what's left of it with shift+home or shift+end. That is the expected behaviour when using a text input.

michgeek: [Screencast from 2024-10-02 20-53-08.webm](https://github.com/user-attachments/assets/61e45082-5918-45c6-b261-1f9a6435edbb)

michgeek: What about the following behavior ? I removed the tests I wrote because I think they're not great and couldn't figure it out.

[Screencast from 2024-10-02 21-52-00.webm](https://github.com/user-attachments/assets/82e9a914-8d64-4533-8664-011ea0b15dca)

[Screencast from 2024-10-02 21-55-07.webm](https://github.com/user-attachments/assets/528bdaf0-b37f-4c6d-b3c5-5523d53a3148)

packages/radix-vue/src/TagsInput/TagsInputRoot.vue
```diff
@@ -174,8 +174,10 @@ provideTagsInputRootContext({
       case 'ArrowLeft': {
         const isArrowRight = (event.key === 'ArrowRight' && dir.value === 'ltr') || (event.key === 'ArrowLeft' && dir.value === 'rtl')
         const isArrowLeft = !isArrowRight
-        // only focus on tags when cursor is at the first position
-        if (target.selectionStart !== 0 || target.selectionEnd !== 0)
+
+        // prevent tag focus when there's something in the input
+        // the user must clear the input if he needs to clear already added tags
+        if (target.value.length > 0)
```
zernonia, line 180: This is not the behavior we want. We want to allow user to focus on tags even when there's input value.

michgeek: I understand that's the behavior you intended. However, it's hijacking the default behavior of a text input. Its purpose is to allow a user to enter a new value or perform a search, therefore it is reasonable to expect it to behave like a normal text input.

From a user's perspective, typing indicates the intention of adding or searching. 
Therefore while he is typing, he should be able to select it as he normally would. That intention should be prioritized over the intention of focusing tags that were already added.

I believe you should reconsider this behavior. Otherwise feel free to close the PR.

Many thanks

PR #1321, opened 2024-09-29
fix(ConfigProvider): pass string to `ScrollBodyOption`

### LINKED ISSUE #1319: [Feature]: Allow string for `padding` and `margin` in `ScrollBodyOption`
### Describe the feature

Hello,

Actually, it's only possible to pass a number which is converted to `px`. But this is annoying when everything in the page use `rem` because you have to convert value. Couldn't be possible to pass a string like `0.6rem` instead of `9.6`?

### Additional information

hrynevychroman:
This PR fix #1319 

https://github.com/user-attachments/assets/1d0b8fad-25f7-4037-bce3-2cc4a9337ca8

### DIFF
--- a/packages/radix-vue/src/shared/types.ts
+++ b/packages/radix-vue/src/shared/types.ts
@@ -38,9 +38,14 @@ interface SingleOrMultipleProps<ValidValue = string | string[], ExplicitType = S
 
 }
 
+/**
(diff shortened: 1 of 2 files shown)


zernonia (approved):
LGTM!! Thanks @hrynevychroman ! 🎉

PR #1325, opened 2024-09-30
fix(Editable): `maxLength` not working

iammola:


packages/radix-vue/src/Editable/EditableInput.vue
```diff
@@ -56,6 +56,7 @@ function handleSubmitKeyDown(event: KeyboardEvent) {
     :value="context.modelValue.value"
     :placeholder="placeholder"
     :disabled="disabled"
+    :maxLength="context.maxLength.value"
```
zernonia, line 59:
```suggestion
    :maxlength="context.maxLength.value"
```

zernonia (approved):
Thanks for the PR and changes @iammola  😁

PR #1328, opened 2024-10-01
fix(Combobox): defer highlighted item scrollIntoView

### LINKED ISSUE #1327: [Bug]: Combobox: highlighted element not scrolled into view when using v-model:selectedValue
### Environment

```bash
Node version: 16.0.0
Package manager: pnpm@8.6.0
Radix Vue version: 1.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
Browser: Chrome 90.0.4430.212
```

https://stackblitz.com/edit/gbaaqa?file=src%2FApp.vue

### Steps to reproduce

1. Assign a max height to the `ComboboxContent` component
2. Provide a controlled `selectedValue`, ie: `model:selectedValue="selectedValue"`
3. Start navigating the combobox results with down arrow keys

### Describe the bug

When approaching the bottom of the combobox contents, and pressing down key, the highlighted element is no longer scrolled into view. The item is highlighted, but it's hidden "below the fold". If I manually scroll the list with my mouse, I can see the highlighted element.

### Expected behavior

The highlighted element should always be scrolled into view.

Not sure what the root cause is, but it looks like deferring the [`scrollIntoView` call here](https://github.com/radix-vue/radix-vue/blob/18e5468a74ea0f7f33dfae6783c77f572365378d/packages/radix-vue/src/Combobox/ComboboxRoot.vue#L235-L236) using `nextTick` fixes the issue:

```js
  if (selectedElement.value instanceof Element)
    nextTick(() => selectedElement.value.scrollIntoView({ block: 'nearest' }))
```

CleanShot 2024-10-01 at 09 24 55

ragulka:
Fixes #1327 

Also added a story to showcase and test highlighting behavior in a scrolling combobox.

### DIFF
--- a/packages/radix-vue/src/Combobox/ComboboxRoot.vue
+++ b/packages/radix-vue/src/Combobox/ComboboxRoot.vue
@@ -137,6 +137,9 @@ async function onOpenChange(val: boolean) {
       else
         selectedValue.value = modelValue.value as T
     }
+    // selectedElement is a computed value and is not yet fully resolved.
+    // We need to wait for it to finish processing at this point.
+    await nextTick()
     inputElement.value?.focus()
     scrollSelectedValueIntoView()
   }
@@ -233,7 +236,7 @@ function scrollSelectedValueIntoView() {
   // Find the highlighted element and scroll into view
   // We can put this in Item, but we avoid having too many watcher
   if (selectedElement.value instanceof Element)
-    nextTick(() => selectedElement.value?.scrollIntoView({ block: 'nearest' }))
+    selectedElement.value?.scrollIntoView({ block: 'nearest' })
 }
 
 function focusOnSelectedElement() {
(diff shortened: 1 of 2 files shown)


sunnylost:
I believe this is more accurate:

```ts
// ComboboxRoot.vue
selectedValue.value = ...
// selectedElement is a computed value and is not yet fully resolved. 
// We need to wait for it to finish processing at this point.
await nextTick()

scrollSelectedValueIntoView()
focusOnSelectedElement()
```

ragulka:
@sunnylost you're probably right - should I update my PR?

sunnylost:
Yes, for the same reason, `focusOnSelectedElement()` also received the wrong `selectedElement`.

ragulka:
@sunnylost got it, I've refactored my PR based on your suggestion. I also have to add another `await nextTick()` when the `open` state changes, to ensure that if the selected value is "below the fold", then it will immediately be scrolled into view when opening the options.

zernonia (approved):
Thanks for the fix @ragulka  @sunnylost ! 🚀

PR #1339, opened 2024-10-04
fix(Tooltip): use trigger's `ownerDocument` to track `pointermove`

iammola:
In Popup windows, the tooltip content wasn't closing after moving from the area, only after going back to the original window and moving the pointer there.

https://github.com/user-attachments/assets/45f94c06-1b27-469c-9b95-b21d7fd9075a


zernonia (approved):
Thanks for the fix @iammola ! 💚

iammola:
Hey @zernonia, is there a timeframe on the next release date? Or is that delayed until the v2 release?

PR #1340, opened 2024-10-05
fix: set inheritAttrs to false for components that render fragments

sunnylost:
Fixes https://github.com/unovue/radix-vue/issues/1320

### DIFF
--- a/packages/radix-vue/src/ConfigProvider/ConfigProvider.vue
+++ b/packages/radix-vue/src/ConfigProvider/ConfigProvider.vue
@@ -39,6 +39,10 @@ export interface ConfigProviderProps {
 <script setup lang="ts">
 import { toRefs } from 'vue'
 
+defineOptions({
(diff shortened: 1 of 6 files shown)


sadeghbarati:
I had this problem in `shadcn-vue` too 

- https://github.com/unovue/shadcn-vue/pull/773/files#diff-b6d068c88f37d002b2c45771d25c3f4a80de8220de3fae296f694637f472c085
- https://github.com/unovue/shadcn-vue/blob/dev/apps/www/src/lib/registry/default/example/DatePickerWithRange.vue#L27

```diff
- <DatePickerWithRange class="[&>button]:w-[260px]" />
+ <DatePickerWithRange />
```

Related issue

- https://github.com/unovue/shadcn-vue/issues/793

zernonia (approved):
Thanks for the PR @sunnylost ! 💚

PR #1368, opened 2024-10-13
fix(combobox/item): fix some items not add into allItems map

### LINKED ISSUE #1234: [Bug/v2]: ComboboxItem unhandled error in mounted
### Environment

```bash
Developement/Production OS: Windows 11
Node version: 20
Package manager: pnpm@8.6.0
Reka Vue version: 0.0.2
Vue version: ^3.0.0
Client OS: Windows 11
Browser: Firefox
```

https://github.com/RayGuo-ergou/reka-reproduction

### Steps to reproduce

Clone the repo, run the dev server, then in app search anything, delete all text then search again that will make the `computed` as empty array.

```ts
const people = [
  { id: 1, name: 'Durward Reynolds'},
  { id: 2, name: 'Kenton Towne'},
  { id: 3, name: 'Therese Wunsch'},
  { id: 4, name: 'Benedict Kessler'},
  { id: 5, name: 'Katelyn Rohan'},
]
const selectedPeople = ref(people[0])
const search = ref('')

const searchResult = computed(()=>{
  return people.filter(i => !i.name.startsWith(search.value))
})
```

<details><summary>Warning</summary>
<p>

[Vue warn]: Unhandled error during execution of mounted hook 
  at <ComboboxItem key=1 value= 
Object { id: 1, name: "Durward Reynolds" }
 > 
....
</p>
</details> 

<details><summary>Error</summary>
<p>

Uncaught (in promise) TypeError: can't access property "textContent", u.value is undefined

</p>
</details> 

### Describe the bug

I was trying v2 aka reka with shadcn-ui, and I found that a breaking changing in combobox may break the [tags-input](https://www.shadcn-vue.com/docs/components/tags-input.html)

In https://github.com/radix-vue/radix-vue/commit/ee8a3f2366a5c27c2bf1cc0a1ecbb0fea559a9f7#diff-19766b3de8934964a0c2730b5e4d28c90759d1b67572135ba3be029752b4d31dR50

The `currentElement` will be undefined if the list of items become an empty array.

### Expected behavior

_No response_

GIF 19-08-2024 12-35-05 PM

RayGuo-ergou:
### Description

closes #1234

This pr fixes when `rootContext.isVirtual.value || rootContext.ignoreFilter.value || !rootContext.filterState.search` is false, the item will never add into `allItems` map.

To reproduce this bug, create a async combobox (e.g. the one in the story), before opening the dropdown, give any value to search query.

<hr />

Added an event listener to apply filter when `allItems` size changes.

\* It uses `watch` rather than `watchDebounced` even the items are set one by one because it's handled by [vue `watch`](https://vuejs.org/guide/essentials/watchers.html#callback-flush-timing)

### DIFF
--- a/packages/core/src/Combobox/ComboboxRoot.vue
+++ b/packages/core/src/Combobox/ComboboxRoot.vue
@@ -60,7 +60,7 @@ export interface ComboboxRootProps<T = AcceptableValue> extends Omit<ListboxRoot
 <script setup lang="ts" generic="T extends AcceptableValue = AcceptableValue">
 import { computed, nextTick, reactive, ref, toRefs, watch } from 'vue'
 import { PopperRoot } from '@/Popper'
-import { type EventHookOn, createEventHook, useVModel, watchDebounced } from '@vueuse/core'
+import { type EventHookOn, createEventHook, useVModel } from '@vueuse/core'
 import { ListboxRoot } from '@/Listbox'
 
 const props = withDefaults(defineProps<ComboboxRootProps<T>>(), {
@@ -181,9 +181,9 @@ watch(() => open.value, () => {
   })
 }, { flush: 'post' })
 
-watchDebounced(() => allItems.value.size, (val) => {
+watch(() => allItems.value.size, (val) => {
   filterItems()
-}, { debounce: 10 })
+})
 
 defineExpose({
   highlightedElement,
--- a/packages/core/src/Combobox/ComboboxRoot.vue
+++ b/packages/core/src/Combobox/ComboboxRoot.vue
@@ -169,7 +169,7 @@ function filterItems() {
   filterState.filtered.count = itemCount
 }
 
-watch(() => filterState.search, () => {
+watch([() => filterState.search, () => allItems.value.size], () => {
   filterItems()
 }, { immediate: true })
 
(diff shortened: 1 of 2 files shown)


RayGuo-ergou:
Hi @zernonia, just saw the new releases 😀 which reminds me of this PR, could you review this when you have time? Thanks.

packages/core/src/Combobox/ComboboxRoot.vue
```diff
@@ -181,6 +181,10 @@ watch(() => open.value, () => {
   })
 }, { flush: 'post' })
 
+watch(() => allItems.value.size, (val) => {
```
ineshbose, line 184: I might have spotted this bug as well when a ComboboxItem is mounted mid-search.

Don't you think that we can update L172 as

```ts
watch(() => [filterState.search, allItems.value.size], () => {
```

ineshbose: OK I think I found another issue in ComboboxItem a bit related to this - the `props.textValue || currentElement.value.textContent || currentElement.value.innerText` can update for a node while `id` remains same, and in that case it gets outdated.

RayGuo-ergou: Hi, yes you are right. Thanks for spotting that.

PR #1373, opened 2024-10-14
fix(docs): typo in Stepper component

sebsobseb:
Not sure if it was intended, but the term 'tab' is used instead of 'step' in the Stepper component.


zernonia (approved):
Ahh nice catch! Thanks @sebsobseb !

PR #1408, opened 2024-10-30
feat(TagsInput): support multiple delimiters via RegExp

### LINKED ISSUE #1400: [Feature]: Tags Input. Support multiple delimiters
### Describe the feature

Improve `TagsInputInput` to allow multiple delimiters so that multiple formats of pasted data can be supported.
 
For example I use `TagsInput` component to collect and submit list of email. Occasionally this list will be pasted from other sources for example it may be comma-separated list or tabular data like google sheets.

**Data example**
```
jane.doe@example.com, john.smith@example.com, alex.jones@example.com, taylor.lee@example.com, casey.miller@example.com
```

**Proposed usage**
```tsx
const delimiters = ["\t", "\n\r", "\n", " ", ","]

<TagsInputInput :delimiter="delimiters" />
```

Additionally it would need to strip out empty entries after string split for `modelValue` to be accurate.

Originally for my use-case I was thinking to avoid default `addOnPaste` and implement custom `@paste` handler, but wondering if there's interest to have this functionality built in.

### Additional information

romansp:
Resolves #1400.

Added support for `RegExp` delimiter prop in `TagsInput`.

Open questions:
1. Should we do our own validation of passed RegExp to ensure `global` flag is set. Without `global` flag `String.prototype.replaceAll` will produce the following error: `Uncaught TypeError: String.prototype.replaceAll called with a non-global RegExp argument`
2. Consider support for `string[]` delimiter. If we do then decide:
     1. Convert `string[]` into `RegExp`. Seems somewhat tricky to do it correctly and account for all the edge cases, e.g. escaping special characters.
     2. Iterate delimiters and replace/split for each delimiter. Quite straightforward but will require slightly bigger rework in `handleInput` and `handlePaste`.

Do you have any thoughts on these @zernonia? Or maybe other suggestions?


packages/radix-vue/src/TagsInput/TagsInputInput.vue
```diff
@@ -64,8 +64,12 @@ async function handleCustomKeydown(event: Event) {
...
   context.isInvalidInput.value = false
+  if (event.data === null)
+    return
+
   const delimiter = context.delimiter.value
-  if (delimiter === event.data) {
+  const matchesDelimiter = delimiter === event.data || (delimiter instanceof RegExp && delimiter.test(event.data))
+  if (matchesDelimiter) {
     const target = event.target as HTMLInputElement
     target.value = target.value.replaceAll(delimiter, '')
```
zernonia, line 74: I think we can change it to `replace`, it shouldn't really matter as we handle paste event first then input event.

```suggestion
    target.value = target.value.replace(delimiter, '')
```

zernonia:
We also need to apply the changes to [handlePaste](https://github.com/unovue/radix-vue/pull/1408/files#diff-a854e14885c2868a7b936d3507a32abbd2ca1db5b4fe8296a01db1676e90288dR91-R94)

romansp:
@zernonia thanks for the review.

> We also need to apply the changes to [handlePaste](https://github.com/unovue/radix-vue/pull/1408/files#diff-a854e14885c2868a7b936d3507a32abbd2ca1db5b4fe8296a01db1676e90288dR91-R94)
> 
> 
I don't think any additional changes are needed for `handlePaste`. `String.prototype.split` accepts RegExp argument. I tested it with `/[ ,;]+/` and it worked as expected on paste. 

Would you like me to add an example to docs showcasing multiple delimiters usage?

romansp:
PR updated:
1. switch to `replace` instead of `replaceAll` to not require `global` flag on regexp
2. updated `delimiter` prop description
3. added example for multiple delimiters usage with regexp

zernonia:
Nicely done @romansp !! Working well! 🚀
We have freeze any new feature to `main` branch, perhaps we can push this to `v2`? 😁

Also, do you mind quickly adding in some test case for this?

romansp:
@zernonia Thanks! I'm good with this being pushed to `v2`. Do you want me to change target branch or you do it?

Let me check what type of test case we can add here. We could probably verify that typing text with delimiters creates tags correctly? I will try to add coverage for paste event as well.

zernonia:
Super welcome @romansp !! Once again thanks for the PR!
Sorry I messed up the branch, I've cherry-picked your commit to https://github.com/unovue/radix-vue/pull/1414 😁

PR #1415, opened 2024-11-05
fix(Pagination): ignore clicks if pagination button is disabled

andreww2012:
Pagination buttons (like prev/next) can become any HTML element, for example `<a>`, which, unlike `<button>`, doesn't support `disabled` attribute and still emits `click` events.

If a pagination button component is considered disabled, but rendered as a non-button HTML element, clicking it may update the component's state in a bad way (for example, set the current page to 0 after clicking the "<" button when being on the first page).

This PR adds `!disabled &&` conditions to all pagination buttons components' `click` event handlers.

### DIFF
--- a/packages/radix-vue/src/Pagination/Pagination.test.ts
+++ b/packages/radix-vue/src/Pagination/Pagination.test.ts
@@ -56,6 +56,74 @@ describe('given default Pagination', () => {
   })
 })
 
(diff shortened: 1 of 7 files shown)


zernonia (approved):
Nicely done @andreww2012 !! 

Thanks for adding the test case too! 🚀

PR #1416, opened 2024-11-05
Add onTagRemove emit in TagsInputRoot.vue

jochemvogel:
It’s currently challenging to trigger an action when an element is removed. At the moment, we’re watching modelValue, and when the new array is shorter than the previous one, we identify the missing element and use it as a payload to update our database.

To streamline this, I propose adding an onTagRemove event that emits the removed element as a payload. This will allow us to directly handle the tag removal in our database, making the process more efficient and reliable.

Let me know if you have any questions. Thank you.


zernonia:
Thanks for the PR @jochemvogel ! Good suggestion! However we have freeze feature deployment to `main`. I will cherry-pick your solution to [v2](https://github.com/unovue/radix-vue/tree/v2) 😁

PR #1422, opened 2024-11-14
[V2] Fix CheckboxRoot update:modelValue typing

aguingand:
Typing of Checkbox `emit('update:modelValue')` should be `boolean | 'indeterminate'` instead of `AcceptableValue`.

### DIFF
--- a/packages/core/src/Checkbox/CheckboxRoot.vue
+++ b/packages/core/src/Checkbox/CheckboxRoot.vue
@@ -25,7 +25,7 @@ export interface CheckboxRootProps extends PrimitiveProps, FormFieldProps {
 
 export type CheckboxRootEmits = {
   /** Event handler called when the value of the checkbox changes. */
-  'update:modelValue': [value: AcceptableValue ]
+  'update:modelValue': [value: boolean | 'indeterminate']
 }
 
 interface CheckboxRootContext {
--- a/packages/core/src/Checkbox/CheckboxRoot.vue
+++ b/packages/core/src/Checkbox/CheckboxRoot.vue
@@ -7,13 +7,11 @@ import type { AcceptableValue, FormFieldProps } from '@/shared/types'
 import { useVModel } from '@vueuse/core'
 import { injectCheckboxGroupRootContext } from './CheckboxGroupRoot.vue'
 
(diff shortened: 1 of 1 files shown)


packages/core/src/Checkbox/CheckboxRoot.vue
```diff
@@ -25,7 +25,7 @@ export interface CheckboxRootProps extends PrimitiveProps, FormFieldProps {
 
 export type CheckboxRootEmits = {
   /** Event handler called when the value of the checkbox changes. */
-  'update:modelValue': [value: AcceptableValue ]
+  'update:modelValue': [value: boolean | 'indeterminate']
```
zernonia, line 28: Checkbox now accept other types, not only `boolean` in v2.

```suggestion
  'update:modelValue': [value: AcceptableValue | 'indeterminate']
```

aguingand: TS in my project is failing because AcceptableValue doesn't include `boolean`. I have copied the types of the prop : https://github.com/unovue/radix-vue/blob/76923e561b89e648ca7218f1195ff92031996d30/packages/core/src/Checkbox/CheckboxRoot.vue#L14
but I may be missing something

aguingand: There is the boolean casting quirk commented in AcceptableValue : https://github.com/unovue/radix-vue/blob/fc3927037b591a4869ea3b2414bdaff1af2f6d9f/packages/core/src/shared/types.ts#L51-L53

zernonia: Ahh you are right. Then let's make it ` [value: AcceptableValue | boolean | 'indeterminate']`

aguingand: I have changed to  `AcceptableValue | boolean | 'indeterminate'` and added generic.

zernonia:
Hey @aguingand ! My bad you were actually correct in the first commit, where the type should be just `boolean | 'indeterminate'`. I remember wrongly regarding the changes mada to support for custom types and array (which has been moved to `CheckboxGroupRoot`.

PR #1424, opened 2024-11-14
fix(Dialog): iOS safari scroll lock not working (#1423)

### LINKED ISSUE #1423: [Bug]: Safari iOS Dialog BodyScrollLock not working
### Environment

```bash
Node version: 20.15.0
Package manager: pnpm@3.15.0
Radix Vue version: 1.9.9
Vue version: 3.5.12
Nuxt version: 3.14.159
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.4.14
Client OS: MacOS Version 14.6.1
Browser: Chrome 130.0.6723.117
```

https://www.radix-vue.com/components/dialog.html#nested-dialog

### Steps to reproduce

On iOS Safari:
1. scroll down, such that the native safari footer collapses
2. open dialog
3. scroll outside the dialog

### Describe the bug

This bug allows one to still scroll the body, when a `Dialog` is opened, on mobile safari.

### Expected behavior

Should be able to scroll the body.

This bug only works when the native safari footer is collapsed (e.g. when previously scrolled down)
Otherwise it behaves as expected 🤷 
image
Then it'll allow you to continue to scroll, with the dialog being open.

https://github.com/user-attachments/assets/3d579dca-e46e-405e-84c6-706e01331cc0

94726:
fixes #1423

By adding `touch-action: none`, Safari seems to stop misbehaving. 
There might be some more edgecases hiding elsewhere as you can't ever be to sure with Safari, but this should do. 
At least for now 🙂

### DIFF
--- a/packages/radix-vue/src/shared/useBodyScrollLock.ts
+++ b/packages/radix-vue/src/shared/useBodyScrollLock.ts
@@ -31,6 +31,7 @@ const useBodyLockStackCount = createSharedComposable(() => {
     document.body.style.paddingRight = ''
     document.body.style.marginRight = ''
     document.body.style.pointerEvents = ''
+    document.body.style.touchAction = ''
     document.body.style.removeProperty('--scrollbar-width')
     document.body.style.overflow = initialOverflow.value ?? ''
     isIOS && stopTouchMoveListener?.()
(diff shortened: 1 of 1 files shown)


zernonia:
We don't want to prevent touch action, user might want to perform gesture event (zoom etc). I will have a look at this issue @94726 . Thanks for reporting and attempting the fix ya 😁

PR #1428, opened 2024-11-16
fix: bodyScrollLock not working on safari

zernonia:
Adapt the fix from Vueuse


RuneClaeys:
@zernonia 

I'm not 100% sure, but this commit seems to break the usage of the Apple Pencil within dialogs, popovers, alert dialogs, etc.
I'm relatively new to Radix-Vue (using it through Shadcn-Vue).

For now, I've reverted my Radix-Vue dependency to the version before this commit.

Could this issue be investigated? My current user base primarily consists of iPad users with Apple Pencil (I know this is a very specific case), but others might encounter similar issues.

PR #1429, opened 2024-11-16
fix(Toast): negative duration

sandros94:
This is a follow-up from https://github.com/nuxt/ui/issues/2665

Currently the Toast's duration only starts the timer if false or infinite.

But having a negative duration should also trigger a Toast without a timer.

### DIFF
--- a/packages/radix-vue/src/Toast/ToastRootImpl.vue
+++ b/packages/radix-vue/src/Toast/ToastRootImpl.vue
@@ -66,7 +66,11 @@ const { forwardRef, currentElement } = useForwardExpose()
 const providerContext = injectToastProviderContext()
 const pointerStartRef = ref<{ x: number, y: number } | null>(null)
 const swipeDeltaRef = ref<{ x: number, y: number } | null>(null)
-const duration = computed(() => props.duration || providerContext.duration.value)
+const duration = computed(
+  () => typeof props.duration === 'number'
+    ? props.duration
+    : providerContext.duration.value,
+)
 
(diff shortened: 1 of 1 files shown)


zernonia:
Hi @sandros94 ! Thanks for the PR. I think your proposal make sense 👍🏻

packages/radix-vue/src/Toast/ToastRootImpl.vue
```diff
@@ -79,7 +79,7 @@ const remainingRaf = useRafFn(() => {
 }, { fpsLimit: 60 })
 
 function startTimer(duration: number) {
-  if (!duration || duration === Number.POSITIVE_INFINITY)
+  if (duration <= 0 || duration === Number.POSITIVE_INFINITY)
```
zernonia, line 82:
```suggestion
  if (!duration || duration <= 0 || duration === Number.POSITIVE_INFINITY)
```

In case the duration value is `undefined`

zernonia:
In order to support `:duration="0"`, we need to adjust this to nullish coalescing operator as well

`const duration = computed(() => props.duration && providerContext.duration.value)`

https://github.com/unovue/radix-vue/pull/1429/files#diff-6607e968a50b4047c4bce25917a099b696a722e27d44e742a9ad7572a96e0b72R69-R70

sandros94:
> we need to adjust this to nullish coalescing operator as well

instead of a nullish coalescing operator (that would return `undefined` if `props.duration` isn't set) I decided to explicitly check if the type is a number, otherwise use the provider's value/default

PR #1431, opened 2024-11-17
fix(Combobox): onInputEnter handler should not be triggered when isComposing true

### LINKED ISSUE #1430: [Bug]: (ComboboxInput) onInputEnter handler should not be triggered when isComposing true
### Environment

```bash
Package manager: pnpm@9.12.1
Radix Vue version: 1.9.9
Vue version: 3.5.13
Nuxt version: 3.14.159
```

https://codesandbox.io/p/devbox/github/shoma-mano/radix-vue-reproduction

### Steps to reproduce

1. input ```山手``` manually.
2. press enter

### Describe the bug

In languages that have a conversion concept, conversion confirmation is performed with the Enter key. However, keydown.enter also responds to conversion confirmations, which causes onInputEnter to react to an Enter press intended for conversion as well, leading to unexpected behavior.
I will plan to create PR for this issue.

https://github.com/user-attachments/assets/b7dff216-b215-4313-86ac-1f18871ca0d2

### Expected behavior

_No response_

_No response_

shoma-mano:
This PR will fix #1430

I confirmed this PR will not affect English.

https://github.com/user-attachments/assets/2d0c7da6-6433-4167-a706-74789ea59db8

and will fix problem that is specific to languages that has conversion concept.

https://github.com/user-attachments/assets/19b8c48a-4d2c-446b-910a-f97762ebf6f1

I might be overlooking points that I haven't fully considered. If there's anything that stands out or seems concerning, please let me know.

### DIFF
--- a/docs/components/demo/Combobox/tailwind/index.vue
+++ b/docs/components/demo/Combobox/tailwind/index.vue
@@ -4,7 +4,7 @@ import { ComboboxAnchor, ComboboxContent, ComboboxEmpty, ComboboxGroup, Combobox
 import { Icon } from '@iconify/vue'
 
 const v = ref('')
-const options = ['Apple', 'Banana', 'Blueberry', 'Grapes', 'Pineapple', '林檎']
+const options = ['Apple', 'Banana', 'Blueberry', 'Grapes', 'Pineapple', 'りんご']
 const vegetables = ['Aubergine', 'Broccoli', 'Carrot', 'Courgette', 'Leek']
 </script>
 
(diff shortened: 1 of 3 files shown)


shoma-mano:
I just got to know keypress is deprecated. I will investigate if I can use isComposing to fix this.

zernonia:
Thanks for the issue and PR @shoma-mano . Correct as you pointed out that [@keypress](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event) has been deprecated, thus we are not using it in the package

shoma-mano:
@zernonia 
Sure!

shoma-mano:
I read some articles about composition events and learned that the reason Headless UI delays compositionend is because, in Safari, the keydown event occurs after oncompositionend. I’ll update my PR by mimicking the approach used in Headless UI.

https://qiita.com/alt_yamamoto/items/8663d047a3794dd5605e
image

Issue #908, opened 2024-05-11
v2


zernonia:
# Announcement: Changes Coming to Radix Vue

## Introduction
First and foremost, we really want to express appreciation for all the effort put in by the [Radix UI](https://github.com/radix-ui) team in creating the absolutely powerful, unstyled, high-quality, and accessible components that this project was originally ported from.

## Background
However, as feature requests grew, Radix Vue began to add more components and features, and modify certain behaviors specifically tailored for Vue. As a result, Radix Vue started to feel like a unique component library rather than just the ported version of Radix UI. Therefore, we are planning to give it a different identity.

## What's Coming in v2?
- Brand change
- NPM registry change
- Some component refactoring [link to project](https://github.com/orgs/unovue/projects/3) https://github.com/radix-vue/radix-vue/labels/v2

## Note
1. We will still be pushing bug fixes and introducing some new primitive components prior to the v2 release.
2. We are still brainstorming the new identity name. Please share any ideas you have in the comments below.
3. We are open to suggestions for areas of improvement. Please ensure that any suggestions are sensible and would result in a general improvement rather than addressing only a specific edge-case.

> [!IMPORTANT]
> We understand that these changes may require some major adjustments in your project, and we apologize for any inconvenience in advance.

khairulhaaziq:
Hi @zernonia ! Big decision! I had my concerns but after several thought and looking back at the project, I think a rebrand is a really good idea! I think what you and other contributors have been building and currently building has turned this library into a very big project, and in many sense it has grown larger than the radix primitives. I think a rebrand is sign of growth and by detaching from the previous radix brand, it removes a significant blocker, allowing decisions to be made without the concern of fitting the radix brand. Overall i'm very excited for the future of the library and see what you'll come up with 🚀🚀🚀

gustojs:
How about something that works both ways? Can be regarded as something separate, but still points at the origins. Vuedix, for example.

MellKam:
@zernonia For a long time, I had an idea to create my own UI library and call it `unbound-ui`. I even reserved the package name and scope for it on npm, but I never had the chance to really start the work. If you like the name, I can transfer the ownership to you. I think it nicely represents the purpose of this library and hasn't been used in other projects yet.

edtorba:
Just a hint, you can also look into Latin words, at least it seems that's what the Radix team did, e.g., radix = root (it seems). What you want is something simple and catchy, while also being unique."

Rolanddoda:
[...]

PR #1432, opened 2024-11-17
chore: improve build

cyyynthia:
This PR supersedes and closes #1102 

It does not use the same approach as proposed in #1102 however; seeing the current v2 build uses `preserveModules`. However, because of how the SFC gets split in multiple virtual chunks this causes a lot of unwanted files to clutter the build making it larger than it needs to be.

The approach proposed in this PR mitigates it by using manualChunks to achieve the intended result of a 1:1 mapping between Vue files and the resulting JS files. A little hack has been introduced as well to properly name chunk files with the right file extension and without a hash.

Minification has been disabled: along with the drop of hashes the intent is to make eventual [dependency patches](https://pnpm.io/cli/patch) less brittle (see last comment of #908 for a use-case). This is also yet-another-attempt at building with sourcemaps enabled 😄

Package size measurements: measured by running `pnpm run build-only; du -sh dist` in `packages/core`.
- Current v2: 7.4M
- This PR (no maps): 4.8M
- This PR (w/ maps): 11M
- [#1102](https://github.com/unovue/radix-vue/pull/1102) chunking approach[^1] (no maps): 2.5M
- [#1102](https://github.com/unovue/radix-vue/pull/1102) chunking approach (w/ maps): 5.8M

[^1]: these numbers were obtained by replacing the manual chunking approach from this PR by the manual chunking approach used in #1102. Does not include the more heavy-handed dual config approach for esm and cjs.

### DIFF
--- a/packages/core/vite.config.ts
+++ b/packages/core/vite.config.ts
@@ -3,8 +3,12 @@ import { defineConfig } from 'vite'
 import vue from '@vitejs/plugin-vue'
 import vueJsx from '@vitejs/plugin-vue-jsx'
 import dts from 'vite-plugin-dts'
+import pkg from './package.json'
 
 const projectRootDir = resolve(__dirname)
+
+// A bit of a hack, but lets us use the proper extension in chunk filenames
+let currentFormat = ''
 
 // https://vitejs.dev/config/
 export default defineConfig({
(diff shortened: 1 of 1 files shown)


zernonia:
Yo @cyyynthia ! Thanks again for this! I totally understand the benefit of having sourceMap, however the thought of another additional 3mb for package size is still haunting me 😬

But seeing that [react-aria](https://packagephobia.com/result?p=react-aria) is 16mb, I think we can get this baby in!

cyyynthia:
Well, to be fair given there isn't any mangling going on anymore I wouldn't find it horrible to not have source maps, in which case this PR saves 2.5MB!

I honestly don't have much opinion about whether the maps are important or not, I pretty much did it as I saw there was some demand 😅

PR #1427, opened 2024-11-15
fix(DateField): Reset field segments when model value is undefined

epr3:
Closes #1426.

PR #1435, opened 2024-11-18
chore(DateField): fix typo in DateField

epr3:
Related to #1427. 

Fixes typo in condition.

### DIFF
--- a/packages/radix-vue/src/DateField/DateFieldRoot.vue
+++ b/packages/radix-vue/src/DateField/DateFieldRoot.vue
@@ -199,7 +199,7 @@ watch([modelValue, locale], ([_modelValue]) => {
   if (_modelValue !== undefined) {
     segmentValues.value = { ...syncSegmentValues({ value: _modelValue, formatter }) }
   }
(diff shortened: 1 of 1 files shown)


epr3:
@DamianGlowala I've created a CodeSandbox with the PR version of the package to reproduce the issue: https://codesandbox.io/p/devbox/n3h95q

But it seems that the segments and the `ref` value are in sync. Could you check out the sandbox and let me know what I'm missing?

Thank you!

DamianGlowala:
Hmm a bit odd, I'll swap my local `pkg.pr.new` version with `radix-vue-1.9.9.tgz` included in the above sandbox and report back ASAP (tomorrow morning/afternoon). Thank you for looking into this so far! 💯

DamianGlowala:
@epr3 would you agree that bumping `radix-vue` in the sandbox to `1.9.10` should clear segments when resetting the model value to `undefined`? Strangely, this doesn't seem to be the case.

epr3:
@DamianGlowala true, but I've overlooked the line of code you pointed out on the last PR (https://github.com/unovue/radix-vue/pull/1427#pullrequestreview-2442289614) and this one fixes it

PR #1444, opened 2024-11-21
fix(Combobox): Pass `highlightOnHover` to Listbox

aguingand:
In v2, the new prop Combobox prop `highlightOnHover` is now passed to ListboxRoot.

`ComboboxRootProps` inherits of `highlightOnHover` prop from `ListboxRootProps`, so it must be passed explicitly.


zernonia:
Ahh thanks for this PR! Sorry I missed this out and pushed a similar changes earlier 🙈 
I'll be closing this ya! 🙏🏼

PR #1448, opened 2024-11-22
Stepper: replace onMousedown by onClick for consistency

nicooprat:
Using `mousedown` event can lead to some edge cases.

In our app, changing step changes the modal content, thus its height, and a mouseup event would then be fired outside the modal: so changing step was actually closing the modal... it was hard to find why!

Also, this would be better for accessibility because I think using `enter` actually trigger the `click` event natively.

I can't think of any reason why using `mousedown` would be better, but I might be wrong!

Related: https://github.com/unovue/radix-vue/discussions/1330

### DIFF
--- a/packages/radix-vue/src/Stepper/StepperTrigger.vue
+++ b/packages/radix-vue/src/Stepper/StepperTrigger.vue
@@ -88,7 +88,7 @@ onUnmounted(() => {
     :tabindex="itemContext.isFocusable.value ? 0 : -1"
     :aria-describedby="itemContext.descriptionId"
     :aria-labelledby="itemContext.titleId"
-    @mouseclick="handleMouseDown"
+    @click="handleMouseDown"
     @keydown.enter.space.left.right.up.down="handleKeyDown"
   >
     <slot />


epr3 (commented):
Hello @nicooprat,

Thank you for the PR. I think we're using the `mousedown` handler to have faster responses from the UI and the ability to handle events earlier in the interaction chain. 

Now that I think of it, I guess `pointerdown` would have been better here for the same reason outlined above plus compatibility with various other pointing devices (touchscreen, stylus, etc.). 

Could you try `pointerdown` and see if it works for your use case?

Otherwise, from my side `click` could work too. 
@zernonia would `click` work here instead of `mousedown`/`pointerdown`?

nicooprat:
Hi, thanks for your answer.

I guess `pointerdown` would be a more modern way of doing it indeed, but it still causes the issue described (just tested).

zernonia:
Yoo @nicooprat  ! The Stepper implementation was heavily ported from the Tab component, which has a similar action on Trigger component (specifically `mousedown`). 

I'm not 100% sure regarding the reason but this is what the other major a11y focus library were doing (namely Radix, [React Spectrum](https://react-spectrum.adobe.com/react-spectrum/Tabs.html#tabs))

nicooprat:
Thanks, I answered in a related discussion in Radix repo: https://github.com/radix-ui/themes/discussions/301#discussioncomment-11470711

zernonia:
Awesome @nicooprat ! I will set this to `on-hold` till we have further insight ya 😁

zernonia:
Closing as this targets the old `main` branch / `packages/radix-vue`, which is no longer maintained. `StepperTrigger` on `v2` still uses `@mousedown`, so we'll carry this change over there. Thanks for tracking this down!

PR #1449, opened 2024-11-22
fix(Menu): Pass `positionStrategy` prop to PopperContent

aguingand:
In v2, the new PopperContent `positionStrategy` prop is not forwarded in MenuContent.


zernonia (approved):
Thanks for catching that and the fix 😁

aguingand:
Thank you for this incredible lib !

PR #1456, opened 2024-11-25
Filter a tree

kellerza:
Would you be interested to add filtering to a tree?

Example of filter behavior [here](https://element-plus.org/en-US/component/tree.html#tree-node-filtering)

This PR moves `flattenedItem` and `flattenItems` to utils and adds `flattenFilter` to achieve this

`flattenFilter` needs a filter function and a list of 'visible' branches.
Output the flattenedItems & expanded keys (maybe this could be useful in the future, but no need to do anything with this)
Once filtered, it would be great to expand a certain branch (adding the key to `forceVisible` achieves this)

I would appreciate some guidance on the API

### DIFF
--- a/packages/radix-vue/src/Tree/TreeRoot.vue
+++ b/packages/radix-vue/src/Tree/TreeRoot.vue
@@ -55,20 +55,6 @@ interface TreeRootContext<T = Record<string, any>> {
   handleMultipleReplace: ReturnType<typeof useSelectionBehavior>['handleMultipleReplace']
 }
 
-export type FlattenedItem<T> = {
-  _id: string
-  index: number
-  value: T
-  level: number
-  hasChildren: boolean
-  parentItem?: T
-  bind: {
-    value: T
-    level: number
-    [key: string]: any
-  }
-}
-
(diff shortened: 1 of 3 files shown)


kellerza:
I've taken a new approach. Instead of modifying TreeRoot, I madea new component responsible for filtering the input to TreeRoot, it's called **TreeData**, you pass it _items_ and some filters props and it returns _filteredItems_ in the slot and modifies `expanded`

```ts
  <TreeData
    ref="treedata"
    v-slot="{ filteredItems }"
    v-model:expanded="expanded"
    :items="allItems"
    :filter="textFilter"
    :get-text="(i) => i.title"
  >
    <TreeRoot
      v-model:expanded="expanded"
      :items="filteredItems"
    >
   ...
    </TreeRoot>
  </TreeData>
```

Might be a cleaner approach?

zernonia:
Closing as this targets the old `main` branch / `packages/radix-vue`, which no longer exists. Filtering can be done in userland by filtering the `items` passed to `TreeRoot`. Thanks for exploring this!

PR #1462, opened 2024-11-27
fix: useBodyScrollLock type error

hrubysi:
Fixes type error of useBodyScrollLock introduced in https://github.com/unovue/radix-vue/pull/1428.

`TypeError: Argument 1 ('element') to Window.getComputedStyle must be an instance of Element`.

Apparently, not every target of touchmove event has to be Element. Also good example of why type casting should be avoided.


zernonia (approved):
Thanks for the fix @hrubysi !

PR #1468, opened 2024-12-03
fix: remove use of `cssText` in focus-guards

sadeghbarati:
Violates Content-Security-Policy unless `style-src: 'unsafe-inline'` is permitted otherwise. (from radix-ui)


zernonia:
Thanks for the fix @sadeghbarati !

PR #1473, opened 2024-12-03
build: replace `fast-deep-equal` with `dequal`

sadeghbarati:
Found a better replacement for `fast-deep-equal` here

https://github.com/SukkaW/eslint-config-sukka/blob/master/packages/shared/src/restricted-import.ts#L5

Also, is there any reason not to update the dependencies?

### DIFF
--- a/packages/radix-vue/package.json
+++ b/packages/radix-vue/package.json
@@ -93,16 +93,16 @@
     "vue": ">= 3.2.0"
   },
   "dependencies": {
-    "@floating-ui/dom": "^1.6.7",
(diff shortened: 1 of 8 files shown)


zernonia:
@sadeghbarati  at this stage we are not bumping the dep version much to prevent any possible bug

Hoping not to touch `fast-deep-equal` for now as well. However this will be replaced by [ohash](https://github.com/unjs/ohash?tab=readme-ov-file#isequalobj1-obj2-options) in v2

PR #1477, opened 2024-12-04
feat(Avatar): support referrer policy in `AvatarImage.vue`

sadeghbarati:
(from radix-ui)


sadeghbarati:
I don't know when Vue is gonna support `fallthroughAttributes` and `advanced types support` for **defineProps** 

cause I have a feeling that this is not ok to have an img component that doesn't have related attributes or props when we are in the IDE :'(

zernonia:
Thanks for the PR @sadeghbarati ! 💚

PR #1481, opened 2024-12-05
feat(NavigationMenu)!: replace hardcoded CSS values with custom properties

gu-stav:
Instead of using custom CSS, the `NavigationMenuIndicator` should rely on custom properties, similar to what [`TabsIndicator`](https://reka-ui.com/docs/components/tabs#indicator) is using.

Closes https://github.com/unovue/radix-vue/issues/1470

### DIFF
--- a/docs/components/demo/NavigationMenu/tailwind/index.vue
+++ b/docs/components/demo/NavigationMenu/tailwind/index.vue
@@ -136,7 +136,7 @@ const currentTrigger = ref('')
       </NavigationMenuItem>
 
       <NavigationMenuIndicator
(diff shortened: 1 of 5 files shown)


gu-stav:
@zernonia I am wondering if these properties should be prefixed with `--reka-navigation-navigation-menu`, since the component is called `NavigationMenu` rather than `Menu`? I followed the existing naming conventing for this PR.

zernonia (changes requested):
Could you help update the demo as well? https://reka-ui.com/docs/components/navigation-menu  😁

run `pnpm build` then `pnpm docs:dev`

PR #1482, opened 2024-12-05
feat(NavigationMenu): expose item position to MenuContent

gu-stav:
Similar to https://github.com/unovue/radix-vue/pull/1481 where I'd like to have the item position and size exposed, I'd like to have it available for `Content`.

This allows to position the content overlay underneath the active item, rather than only relative to the overall component.

Example

Screenshot 2024-12-05 at 19 01 17

If you find this useful, I could also add an example story?

### DIFF
--- a/docs/content/docs/components/navigation-menu.md
+++ b/docs/content/docs/components/navigation-menu.md
@@ -179,6 +179,19 @@ Contains the content associated with each trigger.
   ]"
 />
 
+<CssVariablesTable
+  :data="[
+    {
+      cssVariable: '--reka-tabs-navigation-menu-content-size',
+      description: 'The size of the item.',
+    },
+    {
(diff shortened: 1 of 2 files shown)


zernonia:
Actually you can achieve this easily by not using `NavigationViewport` 

https://stackblitz.com/edit/epoeew?file=src%2FApp.vue,package.json

gu-stav:
Thanks, I had not though about this and will try it out! Maybe this would make another nice example, because I'd think it is a common use-case? I you want I could try and create one.

gu-stav:
Works like a charm. Thanks @zernonia. In case you'd like me to add an example let me know.

zernonia:
Yeah would love to have an example for this 😁 Could you help with that?

PR #1485, opened 2024-12-06
fix(DateField): error thrown when value is null and changing locale

zernonia:


packages/radix-vue/src/DateField/DateFieldRoot.vue
```diff
@@ -192,12 +192,12 @@ watch(locale, (value) => {
...
+  if (!isNullish(_modelValue) && (!isEqualDay(placeholder.value, _modelValue) || placeholder.value.compare(_modelValue) !== 0))
     placeholder.value = _modelValue.copy()
 })
 
 watch([modelValue, locale], ([_modelValue]) => {
-  if (_modelValue !== undefined) {
+  if (!isNullish(_modelValue)) {
     segmentValues.value = { ...syncSegmentValues({ value: _modelValue, formatter }) }
   }
   else if (Object.values(segmentValues.value).every(value => value === null) || _modelValue === undefined) {
```
DamianGlowala, line 203: @zernonia, based on [this PR](https://github.com/unovue/radix-vue/pull/1427/files), I think there is `else if` branch to be corrected too (including DateRangeFieldRoot.vue)

PR #1510, opened 2024-12-18
fix(PopoverContent): focus trap

wouterlms:


zernonia:
Hi @wouterlms .. do you mind sharing what's the idea behind these changes?

wouterlms:
@zernonia I have a use case in a project where I want to disable the focus trap in a popover. I noticed in the docs that there’s a prop for disabling it, but it doesn’t seem to have any effect. This PR addresses that. Or am I misunderstanding, and is it intended not to work?

zernonia:
Ahhh thanks @wouterlms ! I understand the problem here. By right the `trapFocus` should be internal prop and user should use `modal` to trap the or untrap the focus.

zernonia:
Closing this as found out the problem is with the internal props being exposed.
Thanks for bringing the issue up and the PR 🙏🏼

https://github.com/unovue/radix-vue/commit/6b5af1627d514a47cdb0dd765458edaeb2d79f9e

wouterlms:
@zernonia I tried to use the `modal` prop but it doesn't seem to have any effect. The focus is trapped with modal set to either `true` or `false`

PR #1513, opened 2024-12-19
feat(SelectItem): Emit a select event when an option is selected

kilobyte2007:
I've implemented a `select` event/emit on the `SelectItem` so it's consistent with `Listbox/Combobox`.
This is especially important when using `multiple` because we cannot rely on `update:modelValue` as we only need the specific item that was last selected.

Let me know if you have any questions or objections!

### DIFF
--- a/packages/core/src/Select/SelectItem.vue
+++ b/packages/core/src/Select/SelectItem.vue
@@ -1,7 +1,7 @@
 <script lang="ts">
 import type { Ref } from 'vue'
 import type { PrimitiveProps } from '@/Primitive'
-import { createContext, useForwardExpose, used, handleSelectCustomEvent } from '@/shared'
+import { createContext, useForwardExpose, useId, handleSelectCustomEvent } from '@/shared'
 import type { AcceptableValue } from '@/shared/types'
 import { useCollection } from '@/Collection'
 
--- a/packages/core/src/Select/SelectItem.vue
+++ b/packages/core/src/Select/SelectItem.vue
@@ -1,7 +1,7 @@
 <script lang="ts">
 import type { Ref } from 'vue'
 import type { PrimitiveProps } from '@/Primitive'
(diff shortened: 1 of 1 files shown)


kilobyte2007:
I see there's a failing check. But it doesn't seem to be caused by my changes as I didn't touch that part of the code.

zernonia:
Yoo @kilobyte2007 ! I've fixed the test issue, can you please rebase the branch on `origin/main` or merge from `origin/main`? 😁

packages/core/src/Select/SelectItem.vue
```diff
@@ -58,9 +66,20 @@ const isFocused = ref(false)
 const textValue = ref(props.textValue ?? '')
 const textId = useId(undefined, 'reka-select-item-text')
 
-async function handleSelect(ev?: PointerEvent) {
+const SELECT_SELECT = 'select.select'
+
+async function handleSelectCustomEvent(ev: PointerEvent | KeyboardEvent) {
+  if (ev.defaultPrevented)
+    return
```
zernonia, line 73: I think we shouldn't have this conditional here, it will prevent the `select` from being emitted.

kilobyte2007: It is needed because this logic is handling the initial mousedown event, when the dropdown just shows up. Without this check the dropdown just shows up on mousedown and disappears on mouseup.

So basically the check that we had previously - has moved from `handleSelect` to `handleSelectCustomEvent`, as this is the place now where we handle the original event. And the check that we have now in `handleSelect` is for actually preventing the artificially created event.

zernonia: Ahh make sense!

zernonia:
Thanks for the feature OR @kilobyte2007 ! Looking good! 😁

PR #1518, opened 2024-12-21
fix(PinInput): display placeholders after resetting pin value programmatically

Reeska:
Hello,

This PR is meant to fix the following issue https://github.com/unovue/radix-vue/issues/1516.
It's about the PinInput component that lost placeholders when value is reset programmatically.


zernonia:
Thanks for the PR @Reeska . I've however adapted and pushed a fix here https://github.com/unovue/radix-vue/pull/1530

Reeska:
Thanks !

PR #1538, opened 2025-01-03
feat(Editable): update model value only on submit

hrynevychroman:
I have faced some issues with understanding how to make a field editable and not change `v-model` when editing and found this example in the Histoire files of the project, I think that this would be useful for others 😉

### DIFF
--- a/docs/content/components/editable.md
+++ b/docs/content/components/editable.md
@@ -150,33 +150,20 @@ Contains the cancel trigger of the editable component.
 
 ### Change only on submit
 
-Use the `default-value` prop to define starting value and `@submit` emit to handle the submit event.
-This is useful when you want to change the value only when the user submits the changes and have `auto-resize` working properly.
+By default the component will submit when `blur` event triggers. We can modify the `submit-mode` prop to alter this behavior.
+In this case, we want to submit only when user click on `EditableSubmitTrigger`, so we change the submit mode to `none`.
 
-```vue line=12-13
-<script setup lang="ts">
-const defaultValue = ref('Default value')
-
-function handleSubmit(label?: string) {
-  defaultValue.value = label
-}
-</script>
-
+```vue line=2,8
 <template>
-  <template>
-    <EditableRoot
-      :default-value="defaultValue"
-      @submit="handleSubmit"
-    >
-      <EditableArea>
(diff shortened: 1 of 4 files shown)


docs/content/components/editable.md
```diff
@@ -146,6 +146,40 @@ Contains the cancel trigger of the editable component.
...
+
+function handleSubmit(label?: string) {
+  defaultValue.value = label
+}
+</script>
+
+<template>
+  <template>
+    <EditableRoot
+      :default-value="defaultValue"
```
zernonia, line 168: how about using `:model-value` instead of `default-value`? 🤔

hrynevychroman: this will not work, I think because of `useVModel` inside `Editable`, we don't make copy to have ability to edit it, refer to stackblitz and video examples 🙌 

Stackblitz: https://stackblitz.com/edit/adsgveqt?file=src%2FApp.vue

Video:

https://github.com/user-attachments/assets/52fb1b3f-3669-4c3e-8270-24df4440a7fa

zernonia: Thanks @hrynevychroman ! I understand the challenge here now. Looking at the example made me realized this might be a pattern that is quite unintuitive.

I would like to refactor `EditableInput` to only update modelValue when we commit it, this inline with the `NumberFieldInput`'s behavior. So when user `commit` the changes, it updates `v-model`, if `cancel`, nothing happened. I believe this is something you are looking for right?

WDYT?

hrynevychroman: @zernonia Yeah, this is something I'm looking for, I will take a look and open a PR when I have some free time, probably this Saturday 🙌

zernonia: Thanks for the changes @hrynevychroman ! I've fixed the test, I think we can update this section of the doc as well to reflect the new behavior of `modelValue`

zernonia: @hrynevychroman  I've updated the docs ya 😁

zernonia (approved):
Thanks for the PR and the refactor ya @hrynevychroman ! Merging 🚀

hrynevychroman:
@zernonia, thank you a lot! 💚

PR #1546, opened 2025-01-06
feat(Teleport): support Vue teleport defer prop

hwkr:
Vue 3.5 added a deffered prop to the built in `<teleport>` component which allows you to defer the teleport until the end of the current render cycle.

This PR exposes it on the Radix-Vue `Teleport` component but I'm not sure if we need to bump the Vue dependency for the project to be 3.5 as well 🤔

### DIFF
--- a/packages/radix-vue/src/Teleport/Teleport.vue
+++ b/packages/radix-vue/src/Teleport/Teleport.vue
@@ -13,6 +13,13 @@ export interface TeleportProps {
    */
   disabled?: boolean
   /**
+   * Defer the resolving of a Teleport target until other parts of the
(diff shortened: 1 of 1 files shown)


zernonia:
Thanks for the PR @hwkr ! This is a good suggestion. We don't neccessary need to bump the peer dependency 😁

Will be closing this as feature will only be pushed to v2 https://github.com/unovue/radix-vue/commit/17d9deba20a0cb7f6fa7d0f0d48673bd33e16669

hwkr:
@zernonia Sounds good! Is there a timeline on V2 right now?

PR #1558, opened 2025-01-15
Infer-multiple

zernonia:


zernonia:
Closing this stale draft — it has diverged from `v2`. Will revisit the `multiple` type inference separately if needed.

PR #1560, opened 2025-01-16
feat(DateField): improve month segment behavior for invalid months starting with 1

acabreragnz:
## Changes
- Modified the month segment behavior in DateField to handle invalid months starting with 1 (13-19) more intuitively
- When typing invalid months like 13-19:
 - Keeps 1 as the month value
 - Automatically uses the second digit as the initial value for the day segment
 - Moves focus to day segment

## Example
When typing "13":
- "1" remains as the month value
- Focus moves to day segment
- "3" becomes the initial value of the day

## Why
This change improves user experience by making the input behavior more intuitive when users type invalid month numbers. Instead of discarding the first digit, it uses both digits meaningfully - preserving the valid part (1) for the month and using the second digit to pre-fill the day.

## Testing
- [ ] Type "13" - should keep "1" as month and move to day with "3"
- [ ] Type "14" - should keep "1" as month and move to day with "4"
- [ ] Type "10", "11", "12" - should work as before (valid months)
- [ ] Type "2"-"9" - should work as before (single digit months)


epr3 (changes requested):
Hello @acabreragnz,

First of all, thank you for your contribution! 🚀 

The feature is very good improvement to the UX, but I have one concern about it, namely if you change the `locale` to `fr` or any other `locale` which uses the `dd/mm/yyyy` format, it moves the focus on the year segment and updates the day which doesn't look like the desired behaviour.

You could add a check to see which is the next segment and/or maybe expand this functionality for the `dd/mm/yyyy` format and `yyyy/mm/dd`, the latter being used by locales such as `zh`.

acabreragnz:
@epr3 thank you for your feedback! the PR was not ready for review; it was my mistake for not setting it as a draft. 

Currently, I'm finalizing the same behavior for transitioning from days to years in the `en` locale.
For instance, when typing the day `34` we should expect `MM/3/4YYY` instead of `MM/4/YYYY`.

Once this is complete, I will implement all the changes you requested.

PR #1570, opened 2025-01-21
Fix document.activeElement not being correct when components are use inside custom elements

### LINKED ISSUE #1564: [Bug]: Web component DropdownMenu and Select arrow key bug
### Environment

```bash
Node version: v18.20.3
Radix Vue version: 1.9.12
Vue version: 3.5.13

Client OS: macOS 12.7.6
Browser: Chromium Engine Version 131.0.6778.265
```

https://stackblitz.com/edit/nexbpjt8-wvsqpwgv?file=src%2Fcustom-elements%2FCustomElement.ce.vue

### Steps to reproduce

Related issue https://github.com/unovue/radix-vue/issues/1003

- Create a custom element with a DropdownMenu or a Select
- Make sure the Portal has the prop `disabled` set to `true`
- Click toggle to open DropdownMenuContent or SelectContent
- Press down arrow to navigate the options

This can be tested using the reproduction link.

### Describe the bug

Since the component is inside a custom element it seems like the current `activeElement` is incorrect. From my tests `document.activeElement` when using inside a custom-element is the custom element itself rather than the selected option.

I think the follow code may be the problem line may be the problem:
- https://github.com/unovue/radix-vue/blob/290ae2914f2ce405c4af2c3807f46d9f191e3f34/packages/radix-vue/src/Menu/MenuContentImpl.vue#L179
- https://github.com/unovue/radix-vue/blob/290ae2914f2ce405c4af2c3807f46d9f191e3f34/packages/radix-vue/src/Menu/utils.ts#L38

It seems like this has already been addressed for radio groups so possibly the same fix could work https://github.com/unovue/radix-vue/pull/1314

### Expected behavior

I expect the arrow key navigation to work the same inside a custom element as it does outside.

https://github.com/user-attachments/assets/15376ff0-914a-454a-bc98-f5db8b68c919

pmairoldi:
Fixes #1564 
Reverts fixes from #1314 

This pull requests address the fact that `docuement.activeElement` will be incorrect when components are used inside of a custom element. After some tests here is what I found: 

1. if the focus is inside a custom element the `document.activeElement` will be the host of the custom element and the actual active element can be retrived by `activeElement.shadowRoot.activeElement`
2. if a custom element is within another custom element includes another custom element as a child and that child has an element that is focused then the same trick as point 1 can be used but one level deeper `activeElement.shadowRoot.activeElement..shadowRoot.activeElement`

With these findings I have implemented a function `useActiveElement` to retrieve the `activeElement` regardless of the nested `shadowRoot`. With this function I've replace all instances of `document.activeElement` with `useActiveElement()`. After some tests it seems like everything still works and we do not need the fixes from #1314 anymore.


packages/radix-vue/src/shared/useActiveElement.ts
```diff
@@ -0,0 +1,12 @@
...
+  if (activeElement == null) {
+    return null
+  }
+
+  while (activeElement != null && activeElement.shadowRoot != null && activeElement.shadowRoot.activeElement != null) {
+    activeElement = activeElement.shadowRoot.activeElement
+  }
+
+  return activeElement
+}
```
pmairoldi, line 12: This function will find the deepest active element regardeless of what shadowRoot it is in.

packages/radix-vue/src/shared/useActiveElement.ts
```diff
@@ -0,0 +1,12 @@
+export function useActiveElement(): Element | null {
```
zernonia, line 1: Since this is plain js function, and need not to rely on Vue's reactivity, let's not use the composable naming convention (`useXXX`).

Instead let's rename this function to be `getActiveElement` 😁

zernonia:
Thanks for the awesome PR @pmairoldi ! Some minor comment on the function naming 😁

packages/radix-vue/src/shared/useActiveElement.ts
pmairoldi: Fair enough. I’ll update it sometime today. Thanks for the review

zernonia (approved):
Nicely done @pmairoldi !!! Thanks for the PR! 🎉🚀

PR #1596, opened 2025-01-30
Fix aria-required not set to required prop value for Checkbox

pmairoldi:
Fixes https://github.com/unovue/radix-vue/issues/1595

This pull requests addresses the issue that the CheckboxRoot component always sets the `aria-required` attribute to false regardless of what the `required` props is.


pmairoldi:
Anytime 😃 

I was wondering though, why implement all this instead of just using an input and using a sibling selector to style extra elements (https://v3.tailwindcss.com/docs/hover-focus-and-other-states#differentiating-peers)? Browser consistency?

PR #1601, opened 2025-02-03
Update Primitive.ts

sedmedgh:
added all slots rendering on Custom component as and reactive as and asChild props changes


sedmedgh:
Without these changes, your component will only render the first child of the default slot.

zernonia:
Closing as this is out of date with `v2` (it reverts the docs link and imports `./Slot.ts`). The underlying issue — `as`/`asChild` not being reactive in `Primitive` — is noted and will be handled separately. Thanks!

PR #1606, opened 2025-02-03
fix(PinInput): error throw from resetPlaceholder's nextTick if component not in the DOM anymore

hartbit:
This resolves the issue I discovered here: https://github.com/unovue/radix-vue/issues/1605
I have a reproduction example app (https://stackblitz.com/edit/bhw6xwfr-hcdmactw?file=src%2FApp.vue,package.json) but I failed to reproduce it in the unit tests :(


zernonia (approved):
Thanks for the fix @hartbit ! 🙏🏼

PR #1610, opened 2025-02-08
fix(DatePicker): add missing parentheses to defineSlots

donaldxdonald:
add missing parentheses to defineSlots.


zernonia (approved):
Thanks for capturing that @donaldxdonald ! 🙏🏼

PR #1612, opened 2025-02-10
fix(RangeCalendar): enable modelValue to be undefined and refactor combined watch condition

### LINKED ISSUE #1611: [Bug]: RangeCalendar display does not update when cleared programmatically
### Environment

```bash
Development/Production OS: Windows 11
Node version: 22
Package manager: pnpm@9
Radix Vue version: 1.9.13
Vue version: 3
CSS framework: tailwindcss@3
Client OS: Windows 11
Browser: Firefox
```

https://stackblitz.com/edit/sjvn9m-8dgymjir?file=src%2FApp.vue,package.json

### Steps to reproduce

1. Open repro
2. Optionally change date
3. Click 'clear' button

### Describe the bug

The UI does not update when the model value's start/end are programmatically set to undefined, even though the model value does actually change.

### Expected behavior

The selected range should be cleared when the 'clear' button is clicked.

I think the root cause here is that the range calendar tries to completely clear its selection by setting the model value to `{ start: undefined, end: undefined }` when a user starts a new selection. This means that while the selection is in progress, the value of the internal `startValue` does not match the value in the model, even though the watch that keeps them in sync is still active.

I'm guessing that undefined `modelValue.start/end` values are ignored in order to prevent them from overwriting `startValue` during the selection, but it seems more reasonable to me to just keep `startValue` always in sync with `modelValue.start` by setting `modelValue` to `{ start: startValue.value, end: undefined }` until the selection is completed. Then, it's up to the caller to handle partial selections as needed.

Btw, here are some ways that partial selections may need to be handled for various use cases:

- treat as start/end undefined (current behavior)
- treat as a range where start == end
- allow partial selection with end: undefined
- allow partial selection with start: undefined
- ignore and use previous selection instead
- ignore and use default value instead
- a date range picker in a popup may need to ignore the partial selection while the popup is open, but commit a value that matches one of the above if the selection is still partial when the popup closes, and may also need different functionality if the selection is cancelled with the esc key

Since there are many ways partial selections may need to be handled, I think it makes the most sense to just expose them to the caller and let them deal with it rather than trying to do anything "intelligent."

epr3:
Closes #1611.

### DIFF
--- a/packages/radix-vue/src/RangeCalendar/RangeCalendarRoot.vue
+++ b/packages/radix-vue/src/RangeCalendar/RangeCalendarRoot.vue
@@ -250,13 +250,11 @@ const {
 })
 
 watch(modelValue, (_modelValue) => {
-  if (_modelValue.start) {
-    if (!startValue.value || !isEqualDay(startValue.value, _modelValue.start))
-      startValue.value = _modelValue.start.copy()
+  if (!_modelValue || !_modelValue.start || (startValue.value && !isEqualDay(_modelValue.start, startValue.value))) {
+    startValue.value = _modelValue?.start?.copy?.()
   }
-  if (_modelValue.end) {
-    if (!endValue.value || !isEqualDay(endValue.value, _modelValue.end))
-      endValue.value = _modelValue.end.copy()
+  if (!_modelValue || !_modelValue.end || (endValue.value && !isEqualDay(_modelValue.end, endValue.value))) {
+    endValue.value = _modelValue?.end?.copy?.()
   }
 })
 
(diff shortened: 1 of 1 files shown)


zernonia:
@epr3  seems like setting to `undefined` (range = undefined) is throwing some error, I think we should cater for this case too 😁

epr3:
@zernonia Could you provide a screenshot of the error? I don't seem to be able to reproduce any error on Histoire.

https://github.com/user-attachments/assets/4c3d0947-bc8f-4fb3-9113-56b88cab61aa

PR #1619, opened 2025-02-13
fix(DatePicker): error when unselecting a date

hartbit:
This resolves https://github.com/unovue/radix-vue/issues/1618.

Again, I failed to reproduce the issue in unit tests. I tried this, unsuccessfully:

```ts
  it('does not throw when selecting the same date twice', async () => {
    const { user, trigger, getByTestId, rerender } = setup({
      datePickerProps: { modelValue: calendarDate },
      emits: { 'onUpdate:modelValue': (data: DateValue) => rerender({ modelValue: data }) },
    })

    await user.click(trigger)
    const selectedDay = getByTestId('calendar').querySelector<HTMLElement>('[data-selected]')!
    await user.click(selectedDay)
  })
```

### DIFF
--- a/packages/radix-vue/src/DateField/DateFieldRoot.vue
+++ b/packages/radix-vue/src/DateField/DateFieldRoot.vue
@@ -192,7 +192,7 @@ watch(locale, (value) => {
 })
 
 watch(modelValue, (_modelValue) => {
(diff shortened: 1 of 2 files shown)


zernonia:
Thanks for the @hartbit ! Yeah the unit test can be abit tricky sometimes 🙈

PR #1625, opened 2025-02-15
fix: focus scope trap

wouterlms:
Currently, even when `modal` is set to `false` in the Popover component, focus remains trapped. This PR resolves the issue by ensuring focus behaves as expected when `modal` is disabled.


zernonia:
Hi @wouterlms , `modal` props is to control interaction with outside elements, when `true` outside elements will be disabled and only popover content will be visible to screen readers.

In your case you need to use `trapFocus=false`

wouterlms:
@zernonia `trapFocus="false"` doesn't seem to change anything, isn't this fix still relevant?

PR #1629, opened 2025-02-18
fix: remove type casting in useBodyScrollLock (radix-vue)

hrubysi:
Follow-up to  #1462, which was not comprehensive solution enough to fix occuring type errors.


zernonia (approved):
Thanks for the fix @hrubysi  😁

PR #1630, opened 2025-02-18
fix: remove type casting in useBodyScrollLock (reka-ui)

hrubysi:
Follow-up to  #1462, which was not comprehensive solution enough to fix occuring type errors.


zernonia:
Closing this as it will be cherry-picked 😁

PR #1632, opened 2025-02-19
fix(useForwardExpose): fix #1631

### LINKED ISSUE #1631: [Bug]: `useForwardExpose` (forwardRef) should forward ref to native DOM element
### Environment

```bash
Node version: 20.11.0
Package manager: pnpm@9.12.1
Radix Vue version: 1.9.15
Vue version: 3.4.31
```

--

### Steps to reproduce

Create a component that uses `useForwardExpose` to forward a certain element, like `span` in this example:
```
// Comp.vue
<div>
  <span :ref="forwardRef" />
</div>
```
And use this component with `ref`:
```
<Comp ref="el" />
```
Now `el.value.$el` is the `div`, and not the `span` I want.

### Describe the bug

This is about [`useForwardExpose`](https://github.com/hartbit/radix-vue/blob/main/packages/radix-vue/src/shared/useForwardExpose.ts). I expect it to be like React's `useRef` that can forward a reference to the DOM object, but it currently can only expose components, not native DOM.

This is from the [docs](https://www.radix-vue.com/utilities/use-forward-expose):
```
<!-- We want to expose div as the template ref's element -->
    <div :ref="forwardRef">
```
And it doesn't work, it won't expose the DOM to this div with `.$el`.

While this works
```
<div :ref="_ => forwardRef({ $el: _ })">
```
From the docs, it said `useForwardExpose` aims to target certain element as the expose element.

### Expected behavior

_No response_

_No response_

felissi:
This fixes #1631. 
This makes exposing the targeted element with `useForwardExpose` possible.

### DIFF
--- a/packages/radix-vue/src/shared/useForwardExpose.test.ts
+++ b/packages/radix-vue/src/shared/useForwardExpose.test.ts
@@ -1,0 +1,165 @@
+import { flushPromises, mount } from '@vue/test-utils'
+import { describe, expect, it } from 'vitest'
+import { Fragment, computed, defineComponent, h, ref } from 'vue'
+
+import { useForwardExpose } from './useForwardExpose'
+
+describe('useForwardRef', async () => {
+  it('should forward plain DOM element ref', async () => {
+    const comp = defineComponent({
+      setup() {
+        const { forwardRef } = useForwardExpose()
+        return { forwardRef }
+      },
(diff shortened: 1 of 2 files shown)


zernonia (approved):
Thanks for reporting the issue and the fix @felissi ! 🙏🏼

PR #1635, opened 2025-02-19
fix: do not call useId() conditionally

g12i:
Calling `useId()` conditionally leads to issues when using Reka's based RadioGroup in [Historie](https://histoire.dev/):

```
runtime-core.esm-bundler.js:1488 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading '0')
    at useId (useId.ts:22:29)
    at ComputedRefImpl.fn (RovingFocusItem.vue:27:46)
    at ComputedRefImpl.fn (RovingFocusItem.vue:29:47)
```

This also happens in Reka's Historie. To reproduce:
- Start Historie
- Open RadioGroup story
- Open another story
- Go back to RadioGroup
- See error in console


zernonia (approved):
Thanks for the fix @g12i ! 🙏🏼

PR #1640, opened 2025-02-20
fix(DateField): do not reset all segments when user delete value from 1 segment

zernonia:


jack-allocate:
@zernonia Thanks for the quick turnaround!

Will this get pushed to radix-vue? Haven't migrated to reka-ui yet

PR #1652, opened 2025-02-23
feat: `useBodyScrollLock` supports optional lock pointer events

serkodev:
This PR allows users to pass the following state to `useBodyScrollLock` to optionally lock scrolling and pointer events:

```ts
interface UseBodyScrollLockState {
  scroll?: boolean
  pointerEvents?: boolean
}
```

This change remains backward compatible since it still supports `boolean`. When set to `true`, both scroll and pointerEvents are locked as before.

## Why

In my project, I needed to lock the body scroll. However, using VueUse’s `useScrollLock` caused some issues. For instance, after locking the body scroll and manually setting the body’s `padding-right`, opening and closing a Radix Dialog would clear the `padding-right`.

To address this, I decided to use Radix’s `useBodyScrollLock` for locking the body scroll. By default, it will set `pointerEvents` to `none`, which prevents clicking on any elements.

While a workaround like the one below exists:

```ts
const locked = useBodyScrollLock()
locked.value = true
nextTick(() => {
  document.body.style.pointerEvents = ''
})
```

It didn’t feel intuitive to me. I believe that "scroll locking" should not disable all pointer events. Therefore, I’ve added an option for `pointerEvents` in `useBodyScrollLock` so that users can choose whether to lock pointer events or not.

### DIFF
--- a/packages/radix-vue/src/shared/useBodyScrollLock.ts
+++ b/packages/radix-vue/src/shared/useBodyScrollLock.ts
@@ -9,16 +9,37 @@ import { injectConfigProviderContext } from '@/ConfigProvider/ConfigProvider.vue
 
 import { nanoid } from 'nanoid/non-secure'
 
+interface UseBodyScrollLockState {
+  scroll?: boolean
+  pointerEvents?: boolean
+}
+
 const useBodyLockStackCount = createSharedComposable(() => {
-  const map = ref<Map<string, boolean>>(new Map())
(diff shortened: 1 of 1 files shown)


serkodev:
@zernonia is there a wrongly merge to my branch? my original PR commit is only https://github.com/unovue/reka-ui/pull/1652/commits/692d073f4b8aa2c0fea227cf103a12a6ce2a99ef

BTW is there any comment on this PR? thank you!

PR #1655, opened 2025-02-25
chore: fix site url

divine:
Well, **`reka.com`** is quite expensive domain name, so reverting it back to **`reka-ui.com`**

### DIFF
--- a/packages/core/src/Dialog/utils.ts
+++ b/packages/core/src/Dialog/utils.ts
@@ -24,7 +24,7 @@ export function useWarning({
 
 If you want to hide the \`${titleName}\`, you can wrap it with our VisuallyHidden component.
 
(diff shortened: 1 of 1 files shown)


packages/core/src/Dialog/utils.ts
```diff
@@ -24,7 +24,7 @@ export function useWarning({
 
 If you want to hide the \`${titleName}\`, you can wrap it with our VisuallyHidden component.
 
-For more information, see https://www.reka.com/components/${componentLink}`
+For more information, see https://www.reka-ui.com/components/${componentLink}`
```
Chilfish, line 27: It should be `https://reka-ui.com/docs/components/${componentLink}`

Issue #1669, opened 2025-02-28
[Feature]: Adding open to slot for AlertDialog


Ericlm:
### Describe the feature

It would be great if the AlertDialog component could have the same exposed v-slot as the Dialog, with the open state 🙂 

### Additional information

- [x] I intend to submit a PR for this feature.
- [ ] I have already implemented and/or tested this feature.

PR #1671, opened 2025-03-01
feat: add open prop slot on AlertDialog

Ericlm:
Add a slot to the `AlertDialog` just like the `Dialog` component, to expose the open state of the dialog.

#1669


packages/core/src/AlertDialog/AlertDialogRoot.vue
```diff
@@ -24,6 +30,6 @@ useForwardExpose()
     v-bind="forwarded"
     :modal="true"
   >
-    <slot />
+    <slot :open="forwarded.open" />
```
zernonia, line 33: It would be better to expose the `open` value from `DialogRoot` because it has internal state


```vue
<DialogRoot v-slot="{ open }">
   <slot :open="open" />
</DialogRoot>

Ericlm: Done 🙂

Ericlm:
Thank you @zernonia! 🙂

PR #1672, opened 2025-03-01
fix(SelectRoot): Fix model generic types

JoJk0:
Fix generic type not being passed into SelectRoot props, causing type errors while using it with generics in parent component.


zernonia:
Fixed via https://github.com/unovue/reka-ui/commit/995ecc43f86842656bd0f63da053f7e65de7d436 🙏🏼

PR #1684, opened 2025-03-07
feat(Popper): Add flip options props

aguingand:
There is some cases when you need to shift position according to boundary but not completely flipping to the other sides (e.g. popovers with contents only making sense underneath a field). I have added `sideFlip` & `alignFlip` props which correspond to [mainAxis](https://floating-ui.com/docs/flip#mainaxis) & [crossAxis](https://floating-ui.com/docs/flip#crossaxis) floating-ui `flip()` options. At first I was thinking of an unique property to disable flipping completely but two prop for each axis seems better for more fine-grained options.


aguingand:
Hello @zernonia @epr3 👋  What do you think about it ?

PR #1688, opened 2025-03-11
In NavigationMenu, allow disabling trigger close on link click

hartbit:
When a `NavigationMenuLink` is clicked inside a `NavigationMenuContent`, it closes the opened menu item. This feature allows disabling this functionality with a `disableLinkClickClose` prop on `NavigationMenuRoot`, similarly to the other props to disable functionality.


zernonia:
@hartbit  we already support this by `@select.prevent` .

https://github.com/unovue/reka-ui/blob/1ae99f95da94c5f36de86606e392d245d02a2b0c/packages/core/src/NavigationMenu/NavigationMenuLink.vue#L7-L13

PR #1693, opened 2025-03-12
feat: export CalendarRoot props

Us3r-gitHub:
### Description
This PR is a follow-up fix for PR #916.

While a fix was previously implemented in PR #916, the issue still persists. In this PR, I have extended the export functionality in the same manner as the previous PR to further address the ongoing issue.


zernonia:
@Us3r-gitHub  why would we want to re-export `DateValue` types ya?

Us3r-gitHub:
> @Us3r-gitHub why would we want to re-export `DateValue` types ya?

Similar to the author of the [shadcn-vue issue](https://github.com/unovue/shadcn-vue/issues/690), my goal is to create a custom component library based on shadcn-vue.

### Problem 1
```sh
src/components/calendar/Calendar.vue:283:1 - error TS2742: The inferred type of 'default' cannot be named without a reference to '.pnpm/@internationalized+date@3.7.0/node_modules/@internationalized/date'. This is likely not portable. A type annotation is necessary.
```

image
The temporary solution I applied was to manually import the DateValue type from @internationalized/date. However, this approach triggers a eslint warning.

Issue #1593, opened 2025-01-30
[Feature] (NumberField): Add Option to Disable snapValueToStep


rooom-nikita:
### Describe the feature

Currently, snapValueToStep is ignored only when no step is provided, which is unlikely in most cases. Since the default step value is 1, this behavior can lead to unexpected rounding, especially when format options like minimumFractionDigits and maximumFractionDigits are specified. This can result in inconsistent number formatting

**Proposed Solutions**
To improve flexibility, I propose one of the following:
1. Add a Prop – Introduce an additional prop to control whether values should snap to the step value.
2. Disable Snapping When Formatting is Applied – Automatically disable snapValueToStep if minimumFractionDigits or maximumFractionDigits are specified, ensuring the input value maintains its intended precision.

**Code Reference**
https://github.com/unovue/radix-vue/blob/33f50be4fd6cb2370fb44b714a3fcde0e020b819/packages/core/src/NumberField/NumberFieldRoot.vue#L147-L157

Would love to hear thoughts on this approach

Additionally, I found a potentially related discussion here: [#977](https://github.com/unovue/radix-vue/discussions/977)

### Additional information

- [ ] I intend to submit a PR for this feature.
- [ ] I have already implemented and/or tested this feature.

rooom-nikita:
And maybe this [#1523](https://github.com/unovue/radix-vue/issues/1523)

emilsgulbis:
Almost lost my mind before found this.
1st option with `snap` props feels the best, "auto snapping" feels really unexpected

EmilMoe:
Thanks for the fix, what is the timeline for the next release? :)

PR #1696, opened 2025-03-13
feat(NumberField/NumberFieldRoot): add stepSnapping prop to disable snapping values to step (#1593)

blakemann:
When a `step` prop is defined (which is generally the case, since it has a default of `1`) for `NumberFieldRoot` component, the value is automatically snapped to an increment of that step. This can be problematic when the step value is intended as a shortcut to help the user make changes, but not necessarily a requirement of the field.

For example, it is commonly desired for a number field to use a step of `1` if whole integers are the most common input, but still allow a user to enter a value such as `1.5` if necessary.

This pull request adds a new prop called `stepSnapping` which controls whether this value to step snapping occurs. The default value is `true` for backwards compatibility, but can be set to `false` to disable the snapping behaviour (does not affect snapping/clamping to min/max values).

Addresses #1593

### DIFF
--- a/docs/content/meta/NumberFieldRoot.md
+++ b/docs/content/meta/NumberFieldRoot.md
@@ -80,6 +80,13 @@
     'type': 'number',
     'required': false,
     'default': '1'
+  },
+  {
+    'name': 'stepSnapping',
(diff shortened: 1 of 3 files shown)


zernonia (approved):
Nicely done! Seems to be working well! Thanks for the PR! 💚

PR #1707, opened 2025-03-14
feat(Slider): add new prop to exclude thumb inbound offset

zernonia:


MickL:
Hello @zernonia , thank for the fix! Has this been released with 2.1.0? Because the prop has no effect and I dont see it in the props:

Bildschirmfoto 2025-03-17 um 17 34 56

cyyynthia:
It's been renamed to thumbAlignment="overflow"

I looked for it myself this afternoon 😅

Issue #1720, opened 2025-03-18
[Bug]: NumberFieldRoot does not emit `update:modelValue` events while user is typing


blakemann:
### Environment

```bash
Node version: 18.19.1
Package manager: yarn@1.22.22
Reka UI version: 2.1.0
Vue version: 3.5.13
Client OS: Windows 10 19045.5608
Browser: Chrome 134.0.6998.89
```

### Link to minimal reproduction

https://codesandbox.io/p/devbox/reka-ui-numberfield-forked-8yd5fd?workspaceId=ws_NdaS5AXB6ZiaVxtpjP9vmS

### Steps to reproduce

Implement a Number Field component into a parent and provide a `modelValue` prop and listen for `update:modelValue` events within the parent, and log or display the updated value on event.

Click into the field and begin typing - no events will be fired. Click out of the field to cause a `change` event to trigger. Only now does the event fire.

### Describe the bug

This may be debatable if this is a bug or intended functionality, and I am happy to close if the latter is decided.

If you use `v-model` on a native `input` with `type="number"` you will find that the model value is changed dynamically while typing in the field (ie, if I click into the field and type `123` and observe the model value, I will see it subsequently change to `1`, then `12`, then `123`.

The Reka UI Number Field does not behave in the same way, as it currently only will update the model value on a `change` event (aside from updates caused by wheel spin, arrow keys, increment/decrement buttons, etc.) which is only fired when focus leaves the field (which is also when formatting/clamping is applied to the user entered value). In this case, if I type `123` and observe the model value, I will only see it change once, straight to `123`.

This can be somewhat limiting in cases where a UI might respond dynamically to value changes while a user is typing in a field.

It is my opinion that this component should behave more consistently with a native number input, and emit `update:modelValue` events after every `input` event instead of only on `change` events (effectively, after every keypress). I'm _not_ suggesting making any changes to when formatting/clamping/snapping is applied to the field, as the current behaviour is ideal and prevents things from changing while the user is still in the middle of making edits... however, the component can still pre-calculate what the formatted/clamped/snapped value will be and emit those events while the user types.

I have implemented changes to address this and will open a PR shortly.

### Expected behavior

NumberFieldRoot component emits `update:modelValue` events when the child `input` element emits an `input` event, instead of only on `change` event.

The value provided in the emitted events continues to be formatted/clamped/snapped the same way as it does currently on `change`.

[...]

PR #1721, opened 2025-03-18
fix(NumberField/NumberFieldRoot): emit update:modelValue events after child input event (#1720)

blakemann:
Previously, the NumberFieldRoot component would only emit `update:modelValue` events when the child input element emitted a `change` event, which only occurs after the user has finished making changes and leaves the field. This is not consistent with the behaviour you would see on a native input where the value changes after every `input` event, after each keypress and intermediary value entered.

This PR makes changes to how the text value is displayed and how the model value is emitted, allowing the model value to emit more frequently as the user types (after an `input` event occurs), while not affecting the existing behaviour of when any formatting, clamping, or snapping is applied to the displayed value.

Additionally, the emitted model value pre-calculates any clamping or snapping to ensure that it reflects what the displayed text value will change to if the user were to hit 'enter' or leave the field.

This is done by removing the existing `computed` property which tied the displayed text value firmly to the current model value, and instead uses a watcher which can react to changes in any prop values that affect the text value, but selectively ignore a change when it is known to be caused by a user input which may not yet be fully complete.

Addresses #1720

### DIFF
--- a/packages/core/src/NumberField/NumberField.test.ts
+++ b/packages/core/src/NumberField/NumberField.test.ts
@@ -298,6 +298,29 @@ describe('numberField', () => {
       await fireEvent.keyDown(input, { key: kbd.ENTER })
       expect(input.value).toBe('EUR 7.00')
     })
+
+    it('should correctly update model value independently of formatting', async () => {
+      const { input, emitted } = setup({
+        defaultValue: 100,
+      })
+
+      // after update, both internal and displayed value are not formatted
+      await fireEvent.update(input, '1000')
+      expect(input.value).toBe('1000')
+      expect(emitted('update:model-value')[0]).toEqual([1000])
+      // update:model-value fired after each consecutive update
+      await fireEvent.update(input, '10000')
+      expect(input.value).toBe('10000')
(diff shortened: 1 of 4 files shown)


zernonia:
@blakemann  the component is design to only format the input when user is done with input. Thus you may see the `update:modelValue` only triggers then.

Your proposal could be a feature prop, maybe we can start from there? 😁

blakemann:
@zernonia the formatting is still only applied when the user is done with the input with these changes - from a user perspective there is no difference in behaviour.

This really just gives a bit more observability into the state of the input to a parent component in the uncommon case that it might need it - but for the most part will not be a noticeable change, which makes me wonder if adding complexity to the component by controlling this with a prop is worthwhile, but can consider it if necessary.

Are there any downsides to having the `update:modelValue` event fire while typing that you can think of, that would warrant this not being the default functionality?

blakemann:
I'm closing this as it sounds like this change is not desired as the default behaviour, and I feel that it adds unnecessary complexity to the component to have it configurable as a prop. Can be re-opened if anyone wants to reconsider in the future.

PR #1744, opened 2025-03-22
fix(Calendar): handle number of months > 2

epr3:
Hello @zernonia 

Just realised this morning that the cell trigger focus calculations were breaking if the `numberOfMonths` was greater than 2. 😅  Sorry about that. This should fix it for good 🙏🏻


zernonia:
Thanks @epr3 !

PR #1750, opened 2025-03-24
fix(Calendar): minValue and maxValue navigation issues

hartbit:
Fixes bugs in `Calendar` and `RangeCalendar` when navigating the focus into months that contain the `minValue` or `maxValue`. I added unit tests that make sure it works for both components with `numberOfMonths=1` and `numberOfMonth=2`.

### DIFF
--- a/packages/core/src/Calendar/CalendarCellTrigger.vue
+++ b/packages/core/src/Calendar/CalendarCellTrigger.vue
@@ -84,7 +84,7 @@ const isFocusedDate = computed(() => {
 const isSelectedDate = computed(() => rootContext.isDateSelected(props.day))
 
 const SELECTOR
-  = '[data-reka-calendar-cell-trigger]:not([data-disabled]):not([data-outside-view]):not([data-outside-visible-view])'
+  = '[data-reka-calendar-cell-trigger]:not([data-outside-view]):not([data-outside-visible-view])'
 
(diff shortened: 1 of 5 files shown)


packages/core/src/Calendar/CalendarCellTrigger.vue
```diff
@@ -85,6 +85,8 @@ const isSelectedDate = computed(() => rootContext.isDateSelected(props.day))
 
 const SELECTOR
   = '[data-reka-calendar-cell-trigger]:not([data-disabled]):not([data-outside-view]):not([data-outside-visible-view])'
+const NOT_OUTSIDE_VIEW_SELECTOR
```
epr3, line 88: @hartbit I think you can replace `SELECTOR` with this altogether

epr3 (changes requested):
Hello @hartbit,

Thank you very much for your contribution! :rocket: 

There's one small thing that needs to be updated on this PR. Other than that, LGTM! 🥇

packages/core/src/Calendar/CalendarCellTrigger.vue
hartbit: Thanks for the suggestion. I made the change, and indeed, it doesn't break any tests.

epr3 (approved):

PR #1751, opened 2025-03-24
fix: correct direction for PageUp/PageDown when slider is vertical inverted

cyyynthia:
Also added relevant tests, made useForwardProps a tad bit more robust when props are added (which I think is only affecting testing QoL), and made Slider able to react to prop changes.

### DIFF
--- a/packages/core/src/Slider/Slider.test.ts
+++ b/packages/core/src/Slider/Slider.test.ts
@@ -52,6 +52,240 @@ describe('given default Slider', () => {
     })
   })
 
+  describe('when inverted', () => {
+    beforeEach(async () => {
+      await wrapper.setProps({ inverted: true })
+    })
+
+    describe('after pressing navigation key', () => {
+      let slider: DOMWrapper<HTMLElement>
+
+      beforeEach(() => {
+        slider = wrapper.find('[role="slider"]')
+      })
+
+      it('arrowRight should decrease by 1', async () => {
+        const currentValue = slider.attributes('aria-valuenow')
+        await slider.trigger('keydown', { key: 'ArrowRight' })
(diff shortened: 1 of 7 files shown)


cyyynthia:
In an *inverted* vertical slider, the min value is at the top and the maximum value at the bottom. The directional arrows correctly reflect this direction change, however the PageUp/Down don't.

zernonia:
It's by design that PageUp increase the value, thus in the case of inverted, min at top, max at bottom, PageUp will cause the thumb to go down.

image

zernonia:
Ahhh you are right @cyyynthia , that is indeed confusing. 
Perhaps it should be updated `Increments/decrements by the step value depending on orientation.`

cyyynthia:
Nonetheless, I do think PageUp and PageDown should match the behavior of ArrowUp and ArrowDown ; especially considering the fact that the APG describes these two by referencing them to ArrowUp and ArrowDown.

Also, the documentation of the keys is unfortunately quite vague ; horizontal sliders have Left/Right arrows inverted but Up/Down are intact, and vertical sliders have Up/Down inverted but not Left/Right. Clearly explaining the cases in which it's inverted might be beneficial for clarity.

cyyynthia:
Well, the typo doesn't come from my changes >:)

I added a bit of clarification for keyboard controls, as discussed.

cyyynthia:
Hi there, are there any updates on this? Either positive or negative, I'd be happy to change things or further discuss anything that might be a blocker for this PR 😄

zernonia (approved):
Yoo @cyyynthia ! Apology for the delayed review 🙏🏼 Yeah after your changes I think it brings much more clarity, and anticipated behavior. Happy to merge this! 😁

PR #1765, opened 2025-03-27
perf(primitive): don't redeclare SELF_CLOSING_TAGS in each instance

J-Michalek:
Probably not a big deal, but it could save some performance along the way.


zernonia (approved):
LGTM! Not sure how much impact would this be, perhaps we can try set up some test for this in future. 
We can merge this for now 😁

PR #1766, opened 2025-03-28
chore/readme: convert 'join our discord' into a link

tuananhlai:
Convert 'Join our discord' into Unovue's Discord invite link since I think a lot of people (including myself) expected it to be that way.


zernonia (approved):
Make sense! Thanks for the PR! 😁

PR #1770, opened 2025-03-29
fix(accordion-content): pass `as` prop down to CollapsibleContent

J-Michalek:
Fix: https://github.com/unovue/reka-ui/issues/1768

I think the problem is just the prop not being passed down to the underlying component.


zernonia (approved):
Thanks for the PR @J-Michalek !

PR #1771, opened 2025-03-29
fix(menu-sub-trigger): allow grace customization

J-Michalek:
Fix: https://github.com/unovue/reka-ui/issues/1729

I am not sure about the naming convention for this prop and would very much appreciate a help in that regard.

Is it perhaps already customizable in some other component so that we can allign the naming/docs convention?

### DIFF
--- a/packages/core/src/Menu/MenuSubTrigger.vue
+++ b/packages/core/src/Menu/MenuSubTrigger.vue
@@ -2,7 +2,13 @@
 import type { MenuItemImplProps } from './MenuItemImpl.vue'
 import type { Side } from './utils'
 
-export interface MenuSubTriggerProps extends MenuItemImplProps {}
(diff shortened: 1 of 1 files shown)


packages/core/src/Menu/MenuSubTrigger.vue
```diff
@@ -2,7 +2,13 @@
 import type { MenuItemImplProps } from './MenuItemImpl.vue'
 import type { Side } from './utils'
 
-export interface MenuSubTriggerProps extends MenuItemImplProps {}
+export interface MenuSubTriggerProps extends MenuItemImplProps {
+  /**
+  * Sets the delay for when the submenu is closed.
```
J-Michalek, line 7: @J-Michalek From my testing this description is not really that accurate as the grace timer seems to target the time it allows the user to move the cursor towards the menu rather than delaying the submenu closing.

This documentation should be improved.

PR #1789, opened 2025-04-01
fix(usePresence): handle undefined animation name in presence hook

rick-hup:
I found that getAnimationName(node.value) returns the string 'undefined', which prevents isPresent from properly updating to false.


zernonia:
Thanks for the PR @rick-hup !

Issue #1793, opened 2025-04-02
[Feature]: Add ignoreDisabledDatesForContiguity as Prop to DateRangePickerRoot and RangeCalendarRoot


JSahrhage:
### Describe the feature

# No Breaking Change

## Affected Components/Files

- [RangeCalendarRoot.vue](https://github.com/unovue/reka-ui/blob/v2/packages/core/src/RangeCalendar/RangeCalendarRoot.vue)
- [DateRangePickerRoot.vue](https://github.com/unovue/reka-ui/blob/v2/packages/core/src/DateRangePicker/DateRangePickerRoot.vue)
- [src/date/comparators.ts](https://github.com/unovue/reka-ui/blob/v2/packages/core/src/date/comparators.ts)
- [useRangeCalendar.ts](https://github.com/unovue/reka-ui/blob/v2/packages/core/src/RangeCalendar/useRangeCalendar.ts)

## Problem

When enforcing a minimum selection (e.g., 3 days), selecting a start date might disable the next 2 days as end dates. Users should still be able to select a valid end date after these disabled days, forming a single range that includes the disabled dates. Currently, disabled dates break range contiguity.

### Introduce a new boolean prop named `ignoreDisabledDatesForContiguity`

- Prop: ignoreDisabledDatesForContiguity
- Type: boolean
- Default: false (to maintain existing behavior)

### When `ignoreDisabledDatesForContiguity = false` (default)

- A disabled date (as determined by `isDateDisabled`) will always be treated as a break in contiguity.
- Nothing Changes

#### And `allowNonContiguousRanges = true` 

- Nothing Changes

#### And `allowNonContiguousRanges = false` 

- Disabled dates within a potential range selection are ignored only for the purpose of checking range contiguity.
- A user can select a range like [Monday, Tuesday(disabled), Wednesday] and it will be treated as a single contiguous selection [Monday, Wednesday].

## Additional Context

This feature provides greater flexibility in configuring date range selection behavior, particularly for scenarios involving holidays, non-working days, or other temporarily unavailable dates within a desired selection period.

### Additional information

- [x] I intend to submit a PR for this feature.
- [x] I have already implemented and/or tested this feature.
- [x] I have submitted a PR for this feature.

PR #1794, opened 2025-04-02
feat(ignoreDisabledDatesForContiguity): 1793 feature add ignore disabled dates for contiguity as prop to DateRangePickerRoot and RangeCalendarRoot

JSahrhage:
#1793

### DIFF
--- a/packages/core/src/RangeCalendar/useRangeCalendar.ts
+++ b/packages/core/src/RangeCalendar/useRangeCalendar.ts
@@ -83,7 +83,7 @@ export function useRangeCalendarState(props: UseRangeCalendarProps) {
       }
     }
 
-    const isValid = props.allowNonContiguousRanges.value || areAllDaysBetweenValid(start, end, props.isDateUnavailable, props.isDateDisabled, props.ignoreDisabledDatesForContiguity)
+    const isValid = props.allowNonContiguousRanges.value || areAllDaysBetweenValid(start, end, props.isDateUnavailable, props.isDateDisabled, props.ignoreDisabledDatesForContiguity.value)
     if (isValid) {
       return {
         start,


epr3:
Hello @JSahrhage ,

I think that this property defeats the purpose of the unavailable dates concept, since the use case outlined by you in the issue is already covered by `isDateUnavailable` and `allowNonContiguousRanges` properties.

The way I see it is that disabled dates shouldn't be focusable/selectable at all. `isDateDisabled` is to be used with complex calendar limits, think upgraded `minValue`/`maxValue` props.

`isDateUnavailable` is more flexible and conceptually covers dates that **may** be selected or included in a range at some based on various conditions.

epr3:
Hello @JSahrhage, 
The highlighted range can be adjusted visually through CSS, as in the end, the `modelValue` that `RangeCalendar` outputs is of the shape `{start: DateValue; end: DateValue }` and it's up to you to do the required data processing to remove the dates you need from the range.

But, I can see that just using `isDateUnavailable` and `isDateDisabled` may become unwieldy if you just want to omit some highlighted dates from the range, or override `isDateUnavailable`/`isDateDisabled` on some days, etc.

Therefore, what I would propose is adding a prop to `CalendarCellTrigger`/`RangeCalendarCellTrigger` named something like `selectable` or `overrideMatcher` or something else (naming things is hard 😅 ) which would short circuit this `if` [here](https://github.com/unovue/reka-ui/blob/7d96922111de9be9781ff6ec73c451672c8e6ef6/packages/core/src/Calendar/CalendarCellTrigger.vue#L90) preventing or allowing the date to be focused/selected regardless of the `isDateUnavailable`/`isDateDisabled`.

What do you think?

epr3:
Hello @JSahrhage ,

It's up to you, as all the commits will get squashed when merging into the live branch.

PR #1796, opened 2025-04-03
fix(Select): add data-placeholder for empty array

xerox0213:
When the select component is set as multiple and the array of selected options is empty, the data-placeholder wasn't set on the SelectTrigger component. This PR should fix this.


zernonia (approved):
Thanks for the PR @xerox0213 ! 🙏🏼

PR #1802, opened 2025-04-06
feat(tree): implement bubbleSelect prop

J-Michalek:
Fix https://github.com/unovue/reka-ui/issues/1458

The `bubbleSelect` prop handles the selection and intermediate state of the parent when it's children are selected.

I would like to tidy up the code a little, but it should be fully working. I would also like to add some tests to capture this behavior and add it to the docs.


zernonia (changes requested):
Thanks for the PR @J-Michalek !!!
Ran into some issue when playing with the feature

You may observe that the parent state persist after deselect all children
There's also a weird case where the grandparent was deselect

https://github.com/user-attachments/assets/fef4ca2b-6648-43fb-b39f-cd770103d3b6


Also it would be awesome to include some tests for this 😁

J-Michalek:
@zernonia Thanks for the feedback! I've looked into this and solved the issue with stale parent items after all children were unchecked:

https://github.com/user-attachments/assets/d0536a82-cd8a-42eb-a7b1-36072c4eba76

I have also added some test cases for `propagateSelect` and `bubbleSelect` props. see https://github.com/unovue/reka-ui/pull/1802/commits/1caa6b039807822f932205969c66a7b4cae3d318

zernonia (approved):
Nicely done @J-Michalek ! LGTM!

PR #1816, opened 2025-04-11
feat(DateRangePicker/RangeCalendar): add `isDateHighlightable` property

JSahrhage:
Feature already discussed in: [#1794](https://github.com/unovue/reka-ui/pull/1794) with @epr3.

### DIFF
--- a/packages/core/src/Calendar/useCalendar.ts
+++ b/packages/core/src/Calendar/useCalendar.ts
@@ -21,7 +21,6 @@ export type UseCalendarProps = {
   pagedNavigation: Ref<boolean>
   isDateDisabled?: Matcher
   isDateUnavailable?: Matcher
-  isDateHighlightable?: Matcher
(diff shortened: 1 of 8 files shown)


packages/core/src/Calendar/useCalendar.ts
```diff
@@ -181,6 +182,12 @@ export function useCalendar(props: UseCalendarProps) {
     return false
   }
 
+  const isDateHighlightable = (date: DateValue) => {
+    if (props.isDateHighlightable?.(date))
+      return true
+    return false
+  }
+
```
epr3, line 190: I would move this to `useRangeCalendar` since it's only useful there.

epr3 (changes requested):
LGTM 🚀  One small change and it's all good. Thank you for your contribution!

epr3 (approved):

Issue #1824, opened 2025-04-14
[Feature]: Reset Splitter


han1548772930:
### Describe the feature

Provide a method to reset

### Additional information

- [ ] I intend to submit a PR for this feature.
- [ ] I have already implemented and/or tested this feature.

zernonia:
Hi @han1548772930 ! What exactly you would want to "reset"? You should be able to use `resize` method exposed by `SplitterPanel` component

han1548772930:
> Hi [@han1548772930](https://github.com/han1548772930) ! What exactly you would want to "reset"? You should be able to use `resize` method exposed by `SplitterPanel` component

Hello, when I change the original size, I hope there is a way to restore my current size to the beginning.
`resize `can only monitor but not change it

han1548772930:
`ResizablePanel.vue`
```vue
 <SplitterPanel data-slot="resizable-panel" v-slot="slotProps" v-bind="forwarded">
    <slot v-bind="slotProps" />
  </SplitterPanel>
```

`use`
```vue
   <SplitterPanel collapsible :collapsed-size="10" :min-size="20" id="splitter-group-1-panel-1"
          v-slot="{ isCollapsed, expand }" class="bg-white border rounded-xl flex items-center justify-center">
          <button v-if="isCollapsed" @click="expand()">Expand</button>
          Panel A
   </SplitterPanel>
```
Thanks @dan-hale #1888

PR #1830, opened 2025-04-16
feat(splitter): expose `resetGroup` method for reseting the group

J-Michalek:
Fix https://github.com/unovue/reka-ui/issues/1824

The issue does not answer all the possibilities the reset could allow to do, here is the list of features that come to my mind but are not implemented in this PR, atleast not yet.

- Reset all nested Splitters when calling the reset method
- Reset a specific panel only (probably by order as a param)


zernonia:
@J-Michalek  as the feature request is bit vague, I would love to hear from the user what exact reset he is looking for. But in general we should be able to control the panel `size` by using `resize` method.

zernonia:
Closing since #1824 has been closed, and panel sizes can be controlled via the existing `resize` method. Thanks for the contribution!

PR #1831, opened 2025-04-17
fix(combobox): search term reset with ComboboxCancel component

J-Michalek:
Fix https://github.com/unovue/reka-ui/issues/1829

From what I've gathered from the docs the `ComboboxCancel` component should only reset the search term but not the `modelValue` of `ComboboxRoot` https://reka-ui.com/docs/components/combobox#cancel

Beneath is a quick summary of how the cancel interaction works now.

User selects Apple option and inputs additional "a" at the end then clicks on ComboboxCancel => the trailing "a" is removed but Apple is kept.


packages/core/src/Combobox/ComboboxRoot.vue
```diff
@@ -22,6 +22,7 @@ type ComboboxRootContext<T> = {
   parentElement: Ref<HTMLElement | undefined>
   resetSearchTermOnSelect: Ref<boolean>
   onResetSearchTerm: EventHookOn
+  tiggerResetSearchTerm: EventHookTrigger
```
zernonia, line 25: let's fix the typo here

```suggestion
  triggerResetSearchTerm: EventHookTrigger
```

zernonia (approved):
Thanks for the PR @J-Michalek ! 😁
1 minor minor comment

packages/core/src/Combobox/ComboboxRoot.vue
J-Michalek: Fixed in https://github.com/unovue/reka-ui/pull/1831/commits/e6b4bf378eac2b504ec84eb1c308f5fad929349e

zernonia:
@J-Michalek  sorry I reverted some of your changes because that's not how the `ComboboxCancel` designed to behave. It should be clearing the input, and not resetting to default searchTerm.

Thanks for the PR nonetheless!

J-Michalek:
@zernonia Thanks for the information; I was unsure about some things, so there may be some mistakes.

PR #1838, opened 2025-04-20
feat(Avatar): improve image loading status handling and add crossOrigin prop

sadeghbarati:
from radix-ui


zernonia:
Thanks @sadeghbarati !!  Can you check the failing test? 😁

packages/core/src/Avatar/utils.ts
```diff
@@ -1,33 +1,73 @@
...
-        const image = new window.Image()
-        loadingStatus.value = 'loading'
-        image.onload = updateStatus('loaded')
-        image.onerror = updateStatus('error')
-        image.src = src
-        if (referrer) {
-          image.referrerPolicy = referrer
-        }
-      }
+    watch([() => src.value, () => referrerPolicy?.value, () => crossOrigin?.value], ([src, referrerPolicy], _, onCleanup) => {
```
zernonia: `onCleanup` in `watch` is 3.5+ feature, I believe this might not a feasible solution for earlier version. Perhaps use `watchEffect`?

sadeghbarati: Hi Zernonia 👋 

Right, will change it

sadeghbarati:
Could you please update the tests 😄

packages/core/src/Avatar/utils.ts
sadeghbarati: Done

sadeghbarati:
Don't understand why the test is failing cause the img element is present (console.log in Avatar utils.ts)

image

zernonia:
No worry @sadeghbarati ! I can help with fixing the failed test. We might need to mock `addEventListener`

PR #1839, opened 2025-04-20
fix(tooltip): close tooltip on pointer down when open

sadeghbarati:
from radix-ui


sadeghbarati:
There is so much going on in radix-ui primitive repo, I could only track `tooltip` and `avatar` changes

You might need to check their new `slot` component changes or other changes

PR #1850, opened 2025-04-24
Update showcase.md

devinschumacher:


zernonia:
Yo @devinschumacher I dont see anywhere reka/radix-vue is being use in the given website 😬

cyyynthia:
Looking at it, it seems the thing is built in React on top of Radix-UI and the original shadcn/ui component catalog rather than anything Vue. :eyes:

PR #1852, opened 2025-04-25
feat(DatePicker): Add portal prop

jack-allocate:
Resolves:
- https://github.com/unovue/reka-ui/issues/1847
- https://github.com/unovue/reka-ui/issues/1851

Ideally PopoverPortal would be opt-in like other components (e.g. Combobox), but that would require a breaking change.

### DIFF
--- a/packages/core/src/DatePicker/DatePickerContent.vue
+++ b/packages/core/src/DatePicker/DatePickerContent.vue
@@ -1,8 +1,11 @@
 <script lang="ts">
-import type { PopoverContentEmits, PopoverContentProps } from '..'
+import type { PopoverContentEmits, PopoverContentProps, PopoverPortalProps } from '..'
+import { computed } from 'vue'
 import { PopoverContent, PopoverPortal, useForwardPropsEmits } from '..'
 
-export interface DatePickerContentProps extends PopoverContentProps {}
+export interface DatePickerContentProps extends PopoverContentProps {
+  portal?: PopoverPortalProps
+}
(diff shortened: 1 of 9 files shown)


packages/core/src/DatePicker/DatePickerContent.vue
```diff
@@ -1,20 +1,27 @@
 <script lang="ts">
-import type { PopoverContentEmits, PopoverContentProps } from '..'
+import type { PopoverContentEmits, PopoverContentProps, PopoverPortalProps } from '..'
+import { computed } from 'vue'
 import { PopoverContent, PopoverPortal, useForwardPropsEmits } from '..'
 
-export interface DatePickerContentProps extends PopoverContentProps {}
+export interface DatePickerContentProps extends PopoverContentProps {
+  portal?: PopoverPortalProps
```
epr3, line 10: Hello @jack-allocate,

Thank you for your contribution! There are a few changes that need to be added:

- Add a comment above the prop for the docs (copy the one from `PopoverPortal`
- Add the same prop to `DateRangePickerContent`
- Run `pnpm docs:gen` to regenerate the docs and commit the result

epr3 (changes requested):

packages/core/src/DatePicker/DatePickerContent.vue
jack-allocate: @epr3 done!

epr3 (approved):
LGTM!

PR #1854, opened 2025-04-26
fix(combobox): async options highlight

### LINKED ISSUE #1853: [Bug]: Combobox item highlighting lost when populating from ajax search
### Environment

```bash
Node version: 22.13.1
Reka UI version: 2.2.0
Vue version: 3.4.27
CSS framework: tailwindcss@4.0.17
Client OS: MacOS 14.2
Browser: Arc 1.91.0
```

https://stackblitz.com/edit/9c4y9ocg?file=src%2FApp.vue

### Steps to reproduce

In the StackBlitz sandbox example, you'll see I've simulated an ajax search that takes a little bit of time to return with results.

- From no results state, type in the letter `a`. You'll notice it will find results, but the first option will not be highlighted yet. But if you continue typing `apple`, it'll then highlight the result that was populated from the previously simulated ajax search (of `a`)...

    Image

- If you type `apple` really fast before the simulated ajax search returns any results, you'll notice the same issue, where it won't be able to highlight a result, if the previous combobox state was empty...

    Image

### Describe the bug

When the Combobox options are all hardcoded, there is no issue. However, when the Combobox options are dynamically computed, it seems Combobox is trying to highlight too soon before the computed options are rendered to DOM.

Just speculating, but maybe Combobox's highlighting logic is running a tick too soon, and vue's `nextTick()` could be used to ensure highlighting happens after new options are rendered?

### Expected behavior

When dynamically populating Combobox options (ie. from search results), I'd expect Combobox to always highlight the first item by default, no matter how long those options take to compute and render.

_No response_

J-Michalek:
Fix: https://github.com/unovue/reka-ui/issues/1853

To my understanding the issue here is that the first item is highlighted right away as a reaction to the input event inside of `ListboxFilter` but if the options are only loaded after the input has been set the first option won't get the highlight. To remedy this I used the `highlightFirstItem` function inside of `ComboboxRoot `which is invoked after the search value or number of items changes.

As a sidenote I removed the event parameter from the `highlightFirstItem` function as it does not use it at all but it was required in the types.

### DIFF
--- a/packages/core/src/Combobox/ComboboxCancel.vue
+++ b/packages/core/src/Combobox/ComboboxCancel.vue
@@ -18,7 +18,7 @@ const rootContext = injectComboboxRootContext()
 
 function handleClick() {
   // Reset the search to show all options.
-  rootContext.filterState.search = ''
+  rootContext.filterSearch.value = ''
 
   if (rootContext.inputElement.value) {
     rootContext.inputElement.value.value = ''
--- a/packages/core/src/Combobox/ComboboxEmpty.vue
+++ b/packages/core/src/Combobox/ComboboxEmpty.vue
@@ -14,7 +14,7 @@ const rootContext = injectComboboxRootContext()
 
 const isRender = computed(() => rootContext.ignoreFilter.value
   ? rootContext.allItems.value.size === 0
-  : !!rootContext.filterState.search && rootContext.filterState.filtered.count === 0,
+  : rootContext.filterState.value.count === 0,
 )
 </script>
 
--- a/packages/core/src/Combobox/ComboboxGroup.vue
+++ b/packages/core/src/Combobox/ComboboxGroup.vue
@@ -22,7 +22,7 @@ const props = defineProps<ComboboxGroupProps>()
 const id = useId(undefined, 'reka-combobox-group')
 const rootContext = injectComboboxRootContext()
 
-const isRender = computed(() => rootContext.ignoreFilter.value ? true : !rootContext.filterState.search ? true : rootContext.filterState.filtered.groups.has(id))
+const isRender = computed(() => rootContext.ignoreFilter.value ? true : !rootContext.filterSearch.value ? true : rootContext.filterState.value.groups.has(id))
 
 const context = provideComboboxGroupContext({
   id,
--- a/packages/core/src/Combobox/ComboboxInput.vue
+++ b/packages/core/src/Combobox/ComboboxInput.vue
@@ -46,7 +46,7 @@ function handleInput(event: InputEvent) {
     nextTick(() => {
       if (target.value) {
         rootContext.filterState.search = target.value
-        listboxContext.highlightFirstItem(event)
+        listboxContext.highlightFirstItem()
       }
     })
   }
--- a/packages/core/src/Combobox/ComboboxInput.vue
+++ b/packages/core/src/Combobox/ComboboxInput.vue
@@ -45,13 +45,13 @@ function handleInput(event: InputEvent) {
     rootContext.onOpenChange(true)
     nextTick(() => {
       if (target.value) {
-        rootContext.filterState.search = target.value
+        rootContext.filterSearch.value = target.value
         listboxContext.highlightFirstItem()
       }
     })
   }
   else {
-    rootContext.filterState.search = target.value
+    rootContext.filterSearch.value = target.value
   }
 }
 
@@ -86,14 +86,9 @@ watch(rootContext.modelValue, async () => {
     resetSearchTerm()
 }, { immediate: true, deep: true })
 
-watch(
-  () => props.modelValue,
-  () => {
-    if (props.modelValue !== undefined) {
-      rootContext.filterState.search = props.modelValue
-    }
-  },
-)
+watch(rootContext.filterState, () => {
+  listboxContext.highlightFirstItem()
+})
 </script>
 
 <template>
--- a/packages/core/src/Combobox/ComboboxInput.vue
+++ b/packages/core/src/Combobox/ComboboxInput.vue
@@ -87,7 +87,10 @@ watch(rootContext.modelValue, async () => {
 }, { immediate: true, deep: true })
 
 watch(rootContext.filterState, () => {
-  listboxContext.highlightFirstItem()
+  // we exclude virtualized list as the state would be constantly updated
+  if (!rootContext.isVirtual.value) {
+    listboxContext.highlightFirstItem()
+  }
 })
 </script>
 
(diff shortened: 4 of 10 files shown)


J-Michalek:
In hindsight I think the usage of the `highlightFirstItem` function should be moved out of `ComboboxRoot` to the `ComboboxInput` as I think this feature is not really wanted when there is no `ComboboxInput` present.

jesseleite:
Hey @J-Michalek, thank you for jumping on this! ❤️‍🔥

I'm just playing with your PR here...

Combining my initial StackBlitz example in #1853 with the changes in this PR in package.json, it seems it is greatly improved. I was able to make it fail, but super rarely. For example, I'm not using arrow keys at all here, and it's almost always auto-selecting just fine...

CleanShot 2025-04-28 at 11 39 41

However, if increase the timeout to 2000ms, even with a more proper debounce function (here's a [StackBlitz example](https://stackblitz.com/edit/9c4y9ocg-7bete5jx?file=src%2FApp.vue) for you), I can make it glitch out pretty easily by changing my search query. I realize 2000ms is overkill, but it seems to highlight the issue more clearly...

CleanShot 2025-04-28 at 11 46 02

Weirdly, it's also rendering `<ComboboxEmpty>`'s "no options" along with the successfully queried option result 🤔

Curious on your thoughts?

J-Michalek:
@jesseleite Yeah this is bad. Seems like the code has several race conditions (since the `ComboboxEmpty` gets stuck there) and probably needs a closer look into the internals. This is definitely an issue with using watchers and/or syncing several values resulting in unpredictable state when one thing changes out of the ordinary - in this case the options.

I will try to look more closely into this but no promises as I think I would need to know more about the `Combobox`'s desired functionality to not break other cases which are expected.

Edit: I looked into this a little more and investigated how NuxtUI uses it as their `SelectMenu` component seems to work as expected with async options and I found out that if you use the `ignoreFilter` prop on `ComboboxRoot` then it starts to work as expected. Seems like the issue is somewhere in the internal filtering logic of the `Combobox`.

jesseleite:
> I found out that if you use the `ignoreFilter` prop on `ComboboxRoot` then it starts to work as expected. Seems like the issue is somewhere in the internal filtering logic of the `Combobox`.

@J-Michalek We're actually using the ignoreFilter prop ourselves, because we're using [fuzzysort](https://github.com/farzher/fuzzysort) as our search/sort algorithm with Combobox. Honestly, it works great, aside from the issue in #1853. I intentionally left ignoreFilter/fuzzysort out of the StackBlitz implementation, because it didn't seem to matter, and I was trying to keep the reproducible example as simple as possible.

I'm curious what you're doing differently that ignoreFilter works for you as expected?

J-Michalek:
@jesseleite Oh I see now, it worked for me because I was using the second example you mentioned [here](https://github.com/unovue/reka-ui/pull/1854#issuecomment-2835718817) which includes the content of this PR and there it seems to work when used together with `ignoreFilter` prop, but without this PR it is broken.

jesseleite:
@J-Michalek Ah whoops I'm sorry, yes that example should've had `ignoreFilter`. I've added that prop to [the StackBlitz example](https://stackblitz.com/edit/9c4y9ocg-7bete5jx?file=src%2FApp.vue), and it definitely resolves the `ComboboxEmpty` issue I mentioned above, but it still has trouble highlighting. For example...

- Type 'apple' quickly and wait for highlighted result
- Hit cmd-backspace to clear input text
- Type 'honey' quickly, and it doesn't highlight first result 🤔

CleanShot 2025-04-29 at 12 09 47

jesseleite:
@J-Michalek Ah whoops I'm sorry, yes that example should've had `ignoreFilter`. I've added that prop to [the StackBlitz example](https://stackblitz.com/edit/9c4y9ocg-7bete5jx?file=src%2FApp.vue), and it definitely resolves the `ComboboxEmpty` issue I mentioned above, but it still has trouble highlighting. For example...

- Type 'apple' quickly and wait for highlighted result
- Hit cmd-backspace to clear input text
- Type 'honey' quickly, and it doesn't highlight first result 🤔

CleanShot 2025-04-29 at 12 09 47

I'm also using your latest PR commit in my StackBlitz's package.json, copied from the [bot post above](https://github.com/unovue/reka-ui/pull/1854#issuecomment-2832150713)...

```
"reka-ui": "https://pkg.pr.new/unovue/reka-ui/reka-ui@966dfeb"
```

J-Michalek:
@jesseleite Yes, I just wanted to clear any possible confusion. I already got something cooking in terms of refactoring and fixes. I will try to push it most likely tomorrow if I am satisfied with the results.

J-Michalek:
@jesseleite I actually got the changes going today, I would very much appreciate your help with testing them. I was not able to break it with the use cases you have described.

The code will need more review but for quick assesment of improvement it should be enough.

jesseleite:
Oh nice! Looks like there's conflicts on this branch. Also, is there a way to kick [pkg.pr.new](https://github.com/unovue/reka-ui/pull/1854#issuecomment-2832150713) CI to build for your latest commit? I'm getting this in StackBlitz...

CleanShot 2025-04-29 at 15 54 47

J-Michalek:
@jesseleite Hello, I think every change should be released but due to the fact that there were merge conflicts the pipeline did not run. I resolved the conflicts and the changes should be available with https://pkg.pr.new/reka-ui@f161dbd (npm i https://pkg.pr.new/reka-ui@f161dbd).

jesseleite:
@J-Michalek Testing with latest commit in this updated [StackBlitz](https://stackblitz.com/edit/9c4y9ocg-ook9vzgu?file=src%2FApp.vue), and I can't get it to fail... Working great! ❤️‍🔥👏

J-Michalek:
I will need to update the PR description to reflect what was actually changed as it is now outdated.

J-Michalek:
I decided not to update the PR description so that the conversaion here can be tracked correctly.

@zernonia I would love a review if you got a minute, there are a lot of changes I think it is because of some formatting rule changes? Not sure if I've done something wrong so please let me know if that is the case. The main changes happened in `Combobox` and `Listbox` components that can be found in the changes.

jesseleite:
Hoping this one can still be looked at sometime! 🙏 ... I haven't tested since conflicts were resolved the last time, but it definitely fixes our problem with [the command palette feature coming in Statamic](https://github.com/statamic/cms/pull/11699) ❤️ 

image

zernonia:
Yo @J-Michalek ! can you perhaps resolves the conflicts? I'll review this soon 😁🙏🏼

zernonia (approved):
Thanks for the PR guys! LGTM!

meirroth:
You guys just made my day!!! 🎉🚀

jesseleite:
Yeah, thank you so much for your work on this one!

PR #1860, opened 2025-04-29
fix: Fix select and modal popup errors in wujie micro-frontend project

Sewar-x:
In the wujie micro-frontend environment, the following error will be reported when the pop-up layer is pulled down：

`utils.ts:16 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'closest')
    at isLayerExist (utils.ts:16:37)
    at HTMLDocument.handleFocus (utils.ts:160:29)
`    

The above error is because the dom cannot be obtained in the micro-frontend environment, and an additional layer of judgment is required

### DIFF
--- a/packages/core/src/DismissableLayer/utils.ts
+++ b/packages/core/src/DismissableLayer/utils.ts
@@ -14,7 +14,7 @@ export const POINTER_DOWN_OUTSIDE = 'dismissableLayer.pointerDownOutside'
 export const FOCUS_OUTSIDE = 'dismissableLayer.focusOutside'
 
 function isLayerExist(layerElement: HTMLElement, targetElement: HTMLElement) {
-  const targetLayer = targetElement.closest(
+  const targetLayer = targetElement?.closest(
     '[data-dismissable-layer]',
   ) as HTMLElement
 
--- a/packages/core/src/DismissableLayer/utils.ts
+++ b/packages/core/src/DismissableLayer/utils.ts
@@ -14,9 +14,9 @@ export const POINTER_DOWN_OUTSIDE = 'dismissableLayer.pointerDownOutside'
 export const FOCUS_OUTSIDE = 'dismissableLayer.focusOutside'
 
 function isLayerExist(layerElement: HTMLElement, targetElement: HTMLElement) {
-  const targetLayer = targetElement?.closest(
+  const targetLayer = targetElement.closest(
     '[data-dismissable-layer]',
-  ) as HTMLElement
+  )
 
   const mainLayer = layerElement.dataset.dismissableLayer === ''
     ? layerElement
@@ -27,10 +27,8 @@ function isLayerExist(layerElement: HTMLElement, targetElement: HTMLElement) {
   const nodeList = Array.from(
     layerElement.ownerDocument.querySelectorAll('[data-dismissable-layer]'),
   )
-  if (
-    (targetLayer
-      && mainLayer === targetLayer)
-    || nodeList.indexOf(mainLayer) < nodeList.indexOf(targetLayer)
+
+  if (targetLayer && (mainLayer === targetLayer || nodeList.indexOf(mainLayer) < nodeList.indexOf(targetLayer))
   ) {
     return true
   }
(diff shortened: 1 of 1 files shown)


packages/core/src/DismissableLayer/utils.ts
```diff
@@ -14,7 +14,7 @@ export const POINTER_DOWN_OUTSIDE = 'dismissableLayer.pointerDownOutside'
 export const FOCUS_OUTSIDE = 'dismissableLayer.focusOutside'
 
 function isLayerExist(layerElement: HTMLElement, targetElement: HTMLElement) {
```
J-Michalek, line 16: If the `targetElement` may not exist then the function should type it as such. Now there is the hidden and undocumented mistery of why someone added the `?` if the property is required.

Sewar-x: Background:
I used the selector of reka-ui in the sub-application of the micro-frontend application (wujie micro-frontend). When I clicked the drop-down selection to focus on the drop-down option, the following error occurred:

```
utils.ts:16 Uncaught (in promise) TypeError: Cannot read properties of null (reading 'closest')
at isLayerExist (utils.ts:16:37)
at HTMLDocument.handleFocus (utils.ts:160:29)
```

The above error does not affect the use of components
But there is no such error in the single-page application of vue.
Reason:
According to the above error, the targetElement parameter of the isLayerExist method is null, the code is as follows:

```
function isLayerExist(layerElement: HTMLElement, targetElement: HTMLElement) {
const targetLayer = targetElement.closest(
'[data-dismissable-layer]',
) as HTMLElement

const mainLayer = layerElement.dataset.dismissableLayer === ''
? layerElement
: layerElement.querySelector(
'[data-dismissable-layer]',
) as HTMLElement

const nodeList = Array.from(
layerElement.ownerDocument.querySelectorAll('[data-dismissable-layer]'),
)
if (
(targetLayer
&& mainLayer === targetLayer)
|| nodeList.indexOf(mainLayer) < nodeList.indexOf(targetLayer)
) {
return true
}
else {
return false
}
}

```
According to the call stack, the function that calls this method is useFocusOutside, in which
if (!element.value || isLayerExist(element.value, event.target as HTMLElement)) is called.
The function source code is as follows:


```
export function useFocusOutside( 
onFocusOutside?: (event: FocusOutsideEvent) => void, 
element?: Ref<HTMLElement | undefined>,
) { 
const ownerDocument: Document 
= element?.value?.ownerDocument ?? globalThis?.document 

const isFocusInsideDOMTree = ref(false) 
watchEffect((cleanupFn) => { 
if (!isClient) 
return 
const handleFocus = async (event: FocusEvent) => { 
if (!element?.value) 
return 

await nextTick() 
if (!element.value || isLayerExist(element.value, event.target as HTMLElement)) 
return 

if (event.target && !isFocusInsideDOMTree.value) { 
const eventDetail = { originalEvent: event } 
handleAndDispatchCustomEvent( FOCUS_OUTSIDE,
onFocusOutside,
eventDetail,
)
}
}


ownerDocument.addEventListener('focusin', handleFocus)

cleanupFn(() => ownerDocument.removeEventListener('focusin', handleFocus))
})

return {
onFocusCapture: () => (isFocusInsideDOMTree.value = true),
onBlurCapture: () => (isFocusInsideDOMTree.value = false),
}
}
```

In the micro-frontend environment, the callback parameter event.target of the handleFocus function is null. Continuing to locate, it is found that the ownerDocument variable in the micro-frontend environment is the root element of the shadow of the sub-application, and the wujie micro-frontend uses the proxy proxy method to proxy the document of the sub-application, and the document api of the sub-application is proxied to the root element of the shadow of the sub-application, so the above useFocusOutside In the method, ownerDocument obtains the root element of the shadow of the sub-application, which causes event.target to be null in the handleFocus callback.
Theoretically, this is a bug in the wujie micro-frontend framework. Others have also mentioned this problem: [https://juejin.cn/post/7392535991972937779](使用wujie(无界)微前端过程中的e.target指向问题修复)

Solution:
In reka-ui, add targetElement.closest to targetElement?.closest in the isLayerExist method to be compatible and avoid errors in the micro-frontend environment

zernonia:
Thanks for the PR @Sewar-x  and comment @J-Michalek ! I've updated abit 😁

PR #1878, opened 2025-05-04
feat(pin-input)!: add support for numeric modelValue

### LINKED ISSUE #1877: [Feature]: Support number type for PinInput
### Describe the feature

This was originally requested in https://github.com/nuxt/ui/issues/3751.

Since this is a breaking change, the release must be planned carefully.

### Additional information

J-Michalek:
Fix: https://github.com/unovue/reka-ui/issues/1877

This introduces a breaking change, the release of this needs to be planned.


zernonia (approved):
Thanks for the PR @J-Michalek ! I've set the default type to be `text` so that it wouldn't incur breaking change, and would only render type `number` if the user pass the `type=number` prop 😁

PR #1887, opened 2025-05-06
fix(Select): handling select multiple required

### LINKED ISSUE #1872: [Bug]: Select with multiple and required won't work with more than one item
### Environment

```bash
Developement/Production OS: macOS 15.4.1 (24E263)
Node version: latest
Reka UI version: latest
Vue version: latest
```

https://github.com/Ericlm/reka-select-multiple-required

### Steps to reproduce

1. Creating a Select with `multiple` and `required` on the `SelectRoot`, within a `form`
2. Add 2-3 `SelectItem`, and a `submit` button
3. When triggering the `submit` : 
- If no item selected : ask to select one 👍
- One item selected : submit 👍
- If more than one item : ask to select one 🐛

### Describe the bug

`multiple` with `required` props on a `SelectRoot` make submitting more than one element impossible.

### Expected behavior

When the `multiple` and `required` props are set on a `SelectRoot`, the value should only be required if there is no item in the array, but should submit when there is at least one.

_No response_

Ericlm:
Adds support for `SelectRoot` which are multiple and required. The `SelectRoot` uses a hidden `BubbleSelect` which have been modified to use the Vue v-model for select elements (and manages multiple or not). 

The `BubbleSelect` component now has a `v-model` instead of a `value` and dispatching an `Event` when the value changes.

Fix #1872

### DIFF
--- a/docs/content/meta/ComboboxRoot.md
+++ b/docs/content/meta/ComboboxRoot.md
@@ -101,6 +101,13 @@
     'required': false
   },
   {
+    'name': 'resetModelValueOnClear',
+    'description': '<p>When <code>true</code> the <code>modelValue</code> will be reset to <code>null</code> (or <code>[]</code> if <code>multiple</code>)</p>\n',
+    'type': 'boolean',
+    'required': false,
+    'default': 'false'
+  },
+  {
     'name': 'resetSearchTermOnBlur',
     'description': '<p>Whether to reset the searchTerm when the Combobox input blurred</p>\n',
     'type': 'boolean',
--- a/packages/core/src/Combobox/ComboboxCancel.vue
+++ b/packages/core/src/Combobox/ComboboxCancel.vue
@@ -23,6 +23,9 @@ function handleClick() {
   if (rootContext.inputElement.value) {
     rootContext.inputElement.value.value = ''
     rootContext.inputElement.value.focus()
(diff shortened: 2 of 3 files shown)


Ericlm:
@zernonia could it be possible to merge this, once you're satisfied with the changes? 🙂

packages/core/src/Select/story/SelectFormMultiple.story.vue
```diff
@@ -0,0 +1,144 @@
...
+  <Story
+    title="Select/Multiple"
+    :layout="{ type: 'single', iframe: false }"
+  >
+    <Variant title="multiple">
+      <div class="flex flex-col items-center">
+        <form
+          class="p-12 flex flex-col items-center"
+          @submit.prevent="handleChange"
+        >
```
zernonia: Yo @Ericlm ! Seems like your changes causes the `change` event not bubble up to the form anymore. You can see that `@change` doesn't trigger here

Ericlm: `@change` is not supposed to be fired on the `<form>` element, according to https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event 🤔.

Ericlm:
I wanted to sync my fork but I clicked on the red button… So all of my work gone to waste 🤦 

I will open a new PR I guess 😅

PR #1888, opened 2025-05-07
feat(splitter): expose collapse expand, and resize on SplitterPannel slot

dan-hale:
Instead of needing create a ref to call expand like this:

```vue
<script setup lang="ts">
const panelRef= useTemplateRef('panelRef')
</script>

<template>
...
     <splitter-panel v-slot="{ isCollapsed }" ref="panelRef" :min-size="25" :collapsed-size="5" collapsible>
           <button v-if="isCollapsed"  @click="googleImagePanelRef?.expand()">Expand</div>
...
</template>
```
Could be refactored into this:

```vue
<script setup lang="ts">
...
</script>

<template>
...
     <splitter-panel v-slot="{ isCollapsed , expand}" :min-size="25" :collapsed-size="5" collapsible>
           <button v-if="isCollapsed"  @click="expand()">Expand</div>
...
</template>
```

### DIFF
--- a/packages/core/src/Splitter/SplitterPanel.vue
+++ b/packages/core/src/Splitter/SplitterPanel.vue
@@ -74,6 +74,12 @@ defineSlots<{
     isCollapsed: typeof isCollapsed.value
     /** Is the panel expanded */
     isExpanded: typeof isExpanded.value
+    /** If panel is `collapsible`, collapse it fully. */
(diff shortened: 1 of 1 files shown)


zernonia (approved):
Nice PR! Thanks @dan-hale 🙏🏼

PR #1890, opened 2025-05-08
fix(Progress): omit `aria-valuetext`/`aria-label` for determinate state

### LINKED ISSUE #1871: [Bug]: ProgressRoot should only set a default value label if it is determinate
### Environment

```bash
Node version: 16.0.0
Package manager: pnpm@8.6.0
Reka UI version: 2.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
Browser: Chrome 90.0.4430.212
```

https://codesandbox.io/p/devbox/nsqlsq

### Steps to reproduce

Use `ProgressRoot` with a nullish value and set `aria-label`.

### Describe the bug

When using an indeterminate `ProgressRoot`, the default `getValueLabel`function is used which always returns an invalid label that overwrites the `aria-label` attribute.

Inspect the button in the codesandbox example and check the accessibility tree. The label of the button reads:

"Save NaN%"

### Expected behavior

The expected label in this codesandbox example is:

"Save in progress"

`aria-valuetext` should not be set by default for an indeterminate progress.

`aria-valuetext` should only really be set if `aria-valuenow` is also set. The default value of `aria-label` should not be the same as `aria-valuetext`, as those can sometimes be read together and supplement one another.

kricsleo:
resolves #1871

This PR omit `aria-valuetext`/`aria-label` for determinate state (`modelValue` is `null` or `undefined`); otherwise these will be displayed as `NaN`.

### DIFF
--- a/docs/content/meta/ProgressRoot.md
+++ b/docs/content/meta/ProgressRoot.md
@@ -17,7 +17,7 @@
   {
     'name': 'getValueLabel',
     'description': '<p>A function to get the accessible label text representing the current value in a human-readable format.</p>\n<p>If not provided, the value label will be read as the numeric value as a percentage of the max value.</p>\n',
-    'type': '((value: number, max: number) => string)',
+    'type': '((value: number | null | undefined, max: number) => string | undefined)',
     'required': false,
     'default': '`${Math.round((value / max) * DEFAULT_MAX)}%`'
   },
@@ -61,6 +61,6 @@
   {
     'name': 'getValueLabel',
     'description': '<p>A function to get the accessible label text representing the current value in a human-readable format.</p>\n<p>If not provided, the value label will be read as the numeric value as a percentage of the max value.</p>\n',
-    'type': '(value: number, max: number) => string'
+    'type': '(value: number | null | undefined, max: number) => string | undefined'
   }
 ]" />
--- a/packages/core/src/Progress/ProgressRoot.vue
+++ b/packages/core/src/Progress/ProgressRoot.vue
@@ -20,7 +20,7 @@ export interface ProgressRootProps extends PrimitiveProps {
    *
    *  If not provided, the value label will be read as the numeric value as a percentage of the max value.
    */
-  getValueLabel?: (value: number, max: number) => string
+  getValueLabel?: (value: number | null | undefined, max: number) => string | undefined
 }
 
 const DEFAULT_MAX = 100
@@ -75,8 +75,8 @@ import { computed, nextTick, watch } from 'vue'
 
 const props = withDefaults(defineProps<ProgressRootProps>(), {
   max: DEFAULT_MAX,
-  getValueLabel: (value: number, max: number) =>
-    `${Math.round((value / max) * DEFAULT_MAX)}%`,
+  getValueLabel: (value: number | null | undefined, max: number) =>
+    isNumber(value) ? `${Math.round((value / max) * DEFAULT_MAX)}%` : undefined,
 })
 
 const emit = defineEmits<ProgressRootEmits>()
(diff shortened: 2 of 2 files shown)


markjaniczak:
@kricsleo `aria-valuetext` only overwrites `aria-label` if there is a value in `aria-valuenow` so I'm not sure the best solution is to change the function signature of `getValueLabel`, instead only call it if the value is a valid number? Is there a valid reason to keep setting `aria-label` as the same value of `aria-valuetext`?

kricsleo:
> aria-valuetext only overwrites aria-label if there is a value in aria-valuenow

According to [MDN: aria-valuetext](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-valuetext#description)
> The aria-valuetext attribute is used with the aria-valuenow attribute, not instead of it, unless that value is not known.

"unless that value is not known" means that when `modelValue` is `null | undefined`, `aria-valuetext` can be used alone to describe the current state without `aria-valuenow`. Therefore, "only call it if the value is a valid number" might be inaccurate. Users can set `aria-valuetext` even without `aria-valuenow`, so we should leave the choice to the user by always calling `getValueLabel`.

> change the function signature of getValueLabel

I mostly "correct" the signature (instead of "change" it) 😂. This function could be called with `number | null | undefined` (because the `modelValue` is `number | null | undefined`). The previous signature didn't reflect this.

> keep setting aria-label as the same value of aria-valuetext?

I don't have a strong opinion on this, but users can always override this default behavior to their preference, either by passing `aria-label="..."` or using `getValueLabel`. Do you have any other suggestions?

markjaniczak:
This page from [W3C](https://www.w3.org/WAI/ARIA/apg/practices/range-related-properties/#usingaria-valuetext) seems more prescriptive:

> When the element's values are contained within a range but those values are not numeric (such as "small", "medium" and "large"), or they are numeric but there is value in communicating more information than just a number, aria-valuetext is used to surface the text value to assistive technologies. **Only use aria-valuetext when aria-valuenow is not sufficiently meaningful for users because using aria-valuetext will prevent assistive technologies from communicating aria-valuenow.**

They go on to provide examples where the label of the element differs from the value in `aria-valuetext`. It seems that on some devices, the value of `aria-valuetext` complements the label of the element so that they're read together.

To me at least it makes sense to omit `aria-label` from this component instead of using the same value as `aria-valuetext`, in case some devices read the value out twice from both attributes.

kricsleo:
Now I get what you mean 👍, I agree that these two shouldn't be the same value (they will be repeated by readers), they should complement each other.

Taking the example from [W3C](https://www.w3.org/WAI/ARIA/apg/practices/range-related-properties/#usingaria-valuetext):
```html
<div class="rail">
  <div id="thumb" role="slider" aria-valuemin="0" aria-valuenow="1" aria-valuemax="3"
    aria-valuetext="low" aria-label="Fan speed" tabindex="0" >
  </div>
  <div class="value"> Off </div>
  <div class="value"> Low </div>
  <div class="value"> Medium </div>
  <div class="value"> High </div>
</div>
```
The `aria-label` should describe the purpose of this progress indicator, and the `aria-valuetext` should describe the current progress.

It seems that `getValueLabel` combines these two fields into one (whether it's set in the `aria-label` or `aria-valuetext`). Omitting one of them might resolve the repetition issue, but it doesn't seem like the best solution for the mixed meanings. Perhaps adding another API, `getValueText(...)`, to separate the meanings would be better? I'm not sure; let's hear what the author thinks. cc @zernonia ❤️

zernonia (approved):
Thanks for the PR @kricsleo  and the discussion @markjaniczak . It make sense and the 2 attribute should complement each other, instead of having the same value. We can have another PR to add a new prop as suggested to populate the `ariaLabel` differently. 

For now this PR is good 😁

PR #1893, opened 2025-05-09
feat: extend AcceptableValue to include `bigint` and `symbol`

Norbiros:
Initially discussed here https://github.com/nuxt/ui/pull/4105


zernonia (changes requested):
I believe comparator for `symbol` will cause issue based on the [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol#description).

But it should be good with `bigint` cause internally we are using [ohash](https://github.com/unjs/ohash/blob/main/src/serialize.ts#L64) for serialize and `isEqual` comparator

Let's remove the type for `symbol` 🙏🏼

PR #1907, opened 2025-05-16
fix(TimeField): clear `modelValue` for incomplete input

### LINKED ISSUE #1904: [Bug]: Value of cleared TimeField
### Environment

```bash
Developement/Production OS: macOS 15.4.1 (24E263)
Node version: 24.0.2
Reka UI version: 2.2.1
Vue version: 3.5.13
CSS framework: tailwindcss@4.1.6
```

https://github.com/Ericlm/reka-time-field

### Steps to reproduce

- Create a `TimeFieldRoot` component with a v-model to a `Time` ref. 

### Describe the bug

When clearing the seconds or minutes, the value doesn't update to be `undefined` but stays the initial value, even after clearing all values.

### Expected behavior

I think it should behave like the input time in html : 
- When clearing any of its part, the value becomes undefined
- The inputs stay independent, and keep the last value

Image

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/1904

This PR sets the `modelValue` to `undefined` for incomplete input (e.g., backspacing to delete any time fields).
This makes the behavior align with the `DateField`.

### DIFF
--- a/packages/core/src/TimeField/TimeFieldRoot.vue
+++ b/packages/core/src/TimeField/TimeFieldRoot.vue
@@ -144,9 +144,12 @@ const convertedModelValue = computed({
     return convertValue(modelValue.value)
   },
   set(newValue) {
-    if (newValue)
+    if (newValue) {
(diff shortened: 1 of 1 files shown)


epr3 (approved):
LGTM! 👍🏻

PR #1909, opened 2025-05-16
fix(Calendar): update typings for CalendarRootProps

epr3:
This PR handles an issue outlined by this PR: #1908 in which the Calendar props aren't being picked up properly by the `docs` script due to the fact that the root props use a type alias.


docs/content/meta/DatePickerRoot.md
```diff
@@ -175,15 +175,15 @@
   },
   {
     'name': 'weekdayFormat',
-    'description': '<p>The format to use for the weekday strings provided via the weekdays slot prop</p>\n',
-    'type': '\'narrow\' | \'short\' | \'long\'',
+    'description': '',
+    'type': 'CalendarRootProps',
```
zernonia: seems like it's not parsing correctly tho 🤔

epr3: Fixed

PR #1920, opened 2025-05-20
feat(Progress): add `getValueText` for `aria-valuetext`

### LINKED ISSUE #1871: [Bug]: ProgressRoot should only set a default value label if it is determinate

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/1871

Context: https://github.com/unovue/reka-ui/pull/1890#issuecomment-2865268280

This PR adds a `getValueText` prop for the `Progress` component to resolve the `aria-valuetext` value.

### DIFF
--- a/docs/content/docs/components/stepper.md
+++ b/docs/content/docs/components/stepper.md
@@ -172,6 +172,50 @@ import { StepperDescription, StepperIndicator, StepperItem, StepperRoot, Stepper
 </template>
 ```
 
+### With controls
+
+You can add additional controls for the stepper using buttons and access the typed component instance using `useTemplateRef`.
+
+```vue line=8
+<script setup lang="ts">
+const stepper = useTemplateRef('stepper')
+</script>
+
+<template>
+  <StepperRoot
+    ref="stepper"
+    :default-value="1"
+  >
+    <StepperItem>
+      <StepperIndicator />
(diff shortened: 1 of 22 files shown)


zernonia:
Thanks for the PR @kricsleo !

PR #1924, opened 2025-05-20
feat(TimeField): support interval steps

### LINKED ISSUE #1834: [Feature]: Add minutes and seconds step into TimeField
### Describe the feature

It would be helpful to add support for minutesStep and secondsStep props to the TimeField component.

Currently, TimeField allows users to input or select any value for minutes and seconds without any constraints. In some applications (e.g., scheduling events, setting timers), it would be useful to restrict available minutes and seconds to certain intervals, like every 5, 10, or 15 minutes.

### Additional information

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/1834

This PR adds `step` to controll the stepping interval for the time fields (`year`/`month`/`day`/`hour`/`minute`/`second`/`millisecond`). Defaults to `1`.


packages/core/src/TimeField/TimeFieldRoot.vue
```diff
@@ -103,6 +120,12 @@ const props = withDefaults(defineProps<TimeFieldRootProps>(), {
   readonly: false,
   placeholder: undefined,
   isDateUnavailable: undefined,
+  yearStep: 1,
+  monthStep: 1,
+  dayStep: 1,
+  hourStep: 1,
+  minuteStep: 1,
+  secondStep: 1,
```
zernonia: I think adding each step for each field is abit excessive. What do you think if we have a `step` prop that contain different key for different field. we can use [defu](https://github.com/unjs/defu) to populate default value for field not provided in the prop.
 
eg:
```ts
type Step = {
  day?: number
  month?: number
  ...
}

const props = defineProps<Props & { step?: Step }>()

const step = computed(() => {
  return defu(props.step, {
    day: 1,
    month: 1
    ...
  })
})
```

kricsleo: Nice idea 👍.
Now updated to an aggregated prop `step`

zernonia (approved):
Yet another awesome PR! 🎉

PR #1929, opened 2025-05-21
fix(RangeCalendar): fixed selected and highlighted when `allowNonContiguousRanges`

Tragio:
Hi there 👋

Currently, the range calendar does not highlight unavailable days, which works perfectly as expected. However, when `allowNonContiguousRanges` is enabled, it is possible to select a date range that includes unavailable days, and those days are not highlighted or shown as selected.

**Use case example:**
Consider a car rental website where pickup and return can happen on specific days, even if those days fall within a larger range of unavailable dates, those days are indeed selected, you just can't select them as pickup or return.

CleanShot 2025-05-21 at 16 19 59

This change supports that scenario by showing those "unavailable" days as highlighted and selected inside a range.

Please let me know if the implementation needs to be refactored or adjusted to better align with your coding standards or conventions.

Thank you for the wonderful work 🚀


epr3 (approved):
LGTM 🚀

PR #1935, opened 2025-05-24
feat(NumberField): support `invertWheelChange` to invert the wheel change behavior

### LINKED ISSUE #1934: [Feature]: Invert wheel behavior property for NumberFieldRoot
### Describe the feature

Currently the wheel behaves in an inverted way where if you scroll up it decreases the amount and scrolling down increases it. It would be cool to see this behavior changed by default and having an "invertWheel" property or something similar to change it

### Additional information

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/1934

~~This PR "corrects" the wheel change behavior:~~

||~~Before~~|~~After~~|
|:-|:-:|:-:|
|~~Scroll down (`deltaY > 0`)~~|~~increase~~|~~decrease~~|
|~~Scroll up (`deltaY < 0`)~~|~~decrease~~|~~increase~~|

~~This is more intuitive, other UI libraries also do this, e.g., [Material UI - TextField](https://mui.com/material-ui/react-text-field/#form-props) and [Ant Design - InputNumber](https://ant.design/components/input-number)|~~

This PR adds support for `invertWheelChange` to invert the wheel change behavior.

### DIFF
--- a/docs/content/meta/NumberFieldRoot.md
+++ b/docs/content/meta/NumberFieldRoot.md
@@ -45,6 +45,12 @@
     'required': false
   },
   {
+    'name': 'invertWheelChange',
+    'description': '<p>When <code>true</code>, inverts the direction of the wheel change.</p>\n',
+    'type': 'boolean',
+    'required': false
+  },
+  {
     'name': 'locale',
     'description': '<p>The locale to use for formatting dates</p>\n',
     'type': 'string',
--- a/packages/core/src/NumberField/NumberField.test.ts
+++ b/packages/core/src/NumberField/NumberField.test.ts
@@ -129,6 +129,23 @@ describe('numberField', () => {
       await fireEvent.wheel(input, {
         deltaY: 100, // Positive value for scrolling down
       })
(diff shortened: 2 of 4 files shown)


zernonia:
I think it's debatable which direction is more intuitive 😂  The 'before' feels right to me tho..

Also, the current implementation is inspired by [react-aria](https://react-spectrum.adobe.com/react-spectrum/NumberField.html#numberfield), hence having the same wheel behavior. 

Perhaps we can have a props to invert the behavior, and allow user to modify it?

kricsleo:
Yeah, it seems diverse in the community...

To me, `"down" (deltaY > 0)` aligns more with `"decrease"`, while `"up" (deltaY < 0)` aligns more with `"increase"`.

Since there's no specification for this, I agreed on adding `invertWheelChange` to satisfy both sides 👍.

As for the default behavior, we could temporarily leave it as is to avoid changes. It can be changed in the future if more users request it.

zernonia (approved):
LGTM! 🎉

PR #1945, opened 2025-05-27
perf(Editable): track pointerdown only in editing mode

### LINKED ISSUE #1944: [Bug]: Editable: usePointerDownOutside unnecessarily triggers on every click
### Environment

```bash
see reproduction
```

https://sshudco7.stackblitz.io

### Steps to reproduce

1. click anywhere but the editable trigger
2. see that usePointerOutside does expensive computations (especially in isLayerExists)

Image
Image

### Describe the bug

I think there is no reason to calculate anything when clicking outside of the component and the editable is not in editing state.

### Expected behavior

`usePointerDownOutside` should be disabled when the editable is not in editing state.

I didn't check, but I could imagine this also applies for `useFocusOutside`.

J-Michalek:
Resolves: https://github.com/unovue/reka-ui/issues/1944

The `usePointerDownOutside` event handler can be expensive because it queries the DOM. The `Editable` component only needs to track the pointerdown event while it is in editing mode.

### DIFF
--- a/packages/core/src/DismissableLayer/utils.ts
+++ b/packages/core/src/DismissableLayer/utils.ts
@@ -1,6 +1,6 @@
-import type { Ref } from 'vue'
+import type { MaybeRefOrGetter, Ref } from 'vue'
 import { isClient } from '@vueuse/shared'
-import { nextTick, ref, watchEffect } from 'vue'
+import { nextTick, ref, toValue, watchEffect } from 'vue'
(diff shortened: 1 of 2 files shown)


zernonia (approved):
LGTM! Let's clean up for the console log ya! 😁

Thanks for the PR!

J-Michalek:
@zernonia Thanks for the review, I removed the `console.log`.

Issue #1897, opened 2025-05-12
[Feature]: Allow copy and pasting into date fields


markjaniczak:
### Describe the feature

The date fields have excellent keyboard accessibility since they're segmented and can be incremented / decremented easily. On the flip side, a user can no longer highlight the whole input and copy the value since there are multiple inputs. This also prevents them from copying a date string from elsewhere and pasting it into the input.

I propose that if one segment in a date field is focused and a user pastes either an ISO date string or a locale date string (of their current locale) that the date field is updated to that date.

### Additional information

- [x] I intend to submit a PR for this feature.
- [ ] I have already implemented and/or tested this feature.

melervand:
I was also looking for a copy-paste functionality, but I think I've found a workaround.

You can add `@copy` and `@paste` listeners, but `Ctrl+C` or `Ctrl+V` won't work because `DatePickerInput` has it's own `@keydown` listener, and that handler calls `preventDefault()` on event.

So you can add your own `@keydown` listener to the `DatePickerInput` element and stop propagation when the user presses `Ctrl+C` or `Ctrl+V`.

```html
<DatePickerInput
	@keydown.capture="onKeydown"
	@copy.capture="onCopy"
	@paste.capture="onPaste"
	:part="item.part">
	{{ item.value }}
</DatePickerInput>
```

```ts
const onKeydown = (event: KeyboardEvent) => {
	if ((event.ctrlKey || event.metaKey) && ['KeyC', 'KeyV'].includes(event.code)) {
		event.stopPropagation()
	}
}

const onCopy = (event: ClipboardEvent) => {
        //set clipboardData
        event.clipboardData?.setData('text/plain', model) //format your model as you want
}

const onPaste = (event: ClipboardEvent) => {
        //access clipboardData
	const value = event.clipboardData?.getData('text/plain') //parse user input and assign to model value
}
```

User just need to move focus on any date segment and press `Ctrl+C/Ctrl+V`

PR #1951, opened 2025-05-28
feat: add copy paste function to date field input

markjaniczak:
Implements #1897.

A few things I'm unsure about...

**Copying the locale string to clipboard**. I used the date field's formatter to get the string from the DateValue because I assumed users are expecting to copy the value they see on the screen.

**Copy, paste and cut macros.** Not sure how reliable these are for every device out there but they seem pretty common. Happy to implement an alternative way to bail out of the preventDefault if someone can suggest one.

**Locale datetime strings with timezone information.** The parsing functions from `@internationalized/date` only handle ISO formats but my guess is that users will want to paste date strings from elsewhere that are in their locale format so I added a fallback to parsing the date using the native Date object. It works pretty well for strings like "May 28 2025" and "28/5/2025" but there's a dozen ways people might provide time zone information.

### DIFF
--- a/packages/core/src/DateField/DateFieldInput.vue
+++ b/packages/core/src/DateField/DateFieldInput.vue
@@ -1,9 +1,9 @@
 <script lang="ts">
 import type { PrimitiveProps } from '@/Primitive'
 import type { SegmentPart } from '@/shared/date'
+import { computed, ref } from 'vue'
 import { Primitive } from '@/Primitive'
 import { useDateField } from '@/shared/date/useDateField'
-import { computed, ref } from 'vue'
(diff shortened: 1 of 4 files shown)


markjaniczak:
After a bit of testing I'm starting to think the copied value should be an ISO string rather than using the locale format. Without a locale aware date parser there will be major differences in how a locale date will be interpreted. For example, I'm in Australia and my locale format is dd/mm/yyyy but the `Date` constructor would interpret that string as mm/dd/yyyy so would reject something like "28/05/2025" or misinterpret "01/01/2025".

The behaviour can always be overridden or extended since the keydown handler is no longer preventing the `copy` and `paste` events.

PR #1954, opened 2025-05-28
fix(ListboxItem): replace `disabled` attribute with `aria-disabled="true"`

radqut:
This PR replaces a `disabled` attribute with `aria-disabled="true"`, similar to how it's implemented in `SelectItem`


zernonia (commented):
Thanks for the fix @MichailShcherbakov ! However I afriad this might cause breaking change in terms of action, user might accidentally click on disabled item.

I'll say let's keep the current `disabled`, we can remove it in another major release

zernonia (changes requested):
https://github.com/unovue/reka-ui/pull/1954#pullrequestreview-2880249584

PR #1958, opened 2025-05-28
fix(ContextMenu): prevent trigger element from being interactive when context menu is opened

### LINKED ISSUE #1923: [Bug]: ContextMenu: pointer-events: auto on trigger elements bypasses global pointer-events: none
### Environment

```bash
Node version: 16.0.0
Package manager: pnpm@8.6.0
Reka UI version: 2.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
Browser: Chrome 90.0.4430.212
```

https://stackblitz.com/edit/e4r1j4vf?file=src%2FApp.vue

### Steps to reproduce

1. Go to the reproduction link
2. Right click on the first ContextMenu trigger zone to open the context menu
3. Left click on one of the trigger zones

### Describe the bug

When a `ContextMenu` is opened, the `<body>` element receives `pointer-events: none` to disable interactions with the rest of the UI. However, each trigger element gets a hardcoded `style="pointer-events: auto"` applied, which overrides this and allows interactions to still occur on the trigger.

This becomes problematic when the trigger element supports both left-click and right-click actions. For example, in a list of cards:

- Left-click navigates to another route
- Right-click opens the context menu

Since the trigger element continues to receive pointer events, users may inadvertently left-click and navigate away, rather than dismissing the context menu as intended.

### Expected behavior

Trigger elements should respect the global pointer-events: none when a context menu is open, to prevent unintended interactions.

_No response_

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/1923

This PR removes the default `pointer-events: auto` from the `ContextMenuTrigger`.

Not sure why it was added in [635b90e](https://github.com/unovue/reka-ui/commit/635b90e72667a6c462e32336593d1d2689ff23ff#diff-03f37eade715fb89953094d4edd7186a915bd873e9a1408eeb723f45893e3f42R96-R111). If I missed anything, please let me know ❤️

### DIFF
--- a/packages/core/src/DismissableLayer/utils.ts
+++ b/packages/core/src/DismissableLayer/utils.ts
@@ -55,7 +55,6 @@ export function useContextMenuOutside(
     = element?.value?.ownerDocument ?? globalThis?.document
 
   const isPointerInsideDOMTree = ref(false)
-  const handleClickRef = ref(() => {})
 
   watchEffect((cleanupFn) => {
     if (!isClient || !toValue(enabled))
--- a/packages/core/src/DismissableLayer/utils.ts
+++ b/packages/core/src/DismissableLayer/utils.ts
@@ -49,7 +49,7 @@ function isLayerExist(layerElement: HTMLElement, targetElement: HTMLElement) {
 export function useContextMenuOutside(
   onContextMenuOutside?: (event: ContextMenuOutsideEvent) => void,
   element?: Ref<HTMLElement | undefined>,
-  enabled: MaybeRefOrGetter<boolean> = true,
+  enabled: MaybeRefOrGetter<boolean> = false,
 ) {
   const ownerDocument: Document
     = element?.value?.ownerDocument ?? globalThis?.document
--- a/packages/core/src/DismissableLayer/utils.ts
+++ b/packages/core/src/DismissableLayer/utils.ts
@@ -6,11 +6,15 @@ import { handleAndDispatchCustomEvent } from '@/shared'
 export type PointerDownOutsideEvent = CustomEvent<{
   originalEvent: PointerEvent
 }>
+export type ContextMenuOutsideEvent = CustomEvent<{
+  originalEvent: MouseEvent
+}>
 export type FocusOutsideEvent = CustomEvent<{ originalEvent: FocusEvent }>
 
 export const DISMISSABLE_LAYER_NAME = 'DismissableLayer'
 export const CONTEXT_UPDATE = 'dismissableLayer.update'
 export const POINTER_DOWN_OUTSIDE = 'dismissableLayer.pointerDownOutside'
+export const CONTEXT_MENU_OUTSIDE = 'dismissableLayer.contextMenuOutside'
 export const FOCUS_OUTSIDE = 'dismissableLayer.focusOutside'
 
 function isLayerExist(layerElement: HTMLElement, targetElement: HTMLElement) {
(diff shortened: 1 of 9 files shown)


zernonia:
Yo @kricsleo ! The reason we add that is to prevent closing of `ContextMenu` and opening a new one, as the right click action still wihtin the context of the same trigger, essentially the core fix for ` flash of closing and opening when interact with trigger` issue

kricsleo:
@zernonia I've re-implemented it to avoid the "flash of closing and opening when interacting with the trigger" without `pointer-events: auto`. 
The main trick here is calculating pointer events if they're happening within the trigger element's rect range; if they are, we consider this a fake "contextmenu" happening on the trigger element. Therefore, the menu doesn't need to close and reopen; it just needs to be "moved".

```ts
// Prevent the default context menu from appearing
event.detail.originalEvent.preventDefault()
// Move the menu to the current pointer position
menuRootContext.triggerPoint.value = {
  x: event.detail.originalEvent.clientX,
  y: event.detail.originalEvent.clientY,
}
```

WDYT? 🤔

NilsEvers:
@kricsleo Just a quick idea: what if we store a reference to the event target and compare it directly, instead of checking against the rect range?

kricsleo:
Probably won't work.

When the menu is open, any click event outside the menu will only be associated with the "mask" (when the trigger element is without `pointer-events: auto`). Because the "mask" obscures all elements except the menu, we don't know which element under the "mask" the user actually wants to click. Therefore, we cannot compare "the currently clicked element" with the element that triggered the menu before. So, I can only confirm the clicked element through this position comparison method.

NilsEvers:
Are you sure there's a mask element? I think pointer events are only being blocked by applying pointer-events: none to the body. As far as I know, there's no overlay — but I could be mistaken.

kricsleo:
Sorry, I'm just making an analogy. There isn't an actual mask element. It's just that the global `pointer-event: none` prevents the user from clicking on any specific element, just like a virtual "mask" covers all elements. Therefore, when a click event is triggered, we can't know which element the user intended to click; we only know the location of the click.

packages/core/src/DismissableLayer/utils.ts
```diff
@@ -37,6 +41,77 @@ function isLayerExist(layerElement: HTMLElement, targetElement: HTMLElement) {
...
 
+/**
+ * Listens for `contextmenu` outside a DOM subtree.
+ * to mimic layer dismissing behaviour present in OS.
+ * Returns props to pass to the node we want to check for outside events.
+ */
+export function useContextMenuOutside(
+  onContextMenuOutside?: (event: ContextMenuOutsideEvent) => void,
+  element?: Ref<HTMLElement | undefined>,
+  enabled: MaybeRefOrGetter<boolean> = true,
```
zernonia: I think it's better to have this default to `false` as this is a specific utility for `ContextMenu`, and we dont need this to be trigger for non-related component 😁

packages/core/src/Menu/utils.ts
```diff
@@ -78,7 +78,7 @@ export function isPointInPolygon(point: Point, polygon: Polygon) {
   return inside
 }
 
-export function isPointerInGraceArea(event: PointerEvent, area?: Polygon) {
+export function isPointerInGraceArea(event: MouseEvent, area?: Polygon) {
```
zernonia: I would like to keep the event as `PointerEvent` 😁 as the function name suggest 'pointer'
same applies to the `isPointerInRect` function below.

kricsleo: I've reverted `isPointerInGraceArea` to use `PointerEvent`, but `isPointerInRect` might need to stay with `MouseEvent` because I'm using it to handle the `contextmenu` event (which is `MouseEvent`, not `PointerEvent`). To align the name with the parameter, I've chosen to rename it to `isMouseInRect` ❤️

packages/core/src/DismissableLayer/utils.ts
kricsleo: Sure 👍

kricsleo:
I have to admit that this PR isn't the "perfect" solution for all situations either:

- Solution 1: Settings trigger element to `pointer-events: auto`
  - ✅: This is the easiest way to fix the "flash of closing and opening."
  - ❌: The trigger element and its children remain interactive (`click`/`hover`/`focus`...) when the menu is opened. This could cause unexpected behavior if the user accidentally interacts with them.

- Solution 2: Removing trigger element's `pointer-events: auto` and handle with the click position (**this PR**)
  - ✅: Also fixes the "flash of closing and opening" while keeping the trigger element non-interactive.
  - ❌: When the trigger element is covered (overlaid) by other elements (e.g., another menu's trigger element)., this hack might unexpectedly still open the outer menu.

- Solution 3: Do nothing (without trigger element's `pointer-events: auto` and without extra logic)
  - ✅: Simplest, and the trigger element is non-interactive when opening the menu.
  - ❌: The menu would "flash of closing and opening" when right-clicking the trigger element multiple times.

I can't say which is better right now... Guys, what are your thoughts? :>

NilsEvers:
Thanks for your effort on this, @kricsleo. I agree that your solution could be problematic in some cases — especially when one trigger area is nested inside another.

Solution 1 introduces a lot of unwanted behavior, as outlined in the original issue, so I’d definitely rule that one  out.

For me, version 3 seems like the best approach. The only downside is that it feels slightly less elegant when right-clicking and opening the context menu multiple times — but I don’t think that’s a significant issue.

NilsEvers:
What do you think @zernonia?

PR #1959, opened 2025-05-28
feat(calendar): add `getWeekNumber` utility function

### LINKED ISSUE #1949: [Feature]: Show week number in Calendar
### Describe the feature

I would like an option to display week numbers in the Calendar component.

Inspiration:

Image

### Additional information

J-Michalek:
Resolves: https://github.com/unovue/reka-ui/issues/1949

Added the `CalendarWeek` and `RangeCalendarWeek` components that calculate the number of the week per given row and month in the case of multiple month calendars.

Calendar:

https://github.com/user-attachments/assets/7fee0c1f-b06f-45ab-b8e2-0d1ffb3eb7d1

RangeCalendar:

https://github.com/user-attachments/assets/bb41c44f-ea7d-418a-9840-e4e077cb662d


epr3:
Hello @J-Michalek :wave:

I have one question regarding the implementation: could you elaborate on the need of a separate component instead of an utility function in `reka-ui/date`? 

In my opinion, the  `CalendarWeek` component doesn't have any special mouse/keyboard interaction requirements and can be any HTML element which could output the result of something like this code (which can reside in the `date` package):
`const computeWeekNumber = (idx:number) => {
     const firstDayOfWeek = firstDayOfGrid.add({ weeks: idx })
      const thursday = firstDayOfWeek.add({ days: 3 })
      const firstDayOfYear = startOfYear(thursday)
      return ((thursday.compare(firstDayOfYear) / 7) | 0) + 1
}`

I am looking forward to your insight on this! 👍🏻

J-Michalek:
I guess I was always thinking about this as a component.

I'm not against turning it into helper functions, but we would need to have two, or the one would need to take additional parameters as the calculation is a bit different for for multiple months. Also how would we access the current calendar's grid from inside of the helper function? I haven't checked the `date` package, so I don't know its ins and outs.

epr3:
You would pass the DateValue to the function itself and have access to the `grid` through the `v-slot`, but I'm thinking that the utility function wouldn't need a specific date since `@internationalized/date` already has a lot of utilities and we can leverage those to get the week number for any date passed as an argument.

I think a function like this would get the week number:
```
import { DayOfWeek, getDayOfWeek, CalendarDate, DateValue } from '@internationalized/date'
/**
 * Returns the locale-specific week number
 */
function getWeekNumber(date: DateValue, locale?: string = 'en-US', firstDayOfWeek?: DayOfWeek): number {
  // Get first day of year
  const firstDayOfYear = new CalendarDate(date.year, 1, 1)

  // Calculate the first day of the first week
  const firstDayOfYearWeekday = getDayOfWeek(firstDayOfYear, locale, firstDayOfWeek)

  // First day of first week
  const firstWeekStart = firstDayOfYear.subtract({ days: firstDayOfYearWeekday })

  // If date is before the first week start
  if (date.compare(firstWeekStart) < 0) {
    // It belongs to the last week of the previous year
    const prevYearDate = new CalendarDate(date.year - 1, 12, 31)
    return getLocaleWeekNumber(prevYearDate, locale, firstDayOfWeek)
  }

  // Calculate days between first week start and our date
  const days = getDaysBetween(firstWeekStart, date) // function from `core/src/date/calendar.ts`

  // Week number is days divided by 7 plus 1
  return Math.floor(days.length / 7) + 1
}
```

J-Michalek:
Yeah, I like that, looks promising, I will try to look into this during this week.

J-Michalek:
Hi @epr3, sorry for the delay, but I finally got to this, I removed the components (`CalendarWeek` and `RangeCalendarWeek`) and added the `getWeekNumber` utility function to `reka-ui/date`, would very much appreciate your feedback!

epr3 (approved):
LGTM 👍🏻

PR #1994, opened 2025-06-17
fix(useDateField): only increment minutes/seconds on `ARROW_UP`/`ARROW_DOWN`

### LINKED ISSUE #1993: [Bug]: TimeField minute value changes when pressing Enter key
### Environment

```bash
Reka UI version: 2.3.1
Vue version: 3.5.16
CSS framework: tailwindcss@4.1.10
Client OS: Windows 11 26100.4351
Browser: Firefox 139.0.4 (64-bit)
```

https://reka-ui.com/docs/components/time-field

### Steps to reproduce

Focus the minute segment in the example at the start of the page, then press the Enter key on the keyboard.

### Describe the bug

The minute value seems to decrement by `step`.

### Expected behavior

I believe the expected behavior is that nothing happens (unless the input is in a form, in which case the form should be submitted without the value of the input changing).

_No response_

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/1993

### DIFF
--- a/packages/core/src/shared/date/useDateField.ts
+++ b/packages/core/src/shared/date/useDateField.ts
@@ -757,7 +757,9 @@ export function useDateField(props: UseDateFieldProps) {
 
     const prevValue = props.segmentValues.value.minute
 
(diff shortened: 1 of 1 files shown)


epr3 (approved):
LGTM! :rocket:

PR #2001, opened 2025-06-23
fix: export constant component types

pimlie:
The builds provide a `reka-ui/constant` export but at the moment those give an error in TS because `dist/constant.d.ts` only has an empty export.

This was caused due to the `constant/` folder being outside the `src/` folder which was the only folder being included in the TS build config.

The `as const`'s are ensuring that the TS exports do not fallback to `string[]`

--edit--

Due to build error also simplified the nuxt module and remove the `flat()` and function  usage. Flat causes a new array to be created which just isn't needed if we use a for loop as in this pr.

With these changes it's also possible to support a per component configuration. Currently you can only add all components in a group, but extending `ModuleOptions.components` with the following type would allow users to toggle individual components. Let me know if you want me to add that.
```typescript
// Toggle components by group
type ComponentGroupConfig = {
  [K in keyof Components]?: boolean
}

// Toggle components invidually (still nested per group)
type ComponentConfig = {
  [K in keyof Components]?: {
    [K2 in Components[K][number]]: boolean
  }
}

export interface ModuleOptions {
  components: ComponentConfig | ComponentGroupConfig | boolean
  prefix: string
}
```

--edit--
The two config types can be combined for even more flexiblity ofc
```typescript
type ComponentConfig = {
  [K in keyof Components]?: boolean | {
    [K2 in Components[K][number]]: boolean
  }
}

export interface ModuleOptions {
  components: ComponentConfig | boolean
  prefix: string
}
```

### DIFF
--- a/packages/plugins/src/resolver/index.ts
+++ b/packages/plugins/src/resolver/index.ts
@@ -1,3 +1,4 @@
+import type { Components } from 'reka-ui/constant'
 import type { ComponentResolver } from 'unplugin-vue-components'
 import { components } from 'reka-ui/constant'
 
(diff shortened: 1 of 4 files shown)


packages/plugins/src/resolver/index.ts
```diff
@@ -18,10 +19,14 @@ export default function (options: ResolverOptions = {}): ComponentResolver {
     resolve: (name: string) => {
       if (name.toLowerCase().startsWith(prefix.toLowerCase())) {
         const componentName = name.substring(prefix.length)
-        if (Object.values(components).flat().includes(componentName)) {
-          return {
-            name: componentName,
-            from: 'reka-ui',
+        let groupName: keyof Components
+        for (groupName in components) {
+          const groupComponents: readonly string[] = components[groupName]
```
pimlie, line 24: FYI: casting here to `string[]` is the actual fix for the previous build error.

zernonia:
LGTM @pimlie ! As for your suggestion above, perhaps we can have another ticket for it. 😁

PR #2012, opened 2025-06-26
Bug/issue 2011 trap focus fix

John-Nik:
Fixing the bug of [this opened issue](https://github.com/unovue/reka-ui/issues/2011)


zernonia:
Thanks for the PR @John-Nik . However I realized the docs is wrong, and shouldn't expose `trapFocus` props, but instead user should use `modal` prop to achieve the desired trap/non-trap behavior

John-Nik:
@zernonia Glad I was of help

PR #2017, opened 2025-06-28
refactor(VisuallyHiddenInput)!: remove curly brackets in prop name format

### LINKED ISSUE #1861: [Bug]: ComboBox name attribute when in form and multiple become [name][index]
### Environment

```bash
Node version: 16.0.0
Package manager: pnpm@8.6.0
Reka UI version: 2.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
Browser: Chrome 90.0.4430.212
```

https://codesandbox.io/p/devbox/reka-ui-combobox-forked-kfj4vw?file=%2Fsrc%2FApp.vue%3A96%2C16

### Steps to reproduce

https://kfj4vw-5173.csb.app/

When comboBox is in form with a name attribute and a multiple select type, the name becomes in a strange format.

### Describe the bug

When comboBox is in form with a name attribute and a multiple select type, the name becomes in a strange format. In the example:

```
<ComboboxRoot name="users" multiple class="relative">
```

When inspecting in devtools the selected elements, the results inthe  hidden input element are:

```
<input data-hidden="" tabindex="-1" name="[users][0]" value="Lettuce" style="position: absolute; border: 0px; width: 1px; height: 1px; padding: 0px; margin: -1px; overflow: hidden; clip: rect(0px, 0px, 0px, 0px); clip-path: inset(50%); white-space: nowrap; overflow-wrap: normal;">
```

### Expected behavior

The unusual name is [users][0], which the request doesn't understand. 
Is that a bug or expected behavior? The right name must be `users[0]`, as when multiple false is `users`.
Many thanks for considering my request.

_No response_

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/1861

Not sure if the previous format (`[names][0]`) is intended, drafting a PR here to request a change to the format without wrapping curly brackets for reference. It's ok to close this PR if we decide not to change the format. ❤️

❗❗This might be a breaking change for users who rely on the format like `[names][0]` instead of `names[0]`.

### DIFF
--- a/packages/core/src/Checkbox/Checkbox.test.ts
+++ b/packages/core/src/Checkbox/Checkbox.test.ts
@@ -260,7 +260,7 @@ describe('given checkboxGroup in a form', async () => {
       const submittedData = Object.fromEntries(formData as any)
 
       expect(handleSubmit).toHaveBeenCalledTimes(1)
-      expect(submittedData).toStrictEqual({ '[test][0][name]': 'jack' })
+      expect(submittedData).toStrictEqual({ 'test[0][name]': 'jack' })
     })
   })
 
--- a/packages/core/src/Slider/Slider.test.ts
+++ b/packages/core/src/Slider/Slider.test.ts
@@ -416,7 +416,7 @@ describe('given slider in a form', async () => {
 
     it('should trigger submit once', () => {
       expect(handleSubmit).toHaveBeenCalledTimes(1)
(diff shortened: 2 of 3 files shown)


zernonia:
I think this might be a mistake previously with the square bracket. Initially it was inspired from headless-ui method, but seems like they are rendering `names[0]` (without the square bracket) as well 

Indeed this is a breaking change, probably could release this in minor version 🤞🏻

PR #2023, opened 2025-07-03
fix(PinInput): provide better autocomplete for `type` field

Teages:
Related https://github.com/nuxt/ui/pull/4346

[As my comment said](https://github.com/nuxt/ui/pull/4346#issuecomment-3032770524), user can only get `'text'` when they are adding the `type` field:

```vue
<UPinInput type="" />
<!--             ^ cursor here, only get 'text' -->
```

This PR fixed the issue by postponing the application of default value, and make `number` also appear in the autocomplete hint. It also makes the type of context more robust.

### DIFF
--- a/packages/core/src/PinInput/PinInputInput.vue
+++ b/packages/core/src/PinInput/PinInputInput.vue
@@ -1,5 +1,5 @@
 <script lang="ts">
-import type { PinInputValue } from './PinInputRoot.vue'
+import type { PinInputContextValue } from './PinInputRoot.vue'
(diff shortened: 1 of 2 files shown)


zernonia:
Thanks for the PR @Teages ! Seems like there's a type check error. Can you check that?

Teages:
fixed, let's use `PinInputContextValue` for context value

zernonia (approved):
LGTM! Thanks for the PR @Teages ! 🎉

PR #2032, opened 2025-07-07
feat(RangeCalendar): make raw-value default v-model for RangeCalendar

### LINKED ISSUE #1820: [Feature]: access to the currently selected state of RangeCalendar
### Describe the feature

Currently, there does not seem to be a way to get direct access to the currently selected dates from `RangeCalendar`.

This is because the `v-mode`l seems to represent the last valid `dateRange` instead of the current selection.
This means it does not reflect when a user deselects the range or only selects a start date — instead, it shows an outdated selection.

---

I'm opening this issue because I'm not quite sure what the best solution would be.

The easiest workaround would be to add `update:endValue`.
This would make it possible to track the changes manually, but it would require quite a bit of code for the user.

Another option would be to add a `v-model:rawDateRange` which represents the selected `dateRange`, including undefined values.

Alternatively, the current v-model could adopt the behavior of `v-model:rawDateRange`, but that would be a breaking change — so probably not the best option.

### Additional information

ToBinio:
This adds a `rawModelValue` to the RangeCalendar representing the currently selected dates, including undefined values.

resolves #1820

### DIFF
--- a/docs/content/meta/RangeCalendarRoot.md
+++ b/docs/content/meta/RangeCalendarRoot.md
@@ -168,6 +168,13 @@
     'required': false
   },
   {
+    'name': 'rawModelValue',
+    'description': '<p>The controlled unchecked state of the calendar.</p>\n',
+    'type': 'DateRange',
+    'required': false,
+    'default': '{ start: undefined, end: undefined }'
+  },
+  {
     'name': 'readonly',
     'description': '<p>Whether or not the calendar is readonly</p>\n',
     'type': 'boolean',
@@ -200,6 +207,11 @@
     'name': 'update:placeholder',
     'description': '<p>Event handler called whenever the placeholder value changes</p>\n',
     'type': '[date: DateValue]'
+  },
+  {
+    'name': 'update:rawModelValue',
+    'description': '<p>Event handler called whenever the rawModel value changes</p>\n',
+    'type': '[date: DateRange]'
   },
   {
     'name': 'update:startValue',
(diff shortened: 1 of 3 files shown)


epr3:
Hello @ToBinio,

Could you try adding this property to both `DateRangeField` and `DateRangePicker`?

ToBinio:
Hi @epr3 

I don't think this is necessary for either, since their behavior of the model already matches the behavior of `rawModelValue`

epr3:
Hi @ToBinio,

Oh, right, then, to achieve consistency, could you update the logic for the `modelValue` instead of adding the new property? 

It would bring the date range components in line in terms of `modelValue` expectations.

Thank you!

ToBinio:
> Oh, right, then, to achieve consistency, could you update the logic for the modelValue instead of adding the new property?

One problem with this, I just realized, is that the calendar has the option to cancel the selection with `ESC` while only a start date is selected. 
How should this be handled?

* Have `modelValue` represent the WIP select of the user and keep an internal "lastValidModel" for reverting?
* Have a setting for the Calendar to switch between a strict (current) mode and a soft mode where the model represents the current selection and no undo existing? - this would also keep the change non-breaking

or something else

note: thoughts on having the WIP selection auto undo if the calendar loses focus?

epr3:
I'm thinking the first option with the `lastValidModel` would make the most sense, also being tied in the escape key functionality.

ToBinio:
I have now made the `v-model` have the `rawModel` behavior - I also added an emit for if there is a new valid `dateRange` to make migration easier

just to note: the behavior of `dateRangePicker` also slightly changed since now, also during selection the model changes

epr3 (approved):
LGTM 👍🏻

PR #2038, opened 2025-07-10
feat(splitter): add CSP nonce support

### LINKED ISSUE #2003: [Bug]: Splitter component violates CSP when applying resize cursor styles
### Environment

```bash
Reka UI version: 2.3.0
Vue version: 3.5.16
CSS framework: tailwindcss@4.1.8
```

https://stackblitz.com/edit/vitejs-vite-v5mhft3v

### Steps to reproduce
   ```bash
   npm run build
   npm run preview
```
_Have a bug with stackblitz, where vite build command refused to exit, so you should kill it after build is finished, then run the preview_ 

Move your mouse over the `<SplitterResizeHandle />` component
Check the browser console - the following CSP error is thrown:
   ```
   Refused to apply inline style because it violates the following Content Security Policy directive: "style-src 'self'". Either the 'unsafe-inline' keyword, a hash ('sha256-XeWHGX/2UKasMCuIJT1X9NeUEOH3CHo3Yf9vc3vYuB0='), or a nonce ('nonce-...') is required to enable inline execution.
```

### Describe the bug

When hovering over `<SplitterResizeHandle />`, the component attempts to dynamically inject inline styles to change the cursor, but this is blocked by CSP restrictions.
The issue occurs in the `setGlobalCursorStyle()` function which creates a `<style>` element and sets its `innerHTML` to apply cursor styles globally.

### Expected behavior

The cursor should change to indicate resize capability when hovering over the splitter 

I looked into the work done with the `<Viewport />` component and its nonce property #1097  - it might be interesting to follow the same approach. Alternatively, providing the ability to modify cursor behavior through global CSS classes could be another solution.

Thanks a lot for your amazing work!

Y0me:
Adds Content Security Policy (CSP) nonce support to the Splitter component's resize handles for dynamic cursor styles.

Fixes #2003

### DIFF
--- a/packages/core/src/Splitter/SplitterResizeHandle.vue
+++ b/packages/core/src/Splitter/SplitterResizeHandle.vue
@@ -1,6 +1,7 @@
 <script lang="ts">
 import type { PrimitiveProps } from '@/Primitive'
 import { ref, toRefs, watch, watchEffect } from 'vue'
+import { useNonce } from '@/shared/useNonce'
 import { useWindowSplitterResizeHandlerBehavior } from './utils/composables/useWindowSplitterBehavior'
 
 export interface SplitterResizeHandleProps extends PrimitiveProps {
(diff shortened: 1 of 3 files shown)


zernonia:
LGTM!

PR #2040, opened 2025-07-11
fix(Tree): issue selection item with v-model

rdjanuar:
This change addresses a problem where checkboxes within the tree component are not accurately reflecting their checked state when using v-model.

Problem
When v-model is applied to the tree checkboxes and an attempt is made to toggle them, v-model updates incorrect data. For example, if a parent tree node is already checked and then an attempt is made to uncheck it, v-model appends the data instead of removing it.

Current Behavior
The checkboxes function correctly and display their state accurately only when v-model is not used.


zernonia:
@rdjanuar  can you share reproduction of the issue mentioned above?

rdjanuar:
https://stackblitz.com/edit/nuxt-starter-c3dbfhmc?file=nuxt.config.ts,app.vue

@zernonia

zernonia:
Thanks for the PR @rdjanuar ! 🙏🏼

PR #2045, opened 2025-07-12
fix(TagsInput): korean IME duplicate tags with nextTick in onCompositionEnd

### LINKED ISSUE #2024: [Bug]: Unexpected Duplicate Tag Creation for Korean Input in ⁠`TagsInput` Component
### Environment

```bash
Developement/Production OS: MacOS Sequoia 15.5
Node version: 20.18.1
Package manager: pnpm@10.12.4
Reka UI version: 2.3.2
Vue version: 3
Nuxt version: 3
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@4
Client OS: MacOS Sequoia 15.5
Browser: Chrome 138.0.7204.93
```

https://reka-ui.com/docs/components/tags-input#tags-input

### Steps to reproduce

1. Focus on the ⁠`TagsInput` component input field.
2. Switch the input method to Korean using an IME.
3. Enter "한글" and press the "Enter" key.
4. Observe the creation of two separate tags: "한글" and "글".

### Describe the bug

When inputting Korean text using an IME (Input Method Editor) in the ⁠`TagsInput` component, pressing the "Enter" key results in the creation of duplicate tags. For example, typing "한글" and hitting "Enter" generates two tags: "한글" and "글" instead of one, which is unexpected behavior. This issue likely stems from the mishandling of composition events associated with IME input, which causes the input event to trigger multiple times during an incomplete composition state.

Proposed Solution:
Implement logic to track the composition state and use this to determine when to update the tags array. Ensure that only when an IME is not active (⁠isComposing = false) should the input be processed to add tags.

### Expected behavior

Only one tag should be created for the full Korean text input after pressing "Enter".

Handling IME composition events appropriately will prevent the unintended multiple triggering of input events, thus eliminating unintended tag duplication for non-Latin scripts such as Korean, Japanese, and Chinese. This ensures better user experience across diverse language inputs.

https://github.com/user-attachments/assets/29872b4a-ac65-4aeb-8453-b3ee08f6084f

jinyongp:
Replaced `requestAnimationFrame` in `onCompositionEnd` with Vue's `nextTick` to safely delay setting `isComposing.value = false` until after DOM updates and event queues are processed.

resolve #2024

### DIFF
--- a/packages/core/src/Listbox/ListboxRoot.vue
+++ b/packages/core/src/Listbox/ListboxRoot.vue
@@ -227,7 +227,7 @@ function onCompositionStart() {
   isComposing.value = true
 }
 function onCompositionEnd() {
-  requestAnimationFrame(() => {
+  nextTick(() => {
     isComposing.value = false
   })
 }
--- a/packages/core/src/TagsInput/TagsInputInput.vue
+++ b/packages/core/src/TagsInput/TagsInputInput.vue
@@ -51,7 +51,7 @@ function onCompositionStart() {
   isComposing.value = true
 }
 function onCompositionEnd() {
-  requestAnimationFrame(() => {
+  nextTick(() => {
     isComposing.value = false
   })
 }


zernonia:
Yo @jinyongp ! Thanks for the issue and PR! I wonder why `requestAnimationFrame` doesn't work for Korean. But tested your solution and it seems to work for other language as well. So LGTM!

Could you also update `onCompositionEnd` in`ListboxRoot`? 😁

Issue #2043, opened 2025-07-11
Vue Language Services v3 and typing issues


cyyynthia:
### The situation
The Vue.js recently released a new major version of the language tools, that brought many improvements across the board and cool new features.

I tried upgrading Reka (with the intent of opening a PR) to the new version of vue-tsc, but that has proven to be quite challenging, because there are many issues that are being surfaced that weren't before. In the handful of instances I've looked at, it seems that there are quite a lot of places where the codebase performs not-so-safe type casts, and made some assumptions along the way that didn't raise any alarms previously, but are being problematic now (e.g. the fact the arg in `:ref="(e) => ..."` is always nullable).

### What needs to be improved
Overall I think there is a lot that can be improved by being a bit more rigorous with the way things are dealt with within Reka. Along with fixing the new errors, I think it'd be very good for Reka to go all-in on robustness, perhaps enabling the various [strict features of the Vue compiler](https://github.com/vuejs/language-tools/blob/1d687065e8961ff8b35ded917fe16769c0efb6f5/extensions/vscode/schemas/vue-tsconfig.schema.json#L43-L82) to help detect and correct all the defects in the current way things are done.

### Additional considerations
The strict features heavily rely on `fallthroughAttributes` being enabled. As Reka is pre-built, it needs to be enabled within the build process of the project for downstream users to be able to enable it within their own projects. It's not that simple for Reka, specifically because of the `Primitive` component that'd require advanced type declarations in order for the language tools to identify the relevant fallthrough attributes.

I tried simply enabling the compiler option, but it completely broke the build (it could no longer properly rollup the types). It is also incompatible with `skipTemplateCodegen`; it already isn't enabled in Reka and it must stay so, despite the potential savings it could provide on the type output declarations.

### Long term goal
I think it's also possible to enforce an even higher standard of quality by configuring eslint more. Reducing the usage of `any` to a bare minimum as well as unsafe casts can be helped via typescript-eslint, and Vue's own eslint plugin has features that can be enabled to enforce a more consistent code-style.

I think all of the things I've mentioned above are very important for the long-term success of Reka, as it is becoming a pillar of the Vue ecosystem with many people relying on it daily, directly or indirectly such as via Nuxt. Personally, I've sometimes been unpleasantly surprised by some of the type casts I've seen in Reka, and VLS surfacing a whole bunch of problems seems like a good occasion to bring this whole thing up.

[...]

PR #2046, opened 2025-07-12
build: use tsdown, improve dx

cyyynthia:
This PR proposes upgrading the build setup to use tsdown (which builds significantly faster; around 20s for the full esm+cjs build on my laptop) and additional changes targeted at improving the DX of lib users.

- `tsdown` reduces CI build time from 3m to 1m. Pretty cool if you ask me.
- `fallthroughAttributes` is enabled. It's helpful for users who want to enable Vue's strict typechecking features although currently very limited because of the `Primitive` component, which can't be statically analysed by Volar and will require writing a more advanced type for it to be truly useful.
  - **edit**: currently does not have an effect because of a bug in tsdown. PR sent upstream to correct that
- Declaration maps have been enabled. This should allow editors to jump to the source definition rather than the bundled .d.ts file, which isn't super convenient to navigate through.
- The source code (src folder) is now included in the package. This may sound weird, but actually saves on space since they are no longer embedded twice in the esm and cjs sourcemaps. *It can also be a starting point for the an ecosystem-wide discussion about whether we should publish Vue libraries unbundled or not...*
  - The package is actually the exact same size (in fact a liiiiiiittle bit smaller) despite shipping additional mappings, comments from Rolldown, and ~500kB of extra declaration files. See [this PR](https://pkg-size.dev/https:%2F%2Fpkg.pr.new%2Freka-ui@2046) vs [latest](https://pkg-size.dev/reka-ui) on pkg-size
  - **edit**: expect a 1MB increase once fallthroughAttributes is actually working
- Configurations have been streamlined, namely the tsconfigs and the Vitest config. Less duplication overall :)
- Upgraded some dev dependencies to their latest versions, specifically those related to building and testing.

Related: #2043


cyyynthia:
It seems that currently the Vue compiler options aren't being passed properly within tsdown. I'll look what's up with that upstream (edit: filed https://github.com/sxzz/rolldown-plugin-dts/pull/60), but when fallthroughAttributes is actually enabled, the build time is slower by a factor of two and the declarations larger by a factor of two.

Vue language tools aggressively unroll the properties so there are many, many duplicates in the generated declaration. Nothing much we can do besides hope the tooling will improve on that 🤞🏻

cyyynthia:
As fallthroughAttributes affect documentation generation, and to make merging this easier, I've opted to not enable it just yet in the PR, but acknowledge its existence with a TODO comment.

I also adjusted the documentation generation script to pick the correct tsconfig (and tweaked minor things to get ts to shut)

zernonia:
Thanks for yet another amazing improvement @cyyynthia ! 🎉

I understand your concern regarding the type cast, and the robustness of the package, this is something I wish to slowly improve as well. However similar as you, I had limited time and resources atm which makes it difficult to spend time refactor, clean up and keep up with all the changes in Vue language tool or Rolldown all the time. 

Really appreciate your dedication and contributions to make this repo better!🙏🏼💚

PR #2050, opened 2025-07-13
feat: add openOnFocus and openOnClick options to combobox

### LINKED ISSUE #1702: [Feature]: Combobox allow showing options on focus
### Describe the feature

I want to use the combobox as a searchable select. To accomplish this, I want the options to show when we focus on the input.

While we can control open with `v-model:open`, [this line](https://github.com/unovue/reka-ui/blob/1f685625029187943ec456764756394e2d399520/packages/core/src/Combobox/ComboboxRoot.vue#L121) makes it impossible to do so on focus of the input. If we set `open = true` on focus, we can an infinite loop.

I'm willing to submit a PR for this, but am seeking direction on what kind of solution we want. My first instinct is to add a prop to get around the `.focus()` call, but open to other ideas.

### Additional information

### LINKED ISSUE #1917: [Feature]: Option to open combobox content on focus of input
### Describe the feature

Description of the combobox component from [W3C](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/):

> The popup is hidden by default, i.e., the default state is collapsed. The conditions that trigger expansion -- display of the popup --are specific to each implementation. Some possible conditions that trigger expansion include:
>
> - It is displayed when the combobox receives focus even if the combobox is editable and empty.

I think this feature would be helpful as an option when the combobox is in single selection mode since without the presence of the popup it feels like the combobox allows any text value. This seems common in some design systems, [Atlassian](https://atlassian.design/components/select/radio-select/examples) for example.

### Additional information

markjaniczak:
Adds the feature described in #1917 and #1702.

Added both `openOnFocus` and `openOnClick` because I found that while `openOnFocus` gets the job done, it may be annoying for keyboard users since the content popover will be immediately announced on focus. `openOnClick` gives mostly the same feature but doesn't change the behaviour for keyboard users.


zernonia (approved):
Thanks for the PR @markjaniczak ! 🚀

meirroth:
Hi @zernonia 👋 I opened a new PR to fix a minor misalignment with the blur functionality https://github.com/unovue/reka-ui/pull/2074

PR #2055, opened 2025-07-14
fix(DatePicker): allow using `initialFocus` to control the calendar's auto-focus when opened

### LINKED ISSUE #2027: [Bug]: <DateRangePickerContent> does not emit openAutoFocus (so you can’t prevent its auto-focus)
### Environment

```bash
Developement/Production OS: MacOS 15.5
Node version: v22.15.0
Package manager: npm@10.9.2
Reka UI version: 2.3.2
Vue version: 3.5.16
Nuxt version: 3.17.5
Nuxt mode: universal
Nuxt target: server
Client OS: MacOS 15.5
Browser: Chrome 90.0.4430.212
```

https://stackblitz.com/edit/nflasqpz?file=src%2FApp.vue

### Steps to reproduce

Click on any of the date inputs;
Focus is grabbed by the popover.

### Describe the bug

When opening `<DateRangePickerContent>` via controlled isOpen, built-in focus management steals focus into the calendar panel—but the component never fires an `openAutoFocus` event that consumers can intercept. As a result, there is no way to cancel the default focus behavior (e.g. to keep focus on the triggering input).

### Expected behavior

`openAutoFocus` is fired and interceptable.

_No response_

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/2027

### DIFF
--- a/packages/core/src/DatePicker/DatePickerCalendar.vue
+++ b/packages/core/src/DatePicker/DatePickerCalendar.vue
@@ -30,7 +30,6 @@ const rootContext = injectDatePickerRootContext()
     }"
     :model-value="rootContext.modelValue.value"
     :placeholder="rootContext.placeholder.value"
-    :initial-focus="rootContext.initialFocus.value"
     :multiple="false"
     @update:model-value="(date: DateValue | undefined) => {
       if (date && rootContext.modelValue.value && isEqualDay(date, rootContext.modelValue.value)) return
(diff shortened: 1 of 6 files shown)


zernonia:
Yo @kricsleo ! I think it's better to have a consistent behavior to other popover, perhaps making ` <DatePickerContent @open-auto-focus.prevent>` works as intended?

zernonia:
Perhaps something like this?

1. Remove `initial-focus` always true props

2. Update `@open-auto-focus` event to highlight calendar ourself
```html
  <PopoverPortal v-bind="portal">
    <PopoverContent
      v-bind="{ ...forwarded, ...$attrs }"
      @open-auto-focus="event => {
        emits('openAutoFocus', event)
        if (event.target && !event.defaultPrevented) {
          handleCalendarInitialFocus(event.target)
          event.preventDefault()
        }
      }"
    >
      <slot />
    </PopoverContent>
  </PopoverPortal>
```

kricsleo:
Yeah, consistent behavior is better. Updated! 😁
(The failed tests seem irrelevant to this PR and do not reproduce locally.)

PR #2058, opened 2025-07-15
feat(Checkbox): add max prop

malik-jouda:
## Summary by CodeRabbit

* **New Features**
  * Added maximum selection limit capability to checkbox groups. Checkboxes automatically disable once the specified maximum number of selections is reached, preventing users from selecting additional items.

### DIFF
--- a/packages/core/src/Checkbox/CheckboxGroupRoot.vue
+++ b/packages/core/src/Checkbox/CheckboxGroupRoot.vue
@@ -16,6 +16,8 @@ export interface CheckboxGroupRootProps<T = AcceptableValue> extends Pick<Roving
   rovingFocus?: boolean
   /** When `true`, prevents the user from interacting with the checkboxes */
   disabled?: boolean
+  /** The maximum number of values that can be selected */
+  max?: number
 }
 
 export type CheckboxGroupRootEmits<T = AcceptableValue> = {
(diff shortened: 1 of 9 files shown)


zernonia:
Hi @malik-jouda . Thanks for the PR. But could put in more description of what this PR is trying to implement?

malik-jouda:
@zernonia 
https://github.com/nuxt/ui/issues/4057

cyyynthia:
While functionally sound, I have concerns <abbr title="with regards to">wrt</abbr> accessibility and user experience. It is very reasonable to come to the conclusion the checkboxes are "broken" if there is no feedback to the user that they have reached the limit (both visually, and via tools like screen readers).

It may be a good use-case for using `aria-disabled` **without** `disabled`: keep the fields discoverable but clearly convey that they cannot be interacted with (along with an appropriate message read aloud by <abbr title="screen readers">SR</abbr>s)

- https://ux.stackexchange.com/questions/16659/how-do-i-limit-the-number-of-options-a-user-may-choose-in-a-list
- (Heads up worth documenting) https://ux.stackexchange.com/questions/147363/how-can-i-relay-to-the-user-that-there-is-a-max-number-of-profiles-available

cyyynthia:
There are also some commits that seem unrelated to the PR... 😕

malik-jouda:
@zernonia Could you review the latest update?

zernonia:
@coderabbitai review

PR #2073, opened 2025-07-24
feat(Checkbox): Adding nested parent checkbox

lepture:
A `parent` prop is added to `CheckboxRoot`, a parent `CheckboxRoot` controls other checkboxes within the same checkbox group.

## Usage

Here is the example of a use case:

Token granted scopes:

- [ ] All user scopes
   - [ ] user profile
   - [ ] user email
- [ ] All Org scopes
   - [ ] Read org data
   - [ ] Edit org data

## Credits

https://base-ui.com/react/components/checkbox-group#parent-checkbox

### DIFF
--- a/docs/content/meta/CheckboxRoot.md
+++ b/docs/content/meta/CheckboxRoot.md
@@ -45,6 +45,12 @@
     'required': false
   },
   {
+    'name': 'parent',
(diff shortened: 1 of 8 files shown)


emilsgulbis:
What happened here?

PR #2074, opened 2025-07-24
feat: close combobox on blur with openOnFocus option

meirroth:
This PR updates the recent feature introduced in [#2050](https://github.com/unovue/reka-ui/pull/2050) to ensure consistent combobox behavior. Now, the combobox will close on blur when using `openOnFocus`—matching the default behavior when selecting an item without using the `openOnFocus` option. This aligns our component’s UX with the example shown in the Atlassian Design System: https://atlassian.design/components/select/country-select/examples.

Ref https://github.com/unovue/reka-ui/issues/1917

### DIFF
--- a/packages/core/src/Combobox/ComboboxInput.vue
+++ b/packages/core/src/Combobox/ComboboxInput.vue
@@ -60,6 +60,11 @@ function handleFocus() {
     rootContext.onOpenChange(true)
 }
 
+function handleBlur() {
+  if (rootContext.openOnFocus.value && rootContext.open.value)
+    rootContext.onOpenChange(false)
+}
+
 function handleClick() {
   if (rootContext.openOnClick.value && !rootContext.open.value)
     rootContext.onOpenChange(true)
@@ -121,6 +126,7 @@ watch(rootContext.filterState, () => {
     @input="handleInput"
     @keydown.down.up.prevent="handleKeyDown"
     @focus="handleFocus"
+    @blur="handleBlur"
   >
     <slot />
   </ListboxFilter>


meirroth:
There are two issues with this PR:
1. Tabbing doesn't close the combobox like it would without using the `openOnFocus` feature, instead focus is captured.
2. Selecting an item after searching for it doesn't get selected, see below:

https://github.com/user-attachments/assets/1086d519-b53b-40d2-be7c-d811ba3ec81c

meirroth:
And when you are already focused, and the combobox is closed, clicking again doesn't open it.
It is really strange because the wrapper we used applies the same exact code:
```vue
<template>
  <ComboboxInput v-bind="forwarded" @click="onFocus" @focus="onFocus" @blur="onBlur" />
</template>

<script setup lang="ts">
// https://github.com/unovue/reka-ui/issues/1702
import { ComboboxInput, injectComboboxRootContext, useForwardPropsEmits } from 'reka-ui'
import type { ComboboxInputEmits, ComboboxInputProps } from 'reka-ui'

const props = defineProps<
  ComboboxInputProps & {
    openOnFocus?: boolean
  }
>()
const emit = defineEmits<ComboboxInputEmits>()
const forwarded = useForwardPropsEmits(props, emit)

const rootCtx = injectComboboxRootContext()

function onFocus() {
  if (props.openOnFocus && !rootCtx.open.value) {
    rootCtx.onOpenChange(true)
  }
}

function onBlur() {
  if (props.openOnFocus && rootCtx.open.value) {
    rootCtx.onOpenChange(false)
  }
}
</script>

```

DJ1TJOO:
Hi, I'm also having the same problem. Would it be possible for someone to look at this?

meirroth:
@zernonia Hi :) any chance you can look into this issue?

zernonia:
Closing as this is now covered on `v2` — `ComboboxInput` closes the combobox on blur when focus leaves the root/content. Thanks for the contribution!

PR #2089, opened 2025-08-02
fix(useDateFormatter): compatibility for AM/PM switching

### LINKED ISSUE #2084: [Bug]: DatePicker AM/PM does not work on some locales
### Environment

```bash
Developement/Production OS: MacOS 15.5
Node version: 22.16.0
Package manager: pnpm@10.13.1
Reka UI version: 2.4.1
Vue version: Latest
```

https://stackblitz.com/edit/q1xfcq2r

### Steps to reproduce

Create a DatePicker component and set the locale to "es-MX". Then, select any date and try to edit the time. Finally, try changing between AM and PM.

### Describe the bug

I found that the AM/PM part of the input does not work on non-english locales. There are two main issues:

1. If I select a date, it sets the default AM/PM value but then I cannot change it. Even if I type `a` or `p` nothing changes.
2. If I press `a` or `p` several times, the model value changes once but the input does not show it. After that, I cannot change the AM/PM value anymore.

### Expected behavior

No matter the locale, I should always be able to change the AM/PM part o the input

_No response_

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/2084

### DIFF
--- a/packages/core/src/shared/useDateFormatter.ts
+++ b/packages/core/src/shared/useDateFormatter.ts
@@ -105,7 +105,8 @@ export function useDateFormatter(initialLocale: string, opts: DateFormatterOptio
       minute: 'numeric',
     }).formatToParts(date)
     const value = parts.find(p => p.type === 'dayPeriod')?.value
-    if (value === 'PM')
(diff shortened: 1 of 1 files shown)


epr3 (approved):
LGTM🥇

PR #2090, opened 2025-08-02
fix(Listbox): prevent focus on `ListboxContent` when `ListboxFilter` exists

### LINKED ISSUE #1808: [Bug]: ListboxFilter breaks keyboard navigation for Listbox
### Environment

```bash
See issue https://github.com/unovue/reka-ui/issues/1752 but Reka 2.2.0
```

https://codesandbox.io/p/devbox/w6jsw5

### Steps to reproduce

See issue https://github.com/unovue/reka-ui/issues/1752

### Describe the bug

Issue https://github.com/unovue/reka-ui/issues/1752 is not fixed in the latest release version, please consider reopening it.

### Expected behavior

_No response_

_No response_

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/1808

Not sure if preventing focus on `ListboxContent` when `ListboxFilter` exists is the right approach. However, based on the previous logic, keyboard navigations are already disabled (e.g., https://github.com/unovue/reka-ui/blob/v2/packages/core/src/Listbox/ListboxContent.vue#L38 and https://github.com/unovue/reka-ui/blob/v2/packages/core/src/Listbox/ListboxRoot.vue#L199), so preventing focus seems reasonable in this context.

Still, I'm wondering if we could keep `ListboxContent` focusable even with `ListboxFilter` and enable all keyboard navigations. 🤔
Any explanation would be greatly appreciated! ❤️


zernonia:
Thanks @kricsleo ! iirc the filter input should prevent the content to be focusable, and allow input to trigger all the keyboard navigation. Your fix seems to be good 👍🏻

PR #2105, opened 2025-08-08
Use `select` event instead f `click` in DropdownMenuItem demo

1nVitr0:
Since the example / demo is most often the place developers look first, it would make sense to use the exposed `select` event here, instead of a generic `click` event.

Originally I had planned to open an issue, because when using the click event, I had some strange behavior when selecting an item: Despite the menu closing, the `pointer-events: none` style was still applied to the `body`element. Despite my best efforts, I was not able to create a minimal (not) working example. My best guess is some weird timing issues in the async functions I used. If I do manage to reproduce the behaviour, I will create an issue.

Short story is, when I used `select` instead of `click`, the `pointer-events: none;` was removed correctly, so maybe this will prevent others from running into the same issue. And, I do think regardless of a possible issue, documenting the `select` event does make more sense than the current `@click`.

### DIFF
--- a/docs/components/demo/DropdownMenu/css/index.vue
+++ b/docs/components/demo/DropdownMenu/css/index.vue
@@ -25,7 +25,7 @@ const checkboxOne = ref(false)
 const checkboxTwo = ref(false)
 const person = ref('pedro')
 
-function handleClick() {
+function handleSelect() {
   // eslint-disable-next-line no-alert
   alert('hello!')
 }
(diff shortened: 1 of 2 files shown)


1nVitr0:
If anyone does come across the same issue I had, it was consistently "caused", essentially by changing an async function somewhere in the `click` callback tree from

```js
async () => {
    // Do some stuff
    return true;
}
```

to

```js
async () => {
    // Do the same stuff as above
    return await Promise.resolve(true);
}
```

Just by plugging that into the `@click` event callback didn't reproduce the issue, and I did not find any other special circumstances in my own code that could cause it. Will open a separate issue if I find out what exactly causes it

zernonia:
That's an odd case where you run into the `click` event issue. Yeah do log a ticket if you found the specific case where it might failed. 

For now, we can update the docs to use `@select` no problem  😁

PR #2119, opened 2025-08-17
fix(Menu):  correct emitted event types to match runtime behavior

### LINKED ISSUE #2118: [Feature]: DropdownMenuContent should emit openAutoFocus
### Describe the feature

I think this might be an oversight, because DropdownMenuSubContent already emits `openAutoFocus`. I don't think there are any technical limitations, the event just needs to be propagated from here: https://github.com/unovue/reka-ui/blob/v2/packages/core/src/Menu/MenuContentImpl.vue#L155.

My goal is to have a handler here to prevent the default auto focus from executing.

### Additional information

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/2118

The `entryFocus`/`openAutoFocus` events were removed in PR https://github.com/unovue/reka-ui/pull/665/files#diff-8c2f58c8ccca9f2d06c2bde73a649a7b23180c696e73ede751e31c6623b5254dR7. I'm not sure why, but this caused the event types to mismatch the runtime behavior (these events are still emitted). So, I think it's better to revert this type omitting change.

### DIFF
--- a/packages/core/src/Menu/MenuContent.vue
+++ b/packages/core/src/Menu/MenuContent.vue
@@ -4,7 +4,7 @@ import type {
   MenuRootContentTypeProps,
 } from './MenuContentImpl.vue'
 
-export type MenuContentEmits = Omit<MenuContentImplEmits, 'entryFocus' | 'openAutoFocus'>
+export type MenuContentEmits = MenuContentImplEmits
 
 export interface MenuContentProps extends MenuRootContentTypeProps {
   /**


zernonia:
@kricsleo I rmb those 2 emit was designed to be use internally only. Exposing it might leads to unintended behavior if user should choose to `event.preventDefault`

kricsleo:
So you're suggesting we remove the runtime behavior as well as the types? 🤔
(Regardless, it's better to align the runtime behavior with types.)

PR #2134, opened 2025-08-26
WIP: feat: upgrade to vue-tsc v3

Y0me:
This PR begins addressing #2043 (opened by @cyyynthia) by upgrading to Vue Language Services v3 and starting to resolve the typing issues it surfaces.

What’s inside :
- packages/core -> vue-tsc@3.0.6 for type-checking.
- initial fixes for type-safety issues across components, mostly around event typing and safer element handling.
- introduced new DOM utilities to replace unsafe casts (need to check if this affects performance, especially in Slider component).

Notes :
- New DOM helpers are strictly typed for HTMLElement, which excludes SVGElement. This shouldn’t affect most use cases, but it’s worth keeping in mind if any components interact with SVG nodes.

This is still a WIP — I haven’t had time to fully test everything yet. Most changes are small and localized, but I’d really appreciate feedback and a fresh pair of eyes to help bring this across the finish line.


packages/core/src/shared/elementUtils.ts
```diff
@@ -0,0 +1,36 @@
+import type { ComponentPublicInstance } from 'vue'
+
+/**
+ * Check if the target is a HTMLElement
+ * @param target - The target to check
+ * @returns True if the target is a HTMLElement, false otherwise
+ */
+export function isHTMLElement(target: unknown): target is HTMLElement {
+  return !!target && target instanceof HTMLElement
```
cyyynthia: The `!!target` gate is unnecessary; null and undefined are valid operands of instanceof and will produce false as expected

packages/core/src/shared/elementUtils.ts
```diff
@@ -0,0 +1,36 @@
...
+  return isHTMLElement(target) ? target : undefined
+}
+
+/**
+ * Retrieves the underlying HTMLElement from a given Element or Vue component instance.
+ *
+ * @param vnode - An Element or Vue ComponentPublicInstance, or null.
+ * @returns The corresponding HTMLElement if available, otherwise undefined.
+ */
+export function getHTMLElementFromVNode(vnode: Element | ComponentPublicInstance | null): HTMLElement | undefined {
```
cyyynthia, line 27: Since Reka already heavily leverages vueuse it might be a good idea to rely on their helpers instead, which are compatible with SVG elements

https://vueuse.org/core/unrefElement

packages/core/src/Toast/ToastViewport.vue
```diff
@@ -172,9 +172,8 @@ function getSortedTabbableCandidates({ tabbingDirection }: { tabbingDirection: '
   >
     <FocusProxy
       v-if="hasToasts"
-      :ref="(node: ComponentPublicInstance) => {
-        headFocusProxyRef = unrefElement(node) as HTMLElement
-        return undefined
+      :ref="(node) => {
```
cyyynthia, line 175: Not sure how to feel about these refs; would perhaps be better to keep the raw ref and unwrap when needed..? 🤔

packages/core/src/Menu/MenuItem.vue
cyyynthia, line 1: Moving the handlers out of the template is a good idea imho, perhaps it'd be a good idea to also do this change in other files? 🤔

cyyynthia (changes requested):
**Disclaimer**: I'm not a maintainer and don't speak on behalf of the Reka maintainers or the unovue team 😅

packages/core/src/Menu/MenuItem.vue
Y0me: Yep, wanting to do that for other files too... But motivation at this time of day was falling apart. 😅

packages/core/src/shared/elementUtils.ts
Y0me: Will look into that this week

stefnotch:
Is this PR still relevant, seeing as how reka-ui has been upgraded to vue-tsc 3?

https://github.com/unovue/reka-ui/blob/2c6d918e84571350cc5a2238775abee02c037b3d/packages/core/package.json#L127

If yes, is there an easy subset of the changes that could be immediately landed?

Issue #2137, opened 2025-08-28
[Bug]: TagsInputInput.vue delimiter problem


JhnBer:
### Environment

```bash
Developement/Production OS: Alpine Linux v3.22
Node version: v22.16.0
Package manager: npm@11.3.0
Reka UI version: 2.5.0
Vue version: 3.4.19
CSS framework: tailwindcss@4.0.14
Client OS: mac OS Sequoia 15.3.2
Browser: Chrome 139.0.7258.67
```

### Link to minimal reproduction

https://stackblitz.com/edit/ewrpv97f

### Steps to reproduce

Enter the tag, select the text, replace with a delimiter.
Add delimiter only.
Add several spaces, then use delimiter to create new tag.

You can use the buttons for all three cases at [Stackblitz](https://stackblitz.com/edit/ewrpv97f) or try to reproduce them directly here https://reka-ui.com/docs/components/tags-input 

### Describe the bug

All of the above actions result in the creation of empty tags, or tags consisting entirely of spaces.

### Expected behavior

Replacing the selected text with a separator symbol clears the field.
Trying to enter only the separator should clear it immediately.
Trying to create separate tags as spaces should also clear the field.

### Context & Screenshots (if applicable)

Fix in #2136

PR #2138, opened 2025-08-30
fix(TagsInputInput.vue): delimiter no longer works if input is empty …

Tell-Your-Cat-i-Said-pspsps:
-added allowEmptyTags prop to TagInputRoot,vue.
-created handleTagAdd function in TagInputInput.vue that will check if empty tags are allowed or not.
#2137


zernonia:
Thanks for the PR @Tell-Your-Cat-i-Said-pspsps ! However this is duplicate of https://github.com/unovue/reka-ui/pull/2136 and I believe we dont need this as a prop 😁

PR #2145, opened 2025-09-05
feat(Checkbox): add generic v-model

DamianGlowala:
This correctly inferes types from passed `modelValue` and avoids type issues for those who e.g. do not need the indeterminate state.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed checkbox accessibility attributes to return proper string values ('true', 'false', 'mixed') for better compatibility with screen readers and assistive technologies.

* **Refactor**
  * Enhanced component type safety with generic type support for improved developer experience and IDE type hints.


DamianGlowala:
@sadeghbarati do you have any idea if the test failure is related to this PR (which only changes types)?

sadeghbarati:
Hi, sry for late response

Yes the types is become runtime thing after Vue compiler, compiles the template

[Vue SFC Playground](https://play.vuejs.org/#eNqNU01v1DAQ/SuDL1ukshGC0yoNgqoScICqXcHFlzSZpC6ObdmTbVDIf2fsdNt0Va1687w3M37zNYrPzq13PYqNyEPllSMISL0rpFGds55gBI8NTNB428GKXVfSSFNZEwi60MJZ5E9WX1FrC7+t1/Wb1Vtp8mxOx4nYIOycLgnZAshv3xfjmIKnKc/YSqgyrifYvetsjfpMCualgIzJPFvEi1NBgb9vVLu+C9aw8jHGS1HZzimN/qcjxfKk2EBiIleyuvvvCSPf4+ker26x+vMCfheGiElx6TGg36EUjxyVvkWa6YvrHzjw+5Fk9b1m7yPkFQar+6hxdvvSm5plL/yS2m+p/8q023AxEJqwLyoKjZ5T8peCZ3J+pPQnuR/WH1OcNBN3McYcjF6XpuXOEyfgTuOQFoD+OoTz2Cesr4mHwCO/sVZjaeAfrJSpkdB3yjCVduMhThmGm7J6CL6xw5W1dOmtC/kWuDFo6nCYeGkWcw01NmWv6Vepe/y0ge0htr23DB9RxPU+38bloi9rhhYNelWx+SqBqUv7W3CxMHZgacrgXOYLhRfFCV/H8fuo1S49+Bm0pXQD/M5m/Pk1TP8BcmBKGQ==)



```ts
export type CheckedState = boolean | 'indeterminate'

export interface CheckboxRootProps<T extends CheckedState = CheckedState> {
  defaultValue?: T
  defaultValueTwo?: boolean | 'indeterminate'
}

const props = defineProps<CheckboxRootProps<T>>()
```

this will become

```ts
props: {
  defaultValue: { type: null, required: false },
  defaultValueTwo: { type: [Boolean, String], required: false }
},
```

generic type will become `null` in runtime instead of `[Boolean, String]` and `aria-checked` test will fail because type is `null`
aria-checked must have nullable check for `checkboxState`

Asking AI for this check returned this

```ts
:aria-checked="isIndeterminate(checkboxState) 
  ? 'mixed' 
  : checkboxState === true 
    ? 'true' 
    : 'false'"
```

zernonia:
@coderabbitai review

zernonia:
Thanks for this, @DamianGlowala, and apologies it sat for a while.

We're going to close this in favor of #2410 (`feat(Checkbox/Switch): add support for custom true/false values`), which already landed on `v2` and provides generic v-model through a different design — it makes the **value itself** generic via `trueValue`/`falseValue`, so consumers can bind arbitrary values (`'yes'`/`'no'`, objects, etc.). Because the model value is exactly what the consumer supplies, that approach stays type-sound without internal casts.

The constrain-to-`CheckedState` approach here runs into two issues we'd rather avoid:
- A checkbox can always emit `true`, `false`, or `'indeterminate'` internally, so narrowing `TCheckedState` requires the `as TCheckedState` casts throughout the body — the narrowed type isn't actually guaranteed at runtime.
- Making the prop generic erases its runtime type to `{ type: null }` (losing the `[Boolean, String]` validation), which is what broke the `aria-checked` test and prompted the workaround.

Really appreciate the effort here — the underlying goal (cleaner inferred types for folks not using `indeterminate`) is covered by what shipped in #2410. Closing this out; happy to revisit if you hit a gap that #2410 doesn't cover. 🙏

Issue #2160, opened 2025-09-12
[Bug]: Select text flickers with the placeholder on change


NoelDeMartin:
I've noticed that changing from one option to another in the Select component, there is a small flicker showing the empty selection placeholder. I've looked into the code, and I've found the root of the problem. You can find a proposal to fix this in #2161, but I'm not sure that's the right fix so I'm also opening this issue to track it separately.

## Reproduction

You can reproduce this [in the playground](https://reka-ui.com/docs/components/select#select), but here's a slowed down video:

https://github.com/user-attachments/assets/bb9abe8e-1545-4d07-8185-bb749e8a0c07

## Origin

After digging into the code, I found that the root of this issue can be found in the [SelectContent](https://github.com/unovue/reka-ui/blob/v2/packages/core/src/Select/SelectContent.vue) component:

https://github.com/unovue/reka-ui/blob/483cd743ffaab033acf21e41f3d9a5f2bb67388a/packages/core/src/Select/SelectContent.vue#L59..L75

Essentially, when the select is open/closed, the content of the select is toggled between a popover or a fragment outside of the DOM. This causes [some internal mounted/unmounted hooks](https://github.com/unovue/reka-ui/blob/483cd743ffaab033acf21e41f3d9a5f2bb67388a/packages/core/src/Select/SelectItemText.vue#L50) that remove all the options from the select, and add them again in the following frame. With the unintended consequence that during one frame, the Select doesn't have any options available, and that's why it's rendering the placeholder.

DamianGlowala:
Came here to report the exact same bug, you were first! 😄 fingers crossed this gets addressed soon, as select isn't much user-friendly with this flicker

Hwacc:
I submitted a PR to address this issue, hoping it can help you.  [#2239](https://github.com/unovue/reka-ui/pull/2239)

chefhasteeth:
@Hwacc since your PR has been open since late October, do you know if there's a way we can fix this in the meantime when creating our own select components based off Reka UI's? Or are we stuck until the fix is merged?

NoelDeMartin:
@chefhasteeth You can apply the changes in the PR to your project using [patch-package](https://www.npmjs.com/package/patch-package) (or `pnpm patch` if you use pnpm). Make sure to modify the built files, not the Typescript source, though.

DamianGlowala:
What's the reason for shipping both `src/` and `dist/` dirs?

NoelDeMartin:
[...]

PR #2161, opened 2025-09-12
#2160 Fix Select flicker

NoelDeMartin:
I have documented this issue in #2160, this PR contains a proposal to fix it.

My solution is, basically, to always render the options in the Fragment; regardless of whether the options are showing in the popover or not. I'm not sure if rendering them twice can have any unintended consequences, but from the few tests I've done, it seems to work fine.

Edit: Well, it seems the tests are not passing so I'm converting this to a draft 😅. I'll look into it, but in the meantime if someone comes across this I'd appreciate some feedback on whether this is the right fix.

Edit 2: I've look into the tests, and honestly they are too coupled to internals of how the component works. I think a maintainer should look at this before I waste any time trying to understand that, so I'm moving this back to Ready for review until I get some feedback.

### DIFF
--- a/packages/core/src/Select/SelectContent.vue
+++ b/packages/core/src/Select/SelectContent.vue
@@ -66,7 +66,7 @@ watch(present, () => {
     </SelectContentImpl>
   </Presence>
 
-  <div v-else-if="fragment">
+  <div v-if="fragment">
     <Teleport :to="fragment">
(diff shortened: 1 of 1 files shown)


zernonia:
Closing in favour of #2329, which carries the Select flicker fix for #2160. Thanks for documenting and proposing a fix!

PR #2180, opened 2025-09-22
feat(TooltipProvider): content configuration

### LINKED ISSUE #2179: [Feature]: global tooltip content configuration
### Describe the feature

It would be useful to be able to globally set the values of `TooltipContent`. The main use cases may be changing the default `side` of the tooltip or configuring the `sideOffset` for all tooltips.

### Additional information

J-Michalek:
Resolves https://github.com/unovue/reka-ui/issues/2179


sadeghbarati (commented):
In TooltipContentImpl.vue it's better to use `useForwardProps` to fix undefined boolean casting curse 😄 

Boolean Casting Curse in Vue is `undefined` boolean props will transfer to `false` instead of staying `undefined`


```ts
const props = withDefaults(defineProps<TooltipContentImplProps>(), {
  side: 'top',
  sideOffset: 0,
  align: 'center',
  avoidCollisions: true,
  collisionBoundary: () => [],
  collisionPadding: 0,
  arrowPadding: 0,
  sticky: 'partial',
  hideWhenDetached: false,
})

const delegatedProps = reactiveOmit(props, ['ariaLabel'])

const popperContentProps = computed(() => defu(delegatedProps.value, providerContext.content.value ?? {}))

const forwarded = useForwardProps(popperContentProps)
```

Then use `forwarded` like

```diff
- v-bind="{ ...$attrs, ...popperContentProps }"
+ v-bind="{ ...$attrs, ...forwarded }"
```

---

Related PR https://github.com/unovue/reka-ui/pull/610
https://github.com/vuejs/vue/issues/4792

J-Michalek:
I've tried refactoring this, but it does not work as intended - the `defu` function gives the left most object the highest priority so if we set the default prop values then the `providerContext.content.value` will never have any effect and without the default values set we cannot use `useForwardProps` as it only returns the keys that are set on the props (atleast to my understanding) which are possibly all undefined as there would be no default values.

J-Michalek:
@sadeghbarati Just checking in if there are some changes to be done or if this can be merged.

sadeghbarati:
Hello 👋 

Zernonia is the maintainer 
He should review, too

I can only merge the small PRs 😁

zernonia:
Im hesitant to this feature actually. Wouldn't it be much simpler if you create a generic Tooltip component (with defined Content configuration), and use it like that?

J-Michalek:
@zernonia I totally get your point, but my use case comes from using NuxtUI, which by default uses the bottom side for tooltips. Being able to configure it like this seems like a better solution than creating arbitrary components with predefined props.

If I was using RekaUI directly for tooltips then I would handle it in the component myself.

zernonia:
@J-Michalek code wise LGTM. Can we add a test for this feature? 😁

zernonia (approved):

J-Michalek:
I've added a test case for this.

PR #2216, opened 2025-10-10
fix(Tooltip): allow SVG as trigger

### LINKED ISSUE #2197: [Bug]: Tooltips triggers weirdly when trigger contains an SVG
### Environment

```bash
Developement/Production OS: Stackblitz
Node version: v20.19.1
Package manager: pnpm@8.15.6
Reka UI version: 2.5.1
Vue version: 3.5.21
Nuxt version: —
Nuxt mode: —
Nuxt target: —
CSS framework: tailwindcss@3.4.13
Client OS: Fedora 42
Browser: Firefox 143.0.1, Chromium 140.0.7339.207
```

https://stackblitz.com/edit/fkeaaicg?file=src%2FApp.vue

### Steps to reproduce

- Copy the tooltip example from the docs
- Replace the icon in the trigger with an SVG
- Place multiple of such elements with individual tooltips in a row

### Describe the bug

I'm not even sure how to exactly describe it, video below shows that something is clearly broken though.

Tooltip show/hide logic breaks somehow -- sometimes the tooltip does not show at all when hovering, sometimes it shows when the mouse cursor *leaves* the target, sometimes it doesn't hide when another tooltip shows.

### Expected behavior

Expected tooltips to show when cursor enters the trigger and to hide when cursor exists the trigger, like it does when there are no SVGs in the trigger.

https://github.com/user-attachments/assets/230df2a4-11b4-424b-a028-e040c2eb0cd6

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/2197

There are other places where we use `instance HTMLElement` as a guard (search for `instance HTMLElement` in the codebase); these might also need to change to `instanceof Element` to allow `SVGElement`. However, to minimize changes, I've left the others as they are for now, pending further reports.

### DIFF
--- a/packages/core/src/shared/useGraceArea.ts
+++ b/packages/core/src/shared/useGraceArea.ts
@@ -44,7 +44,7 @@ export function useGraceArea(triggerElement: Ref<HTMLElement | undefined>, conta
   watchEffect((cleanupFn) => {
     if (pointerGraceArea.value) {
       const handleTrackPointerGrace = (event: PointerEvent) => {
-        if (!pointerGraceArea.value || !(event.target instanceof HTMLElement))
+        if (!pointerGraceArea.value || !(event.target instanceof Element))
           return
 
         const target = event.target


DamianGlowala:
@kricsleo is this PR sufficient to use SVG **path** as a tooltip trigger?

kricsleo:
> [Open in StackBlitz](https://pkg.pr.new/template/ed61895e-d672-4b52-813d-0aaac06b23d4)
> 
> ```
> npm i https://pkg.pr.new/reka-ui@2216
> ```
> 
> _commit: [`d04c282`](https://github.com/unovue/reka-ui/runs/52456144539)_

@DamianGlowala You could try it out with this.

zernonia:
Thanks for the fix @kricsleo ! Indeed that would fix the problem!

PR #1916, opened 2025-05-19
fix(Select): render to the latest state before transition

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/1865

This is a tricky fix for the `Select` component's transition 🤷. It should enable both types of transitions to function correctly:

- Transitions with stateful attributes, e.g., `class="data-[state=closed]:animate-slideLeftAndFade"`

|Before|After|
|:-:|:-:|
|https://stackblitz.com/edit/vda7rpmu?file=src%2FApp.vue <br>(The transition on close didn't work)| https://stackblitz.com/edit/vda7rpmu-9ey4pgkn?file=src%2FApp.vue <br> (Works)|

- Transitions with `<Transition>` component.

|Before|After|
|:-:|:-:|
|https://stackblitz.com/edit/vda7rpmu-ofgwcgww?file=src%2FApp.vue <br>(Works)| https://stackblitz.com/edit/vda7rpmu-g2udpau4?file=src%2FApp.vue <br> (Works)|

PR #2239, opened 2025-10-22
fix(Select): select value flash when selected item

### LINKED ISSUE #2160: [Bug]: Select text flickers with the placeholder on change

Hwacc:
## 📝 Summary

Fix Select component flickering issue

## 🔗 Related Issues

resolves https://github.com/unovue/reka-ui/issues/2160

## 🧪 How it was caused

It needs to be traced back to this PR [#1916](https://github.com/unovue/reka-ui/pull/1916)
Using setTimeout to delay and control the rendering flow is a bad solution for handling animation.
The analysis of the causes of the flickering is as follows:

1. When we use setTimeout, the whole flow:

bug1

2. The original flow likes:

origin1

## ✅ Solutions

Very easy: we fully leverage the `<Presence>` component’s features to control the switching of child components.
but we need tag `<Presence>`  as  `force-mount` to keep it alive, becasue we focus in its child components, don’t want it to be destroyed immediately when closing.

~~In addition, we use `force-mount` in `<Presence>`, so,  `SelectContentProps`'s forceMount prop was no longer works. Should we consider removing it?~~

## 📸 Screenshots 

Now, we can use `data-[state=close]` to control our close animate as usual

https://github.com/user-attachments/assets/fe7dfdd6-3103-45b0-afde-9c7a5fd2ef75


RitvikSardana:
Please merge this ASAP.

This bug is very irritating

jpmsantana:
@Hwacc what do we need to merge this? this bug irritates the users who use the pages that are consuming this component.

RitvikSardana:
Why is no one updating this PR?

Could anyone please work on it?

Hwacc:
I’ve resolved the merge conflicts in v2.7.0 and looking forward to review. It looks like quite a few people have run into issues with this Select component. @kricsleo , do we have plans to merge this PR into the next release? Thanks.

kricsleo:
Sorry for the late response. I normally don't merge PRs 🙏, so I'll leave it to the author, [zernonia](https://github.com/zernonia).

ragingdave:
I think @kricsleo meant to tag @zernonia and not just link to him here. (Also I'm doing it to highlight the fix as well).

zernonia:
Closing in favour of #2329, which cherry-picks this fix (with co-author credit) for #2160. Thanks for tracing the root cause!

PR #2250, opened 2025-10-28
fix: use `toValue` from Vue, but not from VueUse

niusia-ua:
The `toValue` utility function from VueUse is deprecated since v12 in favor of the same function from Vue itself.
In a recent release of VueUse v14, it was completely removed, which causes runtime errors.

This is the only place in Reka UI where an old `toValue` function was still used.


TheAlexLichter:
@zernonia any chance to release this as a fix soon? This breaks Reka-UI usage with VueUse 14, so most Vue apps (that upgraded to v14) 😛.

PR #2251, opened 2025-10-29
fix(components): VisuallyHiddenInput fails Axe tests inside of form element

J-Michalek:
In my project while using Axe I encountered an issue when using RadioGroup inside of a form element. It complains about the hidden input not having a label, but I'm not sure what is the best way to handle this.

This is most likely affecting other component as well.


zernonia:
Thanks for flagging this originally! The VisuallyHiddenInput a11y violations inside forms (nested-interactive + the missing `aria-hidden`/label) are now fixed in #2720, which also keeps consumer `<style scoped>` working on the now multi-root components. Closing this in favour of that one — appreciate you surfacing it. 🙏

PR #2254, opened 2025-10-30
fix (TimeField): call to updateHour to handle 12 hour time

### LINKED ISSUE #2262: [Bug]: TimeField hour input handling broken for 12 hour time
### Environment

```bash
stackblitz
```

https://stackblitz.com/edit/qiv1v48j-e18akmrk?file=src%2FApp.vue

### Steps to reproduce

This only happens in locales that use 12 hour time

1.) Open stackblitz link (locale set to en-US)
2.) Click on hour input
3.) Type '1' 
4.) Day period switches to am (not expected)
5.) Type '1' again
6.) Next field is focused 

### Describe the bug

updateHour method does not handle 12 hour time correctly. You can view a fix here: #2254 

### Expected behavior

I would expect that the time input display 11:00 PM. instead it displays 11:00 AM

_No response_

ChrisSmartGP:
First up i'd just like to say thanks so much for all the work that has gone into this library :)

I've encountered an issue where setting the time does not work when 'pm' is set and the user tries to enter the values 10, 11 or 12 in the hour input. 

When in a locale that uses 12 hour time format:

- the `updateHour` method should have a max accepted value of 12
- the internally represented time should convert back to 12 hour format before using the `updateHour` method. Not doing so will cause the method to return early
- I added a method to determine the hourCycle from the locale but I think maybe this should just use the props.hourCycle. However props.hourCycle is undefined unless it is explicitly passed

(resolves https://github.com/unovue/reka-ui/issues/2262)

## Summary by CodeRabbit

* **New Features**
  * Added 12-hour time format support with proper AM/PM handling, including correct hour wrapping and input behavior consistent with locale hour cycle.

* **Tests**
  * Added comprehensive tests covering 12-hour input, AM/PM conversions, and arrow-key cycling behavior to ensure consistent user-facing time interactions.


zernonia:
Yo @ChrisSmartGP ! Appreciate the PR, could you create a ticket and explain the issue with reproduction please? 🙏🏼

ChrisSmartGP:
@zernonia thanks for giving this some attention 🙏 have added an issue with reproduction #2262

ChrisSmartGP:
Hi @zernonia have you had a chance to look at this ?

sadeghbarati (commented):
Hi @ChrisSmartGP, tested your StackBlitz with pkg.pr.new, it fixes the `PM` switch to `AM` problem, but
When I start from 1 to 12, the `PM` will change to `AM.`

Each time it reaches 12, it toggles between `AM` and `PM.`

---

The code from react-aria might help you here

https://github.com/adobe/react-spectrum/blob/main/packages/%40react-stately/datepicker/src/useDateFieldState.ts

ChrisSmartGP:
@sadeghbarati thanks for acknowledging the fix; when using the arrow keys on the hour input it will cycle between am and pm which is great.

This fix is primarily aimed at stopping the picker from switching back to am when a user types a new value in the hour input.

Thanks for the link to adobe code; they have some handling for 12 hour time [here](https://github.com/adobe/react-spectrum/blob/f057a1f413bbf243d4b34d822f68372b836fcef4/packages/%40react-stately/datepicker/src/useDateFieldState.ts#L618). Reka's updateHour method is missing this and only considers 24 hour time. This PR updates it to handle 12 hour time aswell

PR #2265, opened 2025-11-07
fix: a11y aria attribute in RangeCalendarCellTrigger.vue

### LINKED ISSUE #2261: [Bug]:RangeCalendarCellTrigger contains incorrect aria-selected attribute
### Environment

```bash
Node version: 16.0.0
Package manager: pnpm@8.6.0
Reka UI version: 2.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
Browser: Chrome 90.0.4430.212
```

https://www.reka-ui.com/docs/components/date-range-picker#date-range-picker

### Steps to reproduce

Open DateRangePicker, select any dates and run axe a11y report.

### Describe the bug

The issue is that role of the element is set to button, but `aria-selected` attribute is not allowed on buttons, it should be `aria-pressed`. In this case axe a11y test passes and what even more important Voiseover can spell selected on the button if the day is selected, also it mention the type of button `toggle-button` 

### Expected behavior

a11y test should pass and Voiseover spell the right information

Image

wolandec:
Fixes: https://github.com/unovue/reka-ui/issues/2261

This is a targeted a11y fix that brings the component’s state semantics in line with ARIA best practices for toggle buttons. It improves screen reader announcements and a11y conformance with minimal surface-area change, with the primary actionable impact being potential updates required for tests or selectors that previously relied on `aria-selected`. 

### Why this change was made
- `aria-selected` is intended for selectable items in widgets like `listbox`, `grid`, `tablist`, or `tree` (e.g., `option`, `row`, `gridcell`, `tab`).
- The component renders with `role="button"`, which follows the button/toggle button pattern. For toggle buttons, the correct state attribute is `aria-pressed` (conveys on/off, pressed/not pressed).
- Therefore, using `aria-selected` on a `button`-role element was semantically incorrect and could confuse assistive technologies. Changing to `aria-pressed` aligns the control with ARIA Authoring Practices for toggle buttons.

### How it impacts behavior and results
User-facing behavior (mouse/keyboard interactions) is unchanged; this is a semantic accessibility improvement. Notable impacts:

- Screen reader output:
  - Before: A screen reader could announce a button with an irrelevant/ignored `aria-selected` state, or misinterpret the selection state.
  - After: The element is announced as a toggle button with a proper “pressed” state when the date is selected, improving clarity for non-visual users.

- Accessibility conformance:
  - The component better matches WAI-ARIA patterns, addressing potential violations flagged by a11y linters or automated scanners.

- Styling and DOM querying:
  - CSS/JS that targets `[aria-selected="true"]` on this element will no longer match. However, the component still exposes `:data-selected` with the same condition, which is more stable for styling/tests. If any consumer relied on `aria-selected`, they should switch to `[data-selected="true"]`.

### Rationale in the context of this component
- The calendar day cell is rendered as a button and acts like a toggle to select/deselect a date (subject to `allowNonContiguousRanges` and `isUnavailable`). Using `aria-pressed` correctly communicates its toggled state.

### Risks and edge cases
- Consumers relying on `[aria-selected]` must update selectors.
- If some screen readers previously special-cased `aria-selected` on non-matching roles, users might hear a different announcement now (“pressed” vs. “selected”). This is a correction, not a regression, but be aware of change in terminology.


wolandec:
@zernonia I was wondering if you could have a look at this PR, please?

PR #2294, opened 2025-11-21
fix (TooltipContent): fix empty ariaLabel for nested components

### LINKED ISSUE #1810: [Bug]: TooltipContentImpl don't render tooltip contents if it's within another component
### Environment

```bash
Node version: 16.0.0
Package manager: pnpm@8.6.0
Reka UI version: 2.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
Browser: Chrome 90.0.4430.212
```

https://stackblitz.com/edit/nbbjhz7c?file=src%2FTooltipContents.vue,src%2FApp.vue

### Steps to reproduce

Having another component in [`<TooltipContent>`](https://reka-ui.com/docs/components/tooltip#content) instead of the content itself 

### Describe the bug

Breaks screen readers as the content is not appended in the `<VisuallyHidden>` component. 

### Expected behavior

Content should be appended and read by screen readers. The content should be repeated in the highlighted element:

Image

The `ariaLabel` computed can't handle nested components (`children` will be empty): https://github.com/unovue/reka-ui/blob/v2/packages/core/src/Tooltip/TooltipContentImpl.vue#L70-L84

anjarupnik:
Resolves [#1810](https://github.com/unovue/reka-ui/issues/1810)

### DIFF
--- a/packages/core/src/Tooltip/TooltipContentImpl.vue
+++ b/packages/core/src/Tooltip/TooltipContentImpl.vue
@@ -1,5 +1,4 @@
 <script lang="ts">
-import type { VNode } from 'vue'
 import type { PopperContentProps } from '@/Popper'
 import type { PrimitiveProps } from '@/Primitive'
 import { useForwardExpose } from '@/shared'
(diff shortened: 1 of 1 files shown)


zernonia (changes requested):
Thanks for the PR @anjarupnik , however checking locally I dont think your solution works

anjarupnik:
Hi @zernonia, sorry about that. My first fix only worked when another component with text in its default slot was passed in, but not when text was placed directly inside the component (like in the bug report). I’ve changed the approach so it now reads textContent from currentElement. What do you think? I tested it locally and it seems to work nicely in all the cases. Could you check it again on your side? Thanks!

zernonia:
Thanks @anjarupnik  for the fix, what a simple solution and it's still working for accessibility!

PR #2300, opened 2025-11-22
fix(AccordionItem): prefer using `props.unmountOnHide` (close #2295)

### LINKED ISSUE #2295: AccordionItem's prop unmountOnHide is not working
## Doc
Image

## Source Code
https://github.com/unovue/reka-ui/blob/1e0670da1d9cf9652cb4cf89aa2015265dbd60d0/packages/core/src/Accordion/AccordionItem.vue#L118

## How about this?

```
:unmount-on-hide="props.unmountOnHide ?? rootContext.unmountOnHide.value"
```

BuptStEve:


zernonia:
@BuptStEve  what does this changes fix?

BuptStEve:
> @BuptStEve what does this changes fix?

https://github.com/unovue/reka-ui/issues/2295

For example, I want to control only the first item unmount on hide.

image

zernonia:
Got it. Thanks for the fix @BuptStEve !

PR #2307, opened 2025-11-24
Ensure FocusScope adheres to the Dialog WAI-ARIA design pattern

callumacrae:
Per the documentation for the popover component, the popover component adheres to the Dialog WAI-ARIA design pattern: https://reka-ui.com/docs/components/popover#accessibility

However, when testing, if the popover contains a list of links and then a button, the focus will go to the button, not the link. From my understanding of the design pattern, this isn't correct as links are focusable elements too.

I'm not sure why it was originally like this - the PR that added the code doesn't give any details and there's no comments on the function or usages: https://github.com/unovue/reka-ui/pull/281

(the pattern can be found here: https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/)

----

This is my first contribution to this project - you don't have a CONTRIBUTING.md so let me know if I've done anything wrong! I'm just joined your discord too if that's quicker than communicating here :)

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined focus scope initialization to optimize which elements are eligible for focus assignment when components mount, improving initial focus behavior.

### DIFF
--- a/packages/core/src/FocusScope/FocusScope.vue
+++ b/packages/core/src/FocusScope/FocusScope.vue
@@ -37,7 +37,7 @@ export interface FocusScopeProps extends PrimitiveProps {
 import { isClient } from '@vueuse/shared'
 import { nextTick, reactive, ref, watchEffect } from 'vue'
 import { Primitive } from '@/Primitive'
-import { createFocusScopesStack, removeLinks } from './stack'
+import { createFocusScopesStack } from './stack'
 import {
   AUTOFOCUS_ON_MOUNT,
   AUTOFOCUS_ON_UNMOUNT,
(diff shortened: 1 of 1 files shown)


sadeghbarati:
Related `radix-ui` PRs

- https://github.com/radix-ui/primitives/pull/1796
- https://github.com/radix-ui/primitives/pull/925 (added removeLinks)

---

https://github.com/radix-ui/primitives/blob/main/packages/react/focus-scope/src/focus-scope.tsx#L341C1-L343C2

callumacrae:
Hadn't seen that, thanks!

Seems pretty good logic for dialogs where forms are more common, but not sure it works for popovers - we're seeing a lot of inconsistencies throughout our app as we're using them for nested navigation / action dropdowns / export menus / other things where we mix buttons and links but they all look the same

Either way, clearly this PR needs to change as it's breaking sensible behaviour for dialogs - would you accept a prop as in https://github.com/radix-ui/primitives/pull/1796 if i made the same change here?

sadeghbarati:
> would you accept a prop as in https://github.com/radix-ui/primitives/pull/1796 if i made the same change here?

Hi, thanks for the PR btw

It's up to maintainer

zernonia:
@coderabbitai review

zernonia:
Thanks for the PR @callumacrae  and a well explained concerned. As it was originally adopted from Radix UI, I prefer the link to be part of tabbable candidates, it makes behavior much more natural 😁

Merging!

PR #2317, opened 2025-12-02
fix(Combobox): don't focus the trigger element after closing

### LINKED ISSUE #2236: [Bug]: Clicking from one Combobox to another closes both popovers/context window
### Reproduction

https://v4.shadcn-vue.com/#combobox

### Describe the bug

I encountered an annoying issue on the documentation for combobox https://v4.shadcn-vue.com/#combobox. If click one of the combobox which opens the context, but if you click on the next combobox, both context windows closes down. 
the issue does not happen with date picker, even though it also opens context window, but they are independent. 

https://github.com/user-attachments/assets/8d06c5c1-3f41-46a2-896b-607a58271695

### System Info

```bash
Tried on latest chrome and firefox versions
```

### Contributes

- [ ] I am willing to submit a PR to fix this issue
- [ ] I am willing to submit a PR with failing tests

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/2236

It seems like the auto-focusing of the trigger element after closing could cause issues like https://github.com/unovue/reka-ui/issues/2236. 
Also, suppose a user clicks another input element while the combobox is open, that input won't be focused; instead, the focus will quickly shift from the input element back to the combobox trigger element, which I imagine isn't the desired behavior.

### DIFF
--- a/packages/core/src/Combobox/ComboboxContentImpl.vue
+++ b/packages/core/src/Combobox/ComboboxContentImpl.vue
@@ -3,7 +3,7 @@ import type { Ref } from 'vue'
 import type { DismissableLayerEmits, DismissableLayerProps } from '@/DismissableLayer'
 import type { PopperContentProps } from '@/Popper'
 
-import { createContext, useForwardExpose, useForwardProps, useHideOthers } from '@/shared'
+import { createContext, getActiveElement, useForwardExpose, useForwardProps, useHideOthers } from '@/shared'
 import { useBodyScrollLock } from '@/shared/useBodyScrollLock'
 
 export type ComboboxContentImplEmits = DismissableLayerEmits
@@ -26,7 +26,7 @@ export const [injectComboboxContentContext, provideComboboxContentContext]
 </script>
 
 <script setup lang="ts">
-import { computed, onMounted, ref, toRefs } from 'vue'
+import { computed, onMounted, onUnmounted, ref, toRefs } from 'vue'
 import { DismissableLayer } from '@/DismissableLayer'
 import { ListboxContent } from '@/Listbox'
 import { PopperContent } from '@/Popper'
@@ -76,6 +76,13 @@ onMounted(() => {
     if (isInputWithinContent.value) {
       rootContext.inputElement.value.focus()
     }
+  }
+})
+
+onUnmounted(() => {
+  const activeElement = getActiveElement()
+  if (isInputWithinContent.value && (!activeElement || activeElement === document.body)) {
+    rootContext.triggerElement.value?.focus()
   }
 })
 </script>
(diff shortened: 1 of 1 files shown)


markjaniczak:
Would this have a negative effect on keyboard users? Consider this flow:

- Press tab: navigate to the combobox trigger
- Press spacebar: open the combobox menu, focus is set to the first focusable element in the menu
- Press escape: close the combobox menu, focus is lost on the focusable element in the menu

Without explicitly focusing the trigger in this scenario, would focus be lost and returned to the start of the document? If so, that could be extremely frustrating for keyboard users filling out a long form with a combobox.

Almost every variant of menu, dropdown, combobox etc has this problem in combination with each other as they all try and focus their contents on open at the same time another component is focusing its trigger element. Would it work to shift the focus of the content in these components forward by a tick so that it's always performed after?

kricsleo:
> would focus be lost and returned to the start of the document?

No, the focus can still be maintained.
But you're right, I should be more cautious about managing the focus. I've updated it to keep the previous behavior unless a shifted focus is detected.

sadeghbarati (approved):
Thanks MAN, working fine also with animations

https://stackblitz.com/edit/vitejs-vite-v7k19jxr?file=src%2FApp.vue

matt-stow-lyka:
In our custom Combobox, it no longer received focus when you close it with the mouse. Do we need to a the Trigger or Anchor?

```
<ComboboxRoot
    :default-value="defaultValue"
    :data-id="dataId"
    :open-on-click="openOnClick"
    :ignore-filter="ignoreFilter"
    @highlight="$emit('highlight', $event)"
    @update:open="$emit('update:open', $event)"
  >
    <template #default="{ open }">
      <ComboboxInput
        asChild
        :display-value="displayValue"
        :value="modelValue"
        @update:modelValue="$emit('update:modelValue', $event)"
      >
        <LInput
          :id="id ?? name"
          :invalid="invalid"
          :disabled="disabled"
          :name="name"
          :placeholder="placeholder"
          :readonly="readonly"
          v-bind="inputProps"
        />
      </ComboboxInput>
      <LIcon
        :is="ChevronDownIcon"
        :class="{['chevron-icon--open': open && $slots.items && !readonly }]"
        size="md"
      />

      <ComboboxContent
        :class="$slots.items && !readonly ? classes.content() : classes.hidden()"
        role="listbox"
        :aria-label="listboxLabel"
      >
        <ComboboxViewport :class="classes.viewport()">
          <slot v-if="!busy" name="empty"><LComboboxEmptyItem /></slot>
          <slot v-if="$slots.items" name="items" />
          <slot v-if="busy" name="busy"><LComboboxBusyItem /></slot>
        </ComboboxViewport>
      </ComboboxContent>
    </template>
  </ComboboxRoot>
  ```

kricsleo:
@matt-stow-lyka, would you mind creating a minimal reproduction on StackBlitz? 🙏

PR #2321, opened 2025-12-06
feat(Menu): allow more value types

LeBenLeBen:
Currently MenuRadioGroup only accepts strings as model value in terms of typing, but it actually works just fine with other types.

I have cases where I would like to have numbers or just `null` as a value without having to create a proxy that convert the values from and to strings. The current types doesn’t allow me to do so, so here’s a contribution to fix that.


sadeghbarati (approved):
👍

LeBenLeBen:
Would you like me to update the docs as well?

sadeghbarati:
Thanks, not needed cause [`docs:gen`](https://github.com/unovue/reka-ui/blob/v2/package.json#L19) script will take care of it

Issue #1667, opened 2025-02-28
[Bug]: Components not fully functional in Shadow DOM environments


danieldasilva-smake:
### Environment

```bash
Can be provided if needed
```

### Link to minimal reproduction

Can be provided if needed

### Steps to reproduce

See description

### Describe the bug

First of all, thank you for this incredible component library! It is, in my opinion, the best unstyled component library available.  

However, I have noticed that some components do not work correctly when used inside a Shadow DOM. Specifically:  

1. When placing a `Popover` inside a `Dialog`, clicking on the `Popover` content causes the entire `Dialog` to close unexpectedly.  
2. When using an input field inside a `Dialog` that dynamically updates content within the `Dialog`, the focus keeps switching back to `DialogContent`, disrupting the user experience.  

It would be great if the library provided better support for Shadow DOM environments, especially for users leveraging Vue.js’s `defineCustomElement` function, where the entire application runs inside a Shadow DOM.  

### Possible Solution:  
In most cases, the issue arises from direct usage of `document` instead of retrieving the correct root node. A common fix is to use `.getRootNode()` on the element instead of `document`.  

For example, modifying [FocusScope.vue (L125-L133)](https://github.com/unovue/reka-ui/blob/v2/packages/core/src/FocusScope/FocusScope.vue#L125..L133) to use `container.getRootNode()` instead of `document` might (untested) resolve the focus issue with input fields inside a `Dialog`.  

Would love to hear your thoughts on this, and I’d be happy to help test any potential fixes!

lepture:
This is not a bug, it is a feature request.

We can add a default mount point config in `ConfigProvider`. Currently, by default, everything will be mounted to `document.body`.

danieldasilva-smake:
> This is not a bug, it is a feature request.
> 
> We can add a default mount point config in `ConfigProvider`. Currently, by default, everything will be mounted to `document.body`.

Perhaps this is a feature request in the sense that Shadow DOM support is not explicitly mentioned, and I would like to see that support added. However, no changes to configuration or props are necessary for this. You can already define mounting points in the Portal component (e.g., `PopoverPortal` or `DialogPortal`) using the `to` prop.  

The issue is that the wrong root node is being used internally. In some cases, you may also need to access or add event listeners to both the `document` and the Shadow DOM.

MickL:
Having the same problem:
Elements like popper/datepicker are added to the end of body instead of inside my custom component and therefor they dont have any styles!

[...]

PR #2331, opened 2025-12-10
test(focus-scope): shadow root issue example

ldelhommeau:
#1667 

This is a reproduction branch about the focus isse mentionnedn in #1667 :
> When using an input field inside a Dialog that dynamically updates content within the Dialog, the focus keeps switching back to DialogContent, disrupting the user experience.

## What is not working
When rendering a `Dialog`component (or anything that can be teleported) into the shadow root.
Then rendering another component that will update the DOM (here a combobox opening a list) at some point.
If we have the focus on the inner component (e.g. combobox) when the DOM is updated, we loose the focus.

## How to reproduce

- clone the repo
- install dependencies
- run `pnpm story:dev`
- wait for all stories to load
- go on http://localhost:6006/__sandbox.html?storyId=packages-core-src-combobox-story-comboboxshadowroot-story-vue&variantId=packages-core-src-combobox-story-comboboxshadowroot-story-vue-0
- click on the button to open the dialog inside the shadow root
- focus the input inside the dialog (by click on keyboard)
- start typing inside the input
- you should loose the focus

### Expected behaviour
The focus should not be lost.
Here is an example of how it should behave.

> [!NOTE]  
> Here everything is teleported into the body so it's working. It would be nice to have it working inside the shadow-root:

https://github.com/user-attachments/assets/3f13ced8-7fe3-4bf3-a6d5-ff7b0fb3a3ff

### Faulty behaviour
Here, everything is mounted into shadow-root, the portal target as well. So the `Dialog` overlay and content will mount inside the portal-target within shadow root. The `Combobox` will be mounted inside the `Dialog` content, and its list will be rendered inline.
The focus is lost when we start typing and the list is shown:

https://github.com/user-attachments/assets/f39823a1-deb3-4d1e-9f00-e0184f408570

### DIFF
--- a/packages/core/src/Combobox/story/ComboboxShadowRoot.story.vue
+++ b/packages/core/src/Combobox/story/ComboboxShadowRoot.story.vue
@@ -1,0 +1,85 @@
+<script setup lang="ts">
+import type { Component } from 'vue'
+import { createApp, useTemplateRef, watch } from 'vue'
+
+import _Dialog from './_Dialog.vue'
+
+const options = [
+  { name: 'Fruit', children: [
+    { name: 'Apple' },
+    { name: 'Banana' },
+    { name: 'Orange' },
+    { name: 'Honeydew' },
+    { name: 'Grapes' },
+    { name: 'Watermelon' },
+    { name: 'Cantaloupe' },
+    { name: 'Pear' },
+  ] },
+  { name: 'Vegetable', children: [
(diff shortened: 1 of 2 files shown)


zernonia:
Closing this reproduction/test branch to keep the PR list tidy — shadow DOM support continues in #2380 and #1667. Thanks a lot for the detailed investigation!

PR #2336, opened 2025-12-12
fix(useHideOthers): fix edge case with useHideOthers when nested in a a native popover

### LINKED ISSUE #2240: [Bug]: Combobox with `open=true` adds `aria-hidden` on everything else
### Environment

```bash
Reka UI version: 2.6.0
Vue version: latest
```

https://stackblitz.com/edit/wha1wwjp

### Steps to reproduce

1. Run the provided demo
2. Inspect the elements

### Describe the bug

When using Combobox with the `open` option se to true (to always show it open), it adds `aria-hidden=true` and `data-aria-hidden=true` to basically every other element on the page.

I narrowed it down to the `hideOthers` function, so this might be an edge case with that.

While I think I understand the logic behind it (keeping only the Combobox itself visible for screen-readers), it still seems like something that shouldn't happen, especially when used like in the provided example, where the Combobox is placed inside a popover, and thus not being part of the normal document flow

### Expected behavior

`aria-hidden` and `data-aria-hidden` should not get added on other elements

_No response_

plycos:
resolves #2240

this should also fix it for other elements that are using native popover.

## Summary
  Fixes `useHideOthers` applying `aria-hidden` when nested inside a closed native popover.

## Problem
  When a component using `useHideOthers` is inside a native `<div popover>`, it runs even when the popover is closed.

## Solution
  Skip `hideOthers` when inside a closed native popover. This is a simple non-reactive fix. It doesn't track when the popover opens/closes. Could be made reactive if needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed behavior of native popovers to properly manage visibility of surrounding elements when popovers are closed.

### DIFF
--- a/packages/core/src/shared/useHideOthers.ts
+++ b/packages/core/src/shared/useHideOthers.ts
@@ -16,7 +16,8 @@ export function useHideOthers(target: MaybeElementRef) {
     // disable hideOthers on test mode
     if (import.meta.env.MODE === 'test')
       return
-    if (el)
+    // Skip if inside a closed native popover
+    if (el && !el.closest('[popover]:not(:popover-open)'))
       undo = hideOthers(el)
     else if (undo)
       undo()


zernonia:
@coderabbitai review please

PR #2348, opened 2025-12-18
fix(2347): always show select arrow and leave display to the select content

### LINKED ISSUE #2347: [Bug]: SelectArrow disappears before animation has completed
### Environment

```bash
Developement/Production OS: macOS 15.6.1 (24G90)
Node version: 22.14.0
Package manager: npm@10.9.2
Reka UI version: 2.6.1
Vue version: 3.5.25
Client OS: macOS 15.6.1 (24G90)
Browser: Chrome 143.0.7499.110
```

https://stackblitz.com/edit/4trk4hsd

### Steps to reproduce

Open the link provided and click the select to open and close the content

### Describe the bug

When closing the content the `SelectArrow` disappears immediately before the animation completes.

### Expected behavior

The `SelectArrow` should animate with the content of the select.

https://github.com/user-attachments/assets/be16997a-5f91-49ad-a06e-2d6f8cacae48

pmairoldi:
Closes #2347 

The `SelectArrow` was toggled on and off based on if the `rootContext.open.value` is set which means that when the select is closed this values immediately gets set to false and the arrow disappears without animation. I looked at the other components that use the `PopperArrow` and they do not have such behaviour so I removed it.

I don't have insights into if the `rootContext.open.value` is really needed or no. So let me know if there is a reason for it.

Thanks!

### DIFF
--- a/packages/core/src/Select/SelectArrow.vue
+++ b/packages/core/src/Select/SelectArrow.vue
@@ -7,20 +7,18 @@ export interface SelectArrowProps extends PopperArrowProps {}
 <script setup lang="ts">
 import { PopperArrow } from '@/Popper'
 import { injectSelectContentContext, SelectContentDefaultContextValue } from './SelectContentImpl.vue'
-import { injectSelectRootContext } from './SelectRoot.vue'
 
 const props = withDefaults(defineProps<SelectArrowProps>(), {
   width: 10,
   height: 5,
   as: 'svg',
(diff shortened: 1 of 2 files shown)


zernonia:
Thanks for the fix @pmairoldi ! Given that we strictly only allow `SelectArrow` to be within `SelectContent`, this changes LGTM 👍

PR #2359, opened 2025-12-27
fix(DatePicker/DateRangePicker/Calendar/RangeCalendar)!: make `weekStartsOn` locale-independent

### LINKED ISSUE #2157: [Bug]: `weekStartsOn` changes depending on locale
### Environment

```bash
Latest version
```

https://stackblitz.com/edit/ng6hiems?file=src%2FApp.vue

### Steps to reproduce

1. Render a Calendar
2. Try out some locales (en `en-US` and `en-GB`) while keeping the same `weekStartsOn`

### Describe the bug

The start of the week changes even when `weekStartsOn` stays the same

### Expected behavior

The start of the week should stay the same no matter what the locale is

_No response_

### LINKED ISSUE #2264: [Bug]: weekStartsOn does not work correctly with non-English locale
### Environment

```bash
Development OS: Windows 10 19043.1110
Node version: 16.0.0
Package manager: pnpm@8.6.0
Reka UI version: 2.0.0
Vue version: 3.0.0
Locale tested: cs-CZ
Nuxt mode: universal
```

https://stackblitz.com/edit/c5xh3s73?file=src%2FApp.vue

### Steps to reproduce

1. Render RangeCalendarRoot or CalendarRoot and pass weekStartsOn=1 and locale='cs-CZ'.
2. Observe that the calendar grid starts on Tuesday, not Monday.
3. Repeat with weekStartsOn=1 and locale='en-US': calendar starts on Monday as expected.
4. This shows weekStartsOn logic does not work correctly for Czech or other non-English locales.

### Describe the bug

When weekStartsOn is set to 1 (Monday) and locale is set to cs-CZ, the calendar week starts on Tuesday instead of Monday. This does not happen when locale is en-US.

It appears that week grid calculation internally uses getDayOfWeek(date, locale), which maps weekdays differently based on locale. The week arithmetic should be locale-insensitive: weekStartsOn=1 should always mean Monday, regardless of locale.

### Expected behavior

Passing weekStartsOn=1 should always make the week start on Monday, independent of locale. Locale should only affect weekday names, not grid calculation.

_No response_

kricsleo:
resolves https://github.com/unovue/reka-ui/issues/2157
resolves https://github.com/unovue/reka-ui/issues/2264

❗❗This may be a breaking change for those who rely on `weekStartsOn` to behave differently across locales.

As I've seen more users confused by this implicit behavior, I think it's better to define `weekStartsOn` in a fixed and explicit way, where the day of the week is always `Sun(0), Mon(1), ..., Sat(6)`. This also aligns with other date libraries, e.g., [date-fns](https://date-fns.org/docs/startOfWeek), making it more predictable.

However, this would break users who rely on the previous behavior by specifying different `weekStartsOn` values for different locales to make the calendar always start on "Monday" or other days.

For example:

Before: `locale="en-GB" :weekStartsOn="0"` === `locale="en-US" :weekStartsOn="1"` === `"week starts on Monday"`

After: `locale="en-GB" :weekStartsOn="1"` === `locale="en-US" :weekStartsOn="1"` === `"week starts on Monday"`

I've also considered adding a new prop (e.g., `fixedWeekStartsOn: boolean`) to only change the behavior when this prop is set, which could avoid breaking changes. However, this would add complexity to the library, and the default behavior would still remain confusing (and we'd have to document it, etc.). WDYT? cc @zernonia 

## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized day-of-week calculations to consistently use Sunday as the reference point, improving accuracy and consistency when determining specific days of the week across different locales.

### DIFF
--- a/packages/core/src/date/comparators.ts
+++ b/packages/core/src/date/comparators.ts
@@ -143,7 +143,13 @@ export function getLastFirstDayOfWeek<T extends DateValue = DateValue>(
   firstDayOfWeek: number,
   locale: string,
 ): T {
-  const day = getDayOfWeek(date, locale)
+  /**
+   * "firstDayOfWeek" is fixed to 0(Sunday) to avoid confusion regarding locales.
+   * This also aligns with other date libraries, e.g., date-fns.
+   *
+   * #see https://github.com/unovue/reka-ui/issues/2157
+   */
+  const day = getDayOfWeek(date, locale, 'sun')
 
   if (firstDayOfWeek > day)
     return date.subtract({ days: day + 7 - firstDayOfWeek }) as T
@@ -159,7 +165,14 @@ export function getNextLastDayOfWeek<T extends DateValue = DateValue>(
   firstDayOfWeek: number,
   locale: string,
 ): T {
-  const day = getDayOfWeek(date, locale)
+  /**
+   * "firstDayOfWeek" is fixed to 0(Sunday) to avoid confusion regarding locales.
+   * This also aligns with other date libraries, e.g., date-fns.
+   *
+   * #see https://github.com/unovue/reka-ui/issues/2157
+   */
+  const day = getDayOfWeek(date, locale, 'sun')
+
   const lastDayOfWeek = firstDayOfWeek === 0 ? 6 : firstDayOfWeek - 1
 
   if (day === lastDayOfWeek)


zernonia:
@coderabbitai review

zernonia:
@kricsleo  thanks for the PR, indeed this might be breaking changes for some user, but I think it's worth it to make this behavior consistent across locale

PR #2382, opened 2026-01-07
fix: dayPeriod returning incorrect value due to missmatched format

### LINKED ISSUE #2381: [Bug]: dayPeriod method within useDateFormatter returning incorrect value for some locales
### Environment

```bash
Node version: 16.0.0
Package manager: pnpm@8.6.0
Reka UI version: 2.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
Browser: Chrome 90.0.4430.212
```

https://stackblitz.com/edit/bsunze1p?file=src%2FApp.vue

### Steps to reproduce

After setting the locale prop on the `DateFieldRoot` component to `en-AU`  the `AM/PM` day period selector no longer updates after interacting with the keyboard. 

### Describe the bug

When `useDateFormatter` is provided with the `en-AU` locale the resulting value of the `dayPeriod` part type is lowercase "pm". This causes a missmatch with the currently supported conditions resulting in the the `dayPeriod` function only ever returning "AM".

### Expected behavior

The `dayPeriod` function within `useDateFormatter` should support lowercase "pm" within the supported conditions so it returns "PM" for these locales

_No response_

javangriff:
This pull request fixes a bug in the `dayPeriod` function within `useDateFormatter` that causes it to only ever return "AM" for some locales e.g. "en-AU".

When `useDateFormatter` is passed the `en-AU` locale the resulting value of the `dayPeriod` part type is lowercase "pm". This causes the supported conditions to fail and the function to only ever return "AM" for locales where this is the case. This breaks the "AM/PM" day period selection functionality in any component where this is provided e.g. DateField.

I've added the lowercase "pm" to the conditions for now, but it may be better to force the value to either upper or lowercase to ensure the condition is not case-sensitive. Happy to update this PR if this is preferred.

Closes [#2381](https://github.com/unovue/reka-ui/issues/2381)

### DIFF
--- a/packages/core/src/shared/useDateFormatter.ts
+++ b/packages/core/src/shared/useDateFormatter.ts
@@ -105,8 +105,8 @@ export function useDateFormatter(initialLocale: string, opts: DateFormatterOptio
       minute: 'numeric',
     }).formatToParts(date)
     const value = parts.find(p => p.type === 'dayPeriod')?.value
-    // Day period can be "AM"/"PM" or "a.m."/"p.m." in some locales
-    if (value === 'PM' || value === 'p.m.')
+    // Day period can be "AM"/"PM" or "am"/"pm" or "a.m."/"p.m." in some locales
+    if (value === 'PM' || value === 'pm' || value === 'p.m.')
       return 'PM'
 
     return 'AM'


epr3 (approved):
LGTM :+1:

PR #2345, opened 2025-12-16
feat: configurable aria label for NavigationMenuRoot

mariusheine:
Closes #2344

PR #2386, opened 2026-01-09
fix(NavigationMenu): allow `aria-label` override by fixing CollectionSlot attrs forwarding

### LINKED ISSUE #2344: [Feature]: Changing aria-label for NavigationMenuRoot
### Describe the feature

`NavigationMenuRoot` currently sets aria-label hard-coded to "Main" . But having different navigation menu e.g. main and side navigation makes it impossible for screen readers impossible to distinguish them.

My proposal would just be to make `aria-label` a prop with the default value of "Main". This would not break anything.

### Additional information

benjamincanac:
Related to nuxt/ui#5141

The `NavigationMenuRoot` component had a hardcoded `aria-label="Main"` that couldn't be overridden because `CollectionSlot` wasn't forwarding attributes.

### Changes
- Removed hardcoded `aria-label="Main"` from `NavigationMenuRoot`
- Fixed `CollectionSlot` to forward `attrs` to the underlying `Slot` component

### Usage

Users should now provide their own accessible label based on the navigation's purpose, avoiding misleading defaults when used for footer/sidebar navigation:

```vue
<template>
  <NavigationMenuRoot aria-label="Header" />
  <NavigationMenuRoot aria-label="Footer" />
</template>
```

### DIFF
--- a/packages/core/src/Collection/Collection.ts
+++ b/packages/core/src/Collection/Collection.ts
@@ -46,12 +46,13 @@ export function useCollection<ItemData = {}>(options: { key?: string, isProvider
 
   const CollectionSlot = defineComponent({
     name: 'CollectionSlot',
-    setup(_, { slots }) {
+    inheritAttrs: false,
+    setup(_, { slots, attrs }) {
       const { primitiveElement, currentElement } = usePrimitiveElement()
       watch(currentElement, () => {
         context.collectionRef.value = currentElement.value
       })
-      return () => h(Slot, { ref: primitiveElement }, slots)
+      return () => h(Slot, { ref: primitiveElement, ...attrs }, slots)
     },
   })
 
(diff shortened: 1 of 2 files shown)


mariusheine:
@benjamincanac I did the same fix a few weeks ago #2345 but no response yet. Despite that your solution seems more generic for other attributes as well

zernonia:
Thanks @benjamincanac @mariusheine ! Yeah I think this solution works much better and make sure other `CollectionSlot` forwards the attrs too 👍🏻

mariusheine:
@zernonia Then we should maybe also close #2344 by this PR. I will mark mine as duplicated!

benjamincanac:
Oh sorry @mariusheine I completely forgot to check for related open issues / PRs 😬

PR #2390, opened 2026-01-09
fix(Tabs): only render `aria-controls` when TabsContent exists

benjamincanac:
### ❓ Type of change

### 📚 Description

This PR adds content registration tracking to `TabsRoot`, so `TabsTrigger` only renders `aria-controls` when a matching `TabsContent` is actually mounted which results in invalid ARIA references otherwise.

> I'm making this change because I encountered a11y issues in Nuxt UI when using it this way: https://ui.nuxt.com/docs/components/tabs#content

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accessibility by ensuring the aria-controls attribute on tab triggers only appears when corresponding tab content is present.

* **Tests**
  * Added test coverage for aria-controls behavior to verify correct attribute binding based on content availability.

### DIFF
--- a/packages/core/src/Tabs/Tabs.test.ts
+++ b/packages/core/src/Tabs/Tabs.test.ts
@@ -1,7 +1,8 @@
 import type { VueWrapper } from '@vue/test-utils'
-import { mount } from '@vue/test-utils'
+import { flushPromises, mount } from '@vue/test-utils'
 import { beforeEach, describe, expect, it } from 'vitest'
 import { axe } from 'vitest-axe'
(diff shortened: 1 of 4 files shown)


benjamincanac:
@zernonia I'm not 100% sure this is the right way to fix this, let me know if it needs changes!

zernonia:
@coderabbitai review

zernonia:
@benjamincanac  LGTM. Seems like AI also agree 😆

PR #2423, opened 2026-01-29
fix(Calendar/RangeCalendar): when the selected date or today is outside the range of allowed dates, cannot use the tab to select selectable date

duan602728596:
### ❓ Type of change

### 📚 Description

* Describe your changes in detail:
When the selected date or today is outside the allowed selection range, tabindex="0" will be set on the trigger of the first selectable date.   
For example, if the selected date is 2026-01-29, and the allowed selection range for Calendar or RangeCalendar is from 2026-01-26 to 2026-01-27, then setting tabindex="0" will trigger the date 2026-01-26.

* Why is this change required? What problem does it solve?   
When the selected date or today is outside the allowed range of dates, tabindex="0" is still on the trigger of the selected date or today, which will cause the date to be unselectable using tab.
For example:
    * Calendar: https://codesandbox.io/p/devbox/f2h23m   
    * RangeCalendar: https://codesandbox.io/p/devbox/psrpxj

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyboard navigation and tabindex/focus selection across single and range calendars, including placeholder and disabled-date scenarios and min/max bounds.
  * More reliable identification of the first focusable date within month views.

* **Tests**
  * Added tests covering tabindex and focus behavior for single and range calendar scenarios.

### DIFF
--- a/packages/core/src/Calendar/useCalendar.ts
+++ b/packages/core/src/Calendar/useCalendar.ts
@@ -359,15 +359,19 @@ export function useCalendar(props: UseCalendarProps) {
 
   const firstFocusableDate = computed(() => {
     for (const month of grid.value) {
-      for (const row of month.rows) {
-        for (const date of row) {
-          if (!isSameMonth(date, month.value) || isDateDisabled(date) || isDateUnavailable(date))
-            continue
-          return date
-        }
+      if (props.minValue.value && isBefore(month.value, props.minValue.value))
+        continue
+
+      const daysInMonth = getDaysInMonth(month.value)
+      const startDay = props.minValue.value && isSameMonth(props.minValue.value, month.value) ? props.minValue.value.day : 1
+
+      for (let day = startDay; day <= daysInMonth; day++) {
+        const date = month.value.set({ day })
+        if (isDateDisabled(date) || isDateUnavailable(date))
+          continue
+        return date
       }
     }
-    return undefined
   })
 
   return {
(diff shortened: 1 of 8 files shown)


packages/core/src/Calendar/useCalendar.ts
```diff
@@ -336,6 +353,23 @@ export function useCalendar(props: UseCalendarProps) {
 
   const fullCalendarLabel = computed(() => `${props.calendarLabel.value ?? 'Event Date'}, ${headingValue.value}`)
 
+  const isPlaceholderFocusable = computed(() => {
+    return !(isDateDisabled(props.placeholder.value) || isDateUnavailable(props.placeholder.value) || isOutsideVisibleView(props.placeholder.value))
+  })
+
+  const firstFocusableDate = computed(() => {
```
epr3, line 360: The approach looks good here, but there's room for optimisation since I think we can still determine the `firstFocusableDate` without iterating the whole grid.

1. In the first loop (`const month of grid.value`), you should skip the "padding" days and start with the first day of the month because those days shouldn't be focusable anyway: you should iterate the days of the month. (there's the `getDaysInMonth` function which should help).
2. You can also skip more iterations by checking if the `minValue` is in the same month, if true, set the iteration start from the `minValue` and continue from there.

Implementing this approach should only leave you with 2 loops instead of 3 and the second will only iterate what's needed.

epr3 (changes requested):

packages/core/src/Calendar/useCalendar.ts
duan602728596: I have already optimized it according to this.

epr3 (approved):
LGTM 👍🏻

duan602728596:
Preview link after bug fix:
* Calendar: https://codesandbox.io/p/devbox/reka-ui-calendar-tabindex-bug-1-forked-pgvnsj
* RangeCalendar: https://codesandbox.io/p/devbox/reka-ui-range-calendar-tabindex-bug-1-forked-g4r34y

PR #2433, opened 2026-02-03
fix(docs): escape angle brackets in autogen

onmax:
Type strings with `<` and `>` were breaking template parsing in meta files.

Escapes angle brackets to `&lt;` and `&gt;` in autogen output.

## Documentation

* Improved type annotation rendering across all component documentation by adding proper HTML entity escaping for generic brackets and arrows in TypeScript type signatures. This ensures correct display of complex types in auto-generated reference docs without impact to any functional APIs or code behavior.


zernonia:
Yo @onmax ! Just out of curiosity which meta file is broken? 🤔

PR #2477, opened 2026-02-27
fix(Tabs): only force-mount TabsContent during SSR

### LINKED ISSUE #2476: [Bug]: TabsContent hard-codes force-mount with presence
### Environment

```bash
Reka UI version: 2.8.2
Vue version: 3.5.28
CSS framework: tailwindcss@3.4.19
```

https://github.com/dfidler/reka-ui-issue-2476

### Steps to reproduce

Use TabsCount (without setting forceMount) with multiple tabs and then switch between them in an SPA; inspect the DOM, they're all there.

### Describe the bug

TabsContent hardcodes force-mount when calling presence. 

Introduced in commit https://github.com/unovue/reka-ui/commit/95263567b337b2b5b547f7c81db9d5294359c70a - looks like it was intended to hydrate tab content for SSR, but it's  breaking change for SPA.  

### Expected behavior

Detect if we're in SSR or SPA mode and honour force-mount prop on TabsContent.

Code should read `:force-mount="!isBrowser || forceMount"`

dfidler:
Fixes https://github.com/unovue/reka-ui/issues/2476

The Presence component inside TabsContent has force-mount hardcoded, keeping all inactive tab panels in the DOM. You can see this either by inspecting the DOM or when CSS display property (e.g. display:flex) is set on TabsContent, as it overrides the hidden attribute's display:none.

This change conditionally applies force-mount only during SSR (when document is undefined) or when the user explicitly passes the forceMount prop, using the existing isBrowser utility.

### 🔗 Linked issue

#2476

### ❓ Type of change

- [X] 🐞 Bug fix (a non-breaking change that fixes an issue)

### 📚 Description

Detects if running in SSR/SPA; respects the desire to hydrate all TabsContent in SSR but also respects documented behaviour for SPA 

### 📝 Checklist

- [X] I have linked an issue or discussion.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server-side rendering compatibility for tab components. Fixed content mounting behavior to properly handle both server-side and client-side rendering. Tabs now correctly mount and display content in server environments while maintaining user-defined mounting controls in the browser, ensuring consistent rendering across all scenarios.

### DIFF
--- a/packages/core/src/Tabs/TabsContent.vue
+++ b/packages/core/src/Tabs/TabsContent.vue
@@ -18,6 +18,7 @@ export interface TabsContentProps extends PrimitiveProps {
 import { computed, onBeforeUnmount, onMounted, ref } from 'vue'
 import { Presence } from '@/Presence'
 import { Primitive } from '@/Primitive'
+import { isBrowser } from '@/shared/browser'
 import { injectTabsRootContext } from './TabsRoot.vue'
 import { makeContentId, makeTriggerId } from './utils'
 
@@ -48,7 +49,7 @@ onBeforeUnmount(() => {
   <Presence
     v-slot="{ present }"
     :present="forceMount || isSelected"
-    force-mount
+    :force-mount="!isBrowser || forceMount"
   >
     <Primitive
       :id="contentId"


dfidler:
Note:  This is a code-fix only; I have not updated documentation though in retrospect, https://github.com/unovue/reka-ui/commit/95263567b337b2b5b547f7c81db9d5294359c70a probably should have documented that SSR forces `forceMount = true `

I can update the docs in my PR if you want or I can leave the wording of that up to the maintainers.  Happy to take  your guidance.

zernonia:
Nice fix @dfidler ! A quick comment:

- **Potential behavior change for SPA users**: Previously, inactive tab content stayed in the DOM (just hidden via `hidden` attribute). With this change, inactive tabs will unmount in the browser unless `forceMount` is set. This is arguably correct behavior, but might be (maybe minor) breaking change for users relying on DOM persistence (e.g. form state).

dfidler:
> Previously, inactive tab content stayed in the DOM (just hidden via `hidden` attribute). With this change, inactive tabs will unmount in the browser unless `forceMount` is set. This is arguably correct behavior, but might be (maybe minor) breaking change for users relying on DOM persistence (e.g. form state).

Indeed. But that's the documented behaviour.  It's the whole point of running tab content through Presence and it's why forceMount exists.

Though now that I think on it should presence ever unmount something in the SSR world? Was https://github.com/unovue/reka-ui/commit/95263567b337b2b5b547f7c81db9d5294359c70a a red herring and forceMount should just be overridden in the SSR world entirely?

I honestly don't know as I don't live in that world.

But then that turns a really small change into a huge change/risk.

dfidler:
Any update for target release for this fix?

zernonia:
Thanks a lot for digging into this, @dfidler, and for the clear repro!

Unfortunately we can't take this approach. The hardcoded `force-mount` on `TabsContent` is intentional and mirrors Radix: the panel wrapper is always rendered with `hidden`, and only its content is unmounted. Gating it on `isBrowser` causes a few problems (verified locally):

1. **Hydration mismatch.** `isBrowser` differs between server and client, so the server renders every `role="tabpanel"` while the client renders nothing for inactive panels, and Vue reports a hydration mismatch.
2. **`unmountOnHide: false` stops working on the client.** Presence now unmounts the whole panel, so content the user explicitly asked to keep alive is destroyed.
3. **Dangling `aria-controls`.** Inactive triggers point at panel ids that no longer exist in the DOM (what #2390 fixed).

With the default `unmountOnHide: true`, inactive tab *content* is already unmounted; only an empty `<div role="tabpanel" hidden>` remains. The visible issue in #2476 comes from `display: flex` overriding the browser's default `[hidden] { display: none }`. Adding `data-[state=inactive]:hidden` (Tailwind) or `[hidden] { display: none !important }` resolves it.

Closing this for now. Thanks again for the contribution!

PR #2499, opened 2026-03-09
fix(Splitter): emit correct units for pixel-sized panels in events

### LINKED ISSUE #2481: [Bug]: Splitter pixel sizing is broken
### Environment

```bash
OS: Windows 11 Pro 26200.7922
Node version: ^24
Reka UI version: ^2.9.0
Vue version: ^3.5
Browser: Brave 1.87.191 (Chromium 145.0.7632.120)
```

https://stackblitz.com/edit/reka-ui-splitter-reproduction?file=src%2FApp.vue

### Steps to reproduce

1. Visit the reproduction link.
2. Open the browser console.
3. Move panels and review the output in the console.

### Describe the bug

The recently added pixel-based sizing feature introduces some difficulties and unexpected behavior:
1. The `@layout` event (`SplitterGroup`) contains percentage values even if some panels are pixel-sized.
2. The `@resize` event (`SplitterPanel`) contains a percentage value even if the panel is pixel-based, while the `getSize` method returns the value in a correct unit.
3. The `@collapse`/`@expand` events (`SplitterPanel`) are not emitted if the panel is pixel-sized. Also, I guess the `isCollapsed`/`isExpanded` properties are not resolved correctly (can't prove since events are not emitted).

### Expected behavior

1. Document the expected behavior or fix the implementation so that values in correct units are emitted for each panel individually.
2. The correct value is emitted.
3. The events are emitted, and the state is correctly resolved.

Here is the PR introducing the pixel-sizing feature: <https://github.com/unovue/reka-ui/pull/2362>.

zernonia:
## Summary
- **`@layout` event**: Now converts internal percentage values to native units (px) for pixel-sized panels before emitting
- **`@resize` event**: Now emits size in the panel's native unit (px for pixel panels) instead of always emitting percentages
- **`@collapse`/`@expand` events**: Fixed not firing for pixel-sized panels — the root cause was comparing layout sizes (in %) against `collapsedSize` (in px); now both are compared in the same unit
- Used `fuzzyNumbersEqual` for size comparisons to prevent floating-point precision issues

Closes #2481

## Test plan
- [ ] Verify `@layout` event returns pixel values for pixel-sized panels and percentage values for percentage-sized panels
- [ ] Verify `@resize` event returns correct pixel values for pixel-sized panels
- [ ] Verify `@collapse` and `@expand` events fire correctly for pixel-sized collapsible panels
- [ ] Verify `isCollapsed`/`isExpanded` state is correct for pixel-sized panels
- [ ] Verify percentage-only groups still work as before (no regression)
- [ ] Test with the [reproduction from the issue](https://stackblitz.com/edit/reka-ui-splitter-reproduction?file=src%2FApp.vue)

## Summary by CodeRabbit

* **Refactor**
  * Improved panel resizing and collapse/expand behavior when mixing pixel and percentage sizes—layouts now emit and apply native pixel values where appropriate for more consistent sizing across resizes and group-size changes.

* **Tests**
  * Added extensive tests covering resize callbacks, unit conversion, and collapse/expand scenarios to ensure stable, predictable panel behavior.

### DIFF
--- a/packages/core/src/Splitter/SplitterGroup.vue
+++ b/packages/core/src/Splitter/SplitterGroup.vue
@@ -193,6 +193,20 @@ function getPanelDataWithPercentConstraints(groupSizeOverride?: number | null) {
 
 const setLayout = (val: number[]) => layout.value = val
 
+/** Convert internal layout (always in %) to native units for each panel */
+function convertLayoutToNativeUnits(internalLayout: number[]): number[] {
+  const { panelDataArray } = eagerValuesRef.value
+  const groupSize = getGroupSizeInPixels()
+
+  return internalLayout.map((size, index) => {
+    const panelData = panelDataArray[index]
+    if (panelData && (panelData.constraints.sizeUnit ?? '%') === 'px' && groupSize != null) {
+      return (size / 100) * groupSize
+    }
+    return size
+  })
+}
+
 useWindowSplitterPanelGroupBehavior({
   eagerValuesRef,
   groupId,
@@ -347,12 +361,13 @@ watch(() => eagerValuesRef.value.panelDataArrayChanged, () => {
       setLayout(nextLayout)
 
       eagerValuesRef.value.layout = nextLayout
-      emits('layout', nextLayout)
+      emits('layout', convertLayoutToNativeUnits(nextLayout))
 
       callPanelCallbacks(
         panelDataArray,
         nextLayout,
         panelIdToLastNotifiedSizeMapRef.value,
+        getGroupSizeInPixels(),
       )
     }
   }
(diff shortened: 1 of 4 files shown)


niusia-ua:
@zernonia Thank you for a fast fix!

However, I noticed a minor issue. When a pixel-sized panel collapses (for example, with `:collapsed-size="50"`), the actual panel size is not exactly 50px. Instead, it has a tiny tail to either side: 50.00000000019456 or 49.99999999984036, which outputs -1 or 1, respectively, for `fuzzyCompareNumbers`.

Therefore, the `expand`/`collapse` events are still not emitted. But if you set the collapsed size to something like 49.99999999, these events are fired correctly.

PR #2512, opened 2026-03-13
fix(useHideOthers): handle browsers without :popover-open support (Safari 18)

### LINKED ISSUE #2511: [Bug]: v2.9.1 introduced selector that doesn't work in many browsers like Safari 18
### Environment

```bash
Reka UI v2.9.1
```

I'm seeing many errors across a wide range of browsers due to an invalid selector. 

`The string did not match the expected pattern.` for example.

### Describe the bug

https://github.com/unovue/reka-ui/blob/afef410d5df10bbba2dce5cd2514a33a271bfe47/packages/core/src/shared/useHideOthers.ts#L20

Everything is okay in v2.9.0

zernonia:
## Summary

- Fixes a crash introduced in v2.9.1 where `closest('[popover]:not(:popover-open)')` throws `"The string did not match the expected pattern"` in browsers that don't support the `:popover-open` pseudo-class (e.g. Safari 18)
- Wraps the selector call in a `try/catch` so unsupported browsers fall back gracefully and continue to work normally
- The original behavior from the fix in #2336 is preserved for browsers that do support the selector

Closes #2511

## Test plan

- [ ] Verify dialogs/modals still work correctly in Chrome/Firefox (`:popover-open` supported)
- [ ] Verify dialogs/modals no longer throw in Safari 18 (`:popover-open` unsupported)
- [ ] Verify nested-in-native-popover edge case still works in supported browsers

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced browser compatibility for popover functionality on Safari and other browsers with limited CSS selector support.

### DIFF
--- a/packages/core/src/shared/useHideOthers.ts
+++ b/packages/core/src/shared/useHideOthers.ts
@@ -17,7 +17,13 @@ export function useHideOthers(target: MaybeElementRef) {
     if (import.meta.env.MODE === 'test')
       return
     // Skip if inside a closed native popover
-    if (el && !el.closest('[popover]:not(:popover-open)'))
+    // Use try/catch as `:popover-open` pseudo-class is not supported in all browsers (e.g. Safari 18)
+    let isInsideClosedPopover = false
+    try {
+      isInsideClosedPopover = !!el?.closest('[popover]:not(:popover-open)')
+    }
+    catch {}
+    if (el && !isInsideClosedPopover)
       undo = hideOthers(el)
     else if (undo)
       undo()


zernonia:
@perdolique can you help verify the fix? I dont have old safari 😬

npm i https://pkg.pr.new/reka-ui@2512

Perdolique:
@zernonia I have neither, unfortunately 😞 . Just saw huge spikes of problems after upgrading in my website monitoring.

zernonia:
No prob! I believe try/catch would eliminate such warning/error from being logged 😁

Perdolique:
Yeah, at least it will not be reported 😄

PR #2522, opened 2026-03-16
fix(Collapsible): make contentId reactive to fix VoiceOver aria-controls

### LINKED ISSUE #2521: [Bug]: Accordion: VoiceOver does not announce expanded/collapsed state (official documentation example also fails)
### Environment

```bash
Developement/Production OS: macOS Tahoe 26.2
Node version: 24.10.1
Package manager: pnpm@9.15.1
Reka UI version: 2.8.0
Vue version: 3.3.0
Browser: Chrome Version 146.0.7680.80
```

https://reka-ui.com/docs/components/accordion

### Steps to reproduce

1. Enable VoiceOver on macOS or iOS.
2. Open the official Accordion example.
3. Navigate to the trigger using Tab.
4. Press Enter or Space to expand/collapse.

**Actual result:**  
VoiceOver does not announce “expanded” or “collapsed”.

**Expected result:**  
VoiceOver should announce the state when aria-expanded changes.

### Describe the bug

I’m using the Accordion component from Reka UI and noticed that VoiceOver (macOS/iOS) does not announce the “expanded” or “collapsed” state when interacting with the trigger, even though:

- the trigger is a <button>
- aria-expanded correctly toggles between true and false
- focus is placed on the button
- the panel has role="region" and aria-labelledby
- no aria-label is used that would override the accessible name

This behavior does not match the expected WAI-ARIA accordion pattern.

### The official documentation example also does not work
To rule out issues in my own implementation, I tested the minimal example from the official documentation:

[Reka UI accordion](https://reka-ui.com/docs/components/accordion)

Even when using that example without any modifications, VoiceOver does not announce the **expanded/collapsed** state.
The screen reader only reads the button’s accessible name, but never announces the state, even though `aria-expanded` updates correctly in the DOM.

This indicates that the issue is not in my code, but in the current implementation of `AccordionTrigger` or the DOM structure generated by Reka UI.

### Actual behavior
VoiceOver never announces the expanded/collapsed state.

This happens:
- in my implementation
- and also in the official documentation example

### Expected behavior

- When focusing the trigger:
“Section title. Button. Collapsed.”
- When expanding: “Expanded.”
If aria-live is used, VoiceOver should also read the panel content. 
The user should be able to navigate the content immediately.
- When collapsing:
“Collapsed.”

This is the expected VoiceOver behavior for an ARIA-compliant accordion.

_No response_

zernonia:
## Summary

- `contentId` in `CollapsibleRootContext` was a plain `string`, not a reactive `Ref<string>`, so when `CollapsibleContent` set the ID on mount, `CollapsibleTrigger`'s `aria-controls` was never reactively updated
- On initial render `aria-controls=""` (empty), meaning VoiceOver had no valid element to associate with the button's expanded/collapsed state — causing it to not announce state changes
- Changed `contentId` to `Ref<string>` across `CollapsibleRoot`, `CollapsibleContent`, and `CollapsibleTrigger`

Fixes #2521

## Test Plan
- [ ] Open an accordion with VoiceOver enabled on macOS/iOS — verify "expanded" / "collapsed" is announced on toggle
- [ ] Verify `aria-controls` on the trigger correctly references the content panel's `id` from initial render
- [x] All existing Accordion and Collapsible tests pass

## Summary by CodeRabbit

* **Refactor**
  * Updated internal component context API for improved reactivity handling in the Collapsible component.

### DIFF
--- a/packages/core/src/Collapsible/CollapsibleContent.vue
+++ b/packages/core/src/Collapsible/CollapsibleContent.vue
@@ -32,7 +32,7 @@ const props = defineProps<CollapsibleContentProps>()
 const emits = defineEmits<CollapsibleContentEmits>()
 
 const rootContext = injectCollapsibleRootContext()
-rootContext.contentId ||= useId(undefined, 'reka-collapsible-content')
+rootContext.contentId.value ||= useId(undefined, 'reka-collapsible-content')
 
 const presentRef = ref<InstanceType<typeof Presence>>()
 const { forwardRef, currentElement } = useForwardExpose()
@@ -102,7 +102,7 @@ useEventListener(currentElement, 'beforematch', (ev) => {
   >
     <Primitive
       v-bind="$attrs"
-      :id="rootContext.contentId"
+      :id="rootContext.contentId.value"
       :ref="forwardRef"
       :as-child="props.asChild"
       :as="as"
--- a/packages/core/src/Collapsible/CollapsibleRoot.vue
+++ b/packages/core/src/Collapsible/CollapsibleRoot.vue
@@ -1,7 +1,7 @@
 <script lang="ts">
 import type { Ref } from 'vue'
 import type { PrimitiveProps } from '@/Primitive'
-import { toRefs } from 'vue'
+import { ref, toRefs } from 'vue'
 import { createContext, useForwardExpose } from '@/shared'
 
 export interface CollapsibleRootProps extends PrimitiveProps {
@@ -21,7 +21,7 @@ export type CollapsibleRootEmits = {
 }
 
 interface CollapsibleRootContext {
-  contentId: string
+  contentId: Ref<string>
   disabled?: Ref<boolean>
   open: Ref<boolean>
   unmountOnHide: Ref<boolean>
@@ -58,8 +58,10 @@ const open = useVModel(props, 'open', emit, {
 
 const { disabled, unmountOnHide } = toRefs(props)
 
+const contentId = ref('')
+
 provideCollapsibleRootContext({
-  contentId: '',
+  contentId,
   disabled,
   open,
   unmountOnHide,
--- a/packages/core/src/Collapsible/CollapsibleTrigger.vue
+++ b/packages/core/src/Collapsible/CollapsibleTrigger.vue
@@ -22,7 +22,7 @@ const rootContext = injectCollapsibleRootContext()
     :type="as === 'button' ? 'button' : undefined"
     :as="as"
     :as-child="props.asChild"
-    :aria-controls="rootContext.contentId"
+    :aria-controls="rootContext.contentId.value"
     :aria-expanded="rootContext.open.value"
     :data-state="rootContext.open.value ? 'open' : 'closed'"
     :data-disabled="rootContext.disabled?.value ? '' : undefined"


eBrou:
This defect is still present in **2.10.3**, and I think the retraction of #2521 closed this PR on
a false negative.

#2521 was retracted because the reporter's VoiceOver settings had been reset — so the *reported
symptom* was user error. But the code defect described in this PR's own summary is independently
real, and it's observable without any screen reader:

```js
// after hydration, Collapsible or Accordion
document.querySelector('button[aria-controls]').getAttribute('aria-controls') // ""
```

Measured on reka-ui 2.10.3 / Vue 3.5.40 / Nuxt 4.5.1:

| Component | SSR | After hydration |
| --- | --- | --- |
| Collapsible | attribute absent | `aria-controls=""` |
| Accordion | attribute absent | `aria-controls=""` |
| Popover | attribute absent | `aria-controls=""` |
| Dialog | absent | absent (fine — no dangling ref) |
| Tabs | absent | resolves correctly |

Two things worth adding to what's already in the summary:

**Accordion inherits this from Collapsible.** `AccordionContent` re-exports `CollapsibleContent`,
and `AccordionTrigger`/`AccordionItem` add no ids of their own. So #2521 being an *Accordion*
report and this being a *Collapsible* fix were the same bug — which is part of why the retraction
was misleading.

**Collapsible and Accordion aren't covered by the sibling work.** #2597 and #2774 address
Select/Combobox/Popover/DropdownMenu; neither mentions Collapsible or Accordion. So if #2774
lands as-is, this case survives.

On severity: `aria-expanded` is present and correct throughout, so state stays programmatically
determinable and axe-core 4.11 doesn't flag it. But a commenter on #2774 reports 4.12.1 raising a
violation for persistent `aria-controls=""`, and `html-validate` already flags the dangling case —
so this is a "passes today, fails on the next tooling bump" problem rather than a live a11y break.

### I verified this patch fixes it on 2.10.3

Rather than test the PR build (which is from March), I applied this PR's change by hand to an
installed **2.10.3** — the same three edits, `contentId: ''` to `ref('')` plus `.value` at the
two read sites — and re-ran the same check:

| Component | 2.10.3 as shipped | 2.10.3 + this patch |
| --- | --- | --- |
| Collapsible | `aria-controls=""` | `aria-controls="reka-collapsible-content-v-0"` — resolves |
| **Accordion** | `aria-controls=""` | `aria-controls="reka-collapsible-content-v-2"` — resolves |
| Popover | `aria-controls=""` | unchanged, still `""` (out of scope for this PR) |
| Dialog | absent | unchanged |
| Tabs | resolves | unchanged |

Two things worth noting from that run:

- **Zero hydration warnings** in dev with `__VUE_PROD_HYDRATION_MISMATCH_DETAILS__` on. SSR still
  omits the attribute and the client fills it in, and Vue does not treat that as a mismatch — so
  the reactive `contentId` doesn't introduce an SSR/client divergence.
- The Accordion id resolves to `reka-collapsible-content-*`, which is a nice confirmation that
  Accordion is Collapsible underneath and that fixing Collapsible fixes both.

So the change is confined to the intended components and behaves correctly on current code.

Would you consider reopening? Happy to open a fresh PR with the same three edits against `main`
instead if that's easier than reviving this one.

Issue #2543, opened 2026-03-25
[Bug]: Tooltip does not close when a scrollable ancestor is scrolled


stillgg:
### Environment

```bash
Developement/Production OS: macOS Tahoe 26.3.1
Node version: v24.14.0
Package manager: pnpm@10.13.1
Reka UI version: 2.9.2
Vue version: 3.5.17
CSS framework: tailwindcss@3.3.3
Client OS: macOS Tahoe 26.3.1
Browser: Chrome 146.0.7680.153
```

### Link to minimal reproduction

https://stackblitz.com/edit/6xmmkjej?file=src%2FApp.vue

### Steps to reproduce

1. Open dialog
2. Hover over the trigger - tooltip opens
3. Scroll the container vertically so the trigger moves away from the cursor

### Describe the bug

`TooltipRoot` does not close when a scrollable parent container is scrolled. The tooltip
stays visible and floats at the position where the trigger used to be, visually detached
from it.
Two browser mechanics combine to produce this:
**1. No `pointerleave` on scroll**
When a scrollable container is scrolled, the trigger element moves away from the cursor.
The browser does **not** fire `pointerleave` in this case — because the *pointer* did not
move, the *element* did. So `TooltipRoot`'s internal hover tracking never detects that
the cursor left the trigger.
**2. `TooltipPortal` is outside the scrollable DOM tree**
`TooltipPortal` renders into `document.body` via `<Teleport>`. Scroll events from the
container do not propagate there, so the portal content is completely unaware of the scroll.

### Expected behavior

The tooltip should close when any scrollable ancestor of the trigger is scrolled,
matching standard UI expectations (and the behavior of native browser tooltips).

stillgg:
@kricsleo please could you take a look at the problem?

PR #2544, opened 2026-03-25
fix(tooltip,popover,hovercard): close on scrollable ancestor scroll

stillgg:
### 🔗 Linked issue

Resolves [#2543]

### ❓ Type of change

### 📚 Description

`TooltipRoot`, `HoverCardRoot`, and `PopoverRoot` did not close their floating content
when a scrollable ancestor container was scrolled. The tooltip/card would stay open and
visually float at its original position, detached from the trigger element.

**Root cause:** two browser mechanics combine to produce the bug:
1. The browser does not fire `pointerleave` when the *element* moves out from under the
   cursor due to scroll — only when the *pointer* moves.
2. `TooltipPortal` renders into `document.body` via `<Teleport>`, so scroll events from
   the ancestor container never reach it.

**Fix:** added a `closeOnAncestorScroll` prop to all three components. When enabled, the
component subscribes to `scroll` events on every overflow ancestor of the trigger element
(found via `getOverflowAncestors` from `@floating-ui/dom`, which is already a direct
dependency) and closes the floating element when any of them fires.

| Component | Default |
|---|---|
| `TooltipRoot` | `true` — tooltip is hover-only, closing on scroll matches user expectation |
| `HoverCardRoot` | `true` — same reasoning as Tooltip |
| `PopoverRoot` | `false` — popover is opened by click and often contains interactive content; unexpected scroll-close would be destructive |

Also fixed `HoverCardTrigger` which was replacing the entire `triggerElement` ref object
in context (`rootContext.triggerElement = currentElement`) instead of setting its value
(`rootContext.triggerElement.value = ...`), which prevented the `watch` in `HoverCardRoot`
from reacting to the trigger being mounted.

### 📝 Checklist

## Summary by CodeRabbit

* **New Features**
  * Tooltip, HoverCard, and Popover gain a closeOnAncestorScroll option to control dismiss-on-scroll (Tooltip/HoverCard default on, Popover default off).
  * Added a dialog example demonstrating scrollable items each with tooltips that can remain open while scrolling.

* **Tests**
  * Added test suites validating open/close behavior, runtime prop toggling, and listener cleanup when ancestor scrolling occurs.

### DIFF
--- a/packages/core/src/Dialog/story/DialogWithScrollableTooltip.story.vue
+++ b/packages/core/src/Dialog/story/DialogWithScrollableTooltip.story.vue
@@ -1,0 +1,171 @@
+<script setup lang="ts">
+import { Icon } from '@iconify/vue'
+import {
+  DialogClose,
+  DialogContent,
+  DialogOverlay,
+  DialogPortal,
+  DialogRoot,
+  DialogTitle,
+  DialogTrigger,
+} from '..'
+import {
+  TooltipArrow,
+  TooltipContent,
+  TooltipPortal,
+  TooltipProvider,
+  TooltipRoot,
+  TooltipTrigger,
+} from '../../Tooltip'
+
+const items = Array.from({ length: 20 }, (_, i) => ({
+  id: i + 1,
+  name: `Item ${i + 1}`,
+  description: `Description for item ${i + 1}`,
+}))
+</script>
+
+<template>
+  <Story
+    title="Dialog/With Scrollable Tooltip"
+    :layout="{ type: 'single', iframe: false }"
+  >
+    <Variant title="closeOnAncestorScroll (default: true)">
+      <div class="flex items-center justify-center p-8">
+        <TooltipProvider>
+          <DialogRoot>
+            <DialogTrigger
+              class="text-violet11 shadow-blackA7 hover:bg-mauve3 inline-flex h-[35px] items-center justify-center rounded-[4px] bg-white px-[15px] font-medium leading-none shadow-[0_2px_10px] focus:shadow-[0_0_0_2px] focus:shadow-black focus:outline-none"
+            >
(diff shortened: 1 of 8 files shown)


zernonia:
Thanks for the detailed work here — the root cause analysis and test coverage are excellent. However, I'm going to close this in favour of a simpler fix in #2557. Here's why:

## Why #2557 instead

**The bug is a one-liner.** `TooltipContentImpl` already had the correct scroll handler and the right `target.contains(trigger)` logic — it just lacked `{ capture: true }`. Without capture mode, `scroll` events from inner overflow containers never reach the `window` listener because `scroll` doesn't bubble. Adding `{ capture: true }` fixes Tooltip entirely. HoverCard gets the same pattern added.

**Comparison:**

| | This PR (#2544) | #2557 |
|---|---|---|
| Lines changed | ~300+ | 9 |
| New props | `closeOnAncestorScroll` on 3 components | None |
| New dependency usage | `getOverflowAncestors` | None |
| Listener count | N listeners (one per overflow ancestor) | 1 listener per open floating element |
| API surface to maintain | Yes | No |

**No library exposes a prop for this.** Base UI, Radix React, and Floating UI all treat scroll-close as unconditional behaviour with no opt-out. There's no known valid use case for keeping a tooltip open while it detaches from a scrolling trigger — and if someone needs that, they can manage `open` externally.

**The `closeOnAncestorScroll: false` escape hatch adds risk.** Opting out produces a broken visual state (floating element detached from anchor) that we'd implicitly be supporting.

The `HoverCardTrigger` ref fix (`triggerElement.value = ...` instead of replacing the ref object) in this PR is valid and appreciated — I've folded that understanding into the new approach as well.

Again, great investigation — the cause analysis in the PR description is spot-on and helped identify the real fix.

PR #2546, opened 2026-03-26
fix(FocusScope): don't move focus if DOM mutation occurred before any nodes had focus

### LINKED ISSUE #2506: [Bug]: Dialog demo doesn't focus first input on subsequent open by default
### Environment

```
Reka UI v2.9.1

macOS Sequoia 15.7.4
Microsoft Edge 145.0.3800.97 (Official build) (arm64)
1Password for Edge extension 8.12.4.46

Safari 26.3 (20623.2.7.18.1)
1Password for Safari extension 8.12.5.2
```

https://stackblitz.com/edit/y5qqkahm?file=src%2FApp.vue

### Steps to reproduce

1. Go to minimal repro: https://stackblitz.com/edit/y5qqkahm?file=src%2FApp.vue
2. Open dialog first time. Name field is focused.
3. Close dialog.
4. Open dialog once again.

### Describe the bug

Noticed that Dialog demo doesn't focus first input (Name) by default when you open dialog second time. Instead `DialogContent` container itself is focused. 

I did some digging and it seems to be some strange interaction between `Icon` component from iconify and `MutationObserver` in `FocusScope`. The following is happening: `Icon` component sets CSS classes -> MutationObserver is triggered -> focus gets re-validated but moved to container instead. If you remove `Icon` component then focus behaves correctly.

Of course it's not just demo's issue and can be reproduced in apps.

**UPDATE March 12** 
This seems to be caused by 1Password browser extension. Disabling this extensions restore correct focus behavior.

### Expected behavior

First input (tabbable element) should be focused on dialog open every time.

_No response_

romansp:
DOM mutation may be triggered by something unrelated to focus, e.g. class changed on an element, so first check if we even have a previously focused element. Also ensure that captured DOM mutations even have removed nodes before restoring focus.

Add test to ensure focus is restored to container on node removal.

### 🔗 Linked issue

resolves #2506 

### ❓ Type of change

### 📚 Description

Did some debugging locally with 1Password extension enabled and noticed that MutationObserver callback might be called even before first element is focused and `lastFocusedElementRef` is updated. For example, `iconify` sets class on SVGs. In this case `lastFocusedElementRef` is still `null` and `container.contains` check returns false and focus is move to container.

Additionally updated FocusScope tests as these didn't properly await `waitFor` results and thus didn't even verify expected results. 

Here's with 1Password extension enabled and mentioned fix. First tabbable element focused correctly each time the dialog is opened.

https://github.com/user-attachments/assets/1d7c217c-275f-4ee1-873f-ac7c8f5d61e9

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved focus handling so containers reliably receive and retain focus when previously focused elements are removed; avoids errors when there was no prior focus and skips adjustments if no DOM removals occur.

* **Tests**
  * Added regression test for dialogs to autofocus the first tabbable field when opened with a dropdown trigger present.
  * Simplified and hardened focus assertions and added a test for removed-focused-element behavior.

### DIFF
--- a/packages/core/src/FocusScope/FocusScope.test.ts
+++ b/packages/core/src/FocusScope/FocusScope.test.ts
@@ -1,8 +1,8 @@
 import type { RenderResult } from '@testing-library/vue'
 import userEvent from '@testing-library/user-event'
-import { render, waitFor } from '@testing-library/vue'
+import { render } from '@testing-library/vue'
 import { beforeEach, describe, expect, it, vi } from 'vitest'
-import { defineComponent } from 'vue'
+import { defineComponent, nextTick } from 'vue'
 import { FocusScope } from '.'
 import { DialogContent, DialogRoot, DialogTitle, DialogTrigger } from '../Dialog'
 import { SelectContent, SelectItem, SelectRoot, SelectTrigger, SelectValue } from '../Select'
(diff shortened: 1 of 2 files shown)


romansp:
Updated: added test case from #2551 to cover #2550. cc @kricsleo

PR #2568, opened 2026-04-03
fix(TimeField,DateField): avoid changing focus prematurely with previous segment value of 0

### LINKED ISSUE #2567: [Bug]: TimeField changes focus prematurely when previous segment value was 0
### Environment

```bash
Reka UI version: 2.8.0
```

https://stackblitz.com/edit/vitejs-vite-ms9dg4r5?file=src%2FApp.vue

### Steps to reproduce

1. Type the value 0 in any segment of the TimeField, then change focus either manually or automatically
2. Go back to the previous segment and type any number other than 0

### Describe the bug

If previous segment has an already set value of 0, when trying to modify it, inputing any number other than 0 will move the focus to the next segment even if the value is not yet complete.

### Expected behavior

_No response_

_No response_

zomakus:
### 🔗 Linked issue

Fixes [#2567](https://github.com/unovue/reka-ui/issues/2567)

### ❓ Type of change

### 📚 Description

In `updateHour` and `updateMinuteOrSecond` functions the `props.lastKeyZero` value was not being resetted when changing focus, it has been fixed so that context is always clean when starting to work on a segment.

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed focus state management in date and time input fields to prevent premature segment switching and ensure proper focus handling when users enter zero values during keyboard input.

* **Tests**
  * Added comprehensive test coverage validating focus behavior and segment navigation across date and time input fields when entering zero values.

### DIFF
--- a/packages/core/src/DateField/DateField.test.ts
+++ b/packages/core/src/DateField/DateField.test.ts
@@ -289,6 +289,28 @@ describe('dateField', async () => {
     }
   })
 
+  it('doesn\'t change focus prematurely with segment value of 0', async () => {
+    const { getByTestId, user, day, month, year } = setup({
+      dateFieldProps: {
+        modelValue: zonedDateTime,
+        granularity: 'second',
+      },
+    })
+    const { hour, minute, second } = getTimeSegments(getByTestId)
+
+    const segments = [month, day, year, hour, minute, second]
+
+    for (const segment of segments) {
+      await user.click(segment)
+      await user.keyboard('{0}')
+      await user.keyboard(kbd.TAB)
+      expect(segment).not.toHaveFocus()
+      await user.click(segment)
+      await user.keyboard('{1}')
+      expect(segment).toHaveFocus()
+    }
+  })
+
   it(`preserve other segment when one segment's value is deleted`, async () => {
     const { user, day, month, year } = setup({
       dateFieldProps: {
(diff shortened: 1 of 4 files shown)


zernonia:
Thanks for the fix! The root cause diagnosis is correct and the changes in `updateHour` and `updateMinuteOrSecond` are the right approach.

One thing worth addressing in this same PR: `updateDayOrMonth` has the identical `hasLeftFocus` reset block but is missing the same `lastKeyZero` reset. The bug is reproducible there too — type `0` in a month/day segment, navigate away, return, then type a non-zero digit and focus jumps prematurely.

The fix would be one additional line at [useDateField.ts ~L368](https://github.com/unovue/reka-ui/blob/v2/packages/core/src/shared/date/useDateField.ts#L368):

```ts
if (props.hasLeftFocus.value) {
  props.hasLeftFocus.value = false
  props.lastKeyZero.value = false  // add this
  prev = null
}
```

Also, a regression test covering the scenario (segment value is `0` → leave → return → type non-zero → verify focus stays) would help prevent this from regressing. The existing test style in `TimeField.test.ts` and `DateField.test.ts` should be easy to follow.

zomakus:
I've added the test cases and extended the fix to the DateField component as requested.

While adding the tests I noticed a related issue, when using a 12 hour locale the hour segment doesn't change focus when pressing the 0 key more than once, it doesn't seem like its intended... Since the root cause has nothing to do with the current issue I'll address it in another PR if thats ok.

zernonia:
LGTM! Thanks @zomakus !

PR #2581, opened 2026-04-06
fix(TimeField): change focus after pressing 0 more than once on hour segment with 12 hour locales

### LINKED ISSUE #2580: [Bug]: TimeField doesn't change focus when pressing the 0 key on 12 hour locales
### Environment

```bash
Reka UI version: 2.9.4
```

https://reka-ui.com/docs/components/time-field

### Steps to reproduce

1. Press 0 in the hour segment repeatedly

### Describe the bug

When pressing the 0 key repeatedly in the hour segment, the focus never changes to the next segment (only happens with 12 hour locales). This is inconsistent with the behaviour of the minute and second segments, where once a value has been set, pressing another digit automatically changes focus to the next segment.

### Expected behavior

_No response_

_No response_

zomakus:
### 🔗 Linked issue

Fixes https://github.com/unovue/reka-ui/issues/2580

### ❓ Type of change

### 📚 Description

Previously on `updateHour` 12 AM and PM were being handled as their display value (i.e. 12) but this was an issue because pressing 0 repeatedly would not change the focus to the next segment. I updated the `handleHourSegmentKeydown` function to make sure they were treated as 0 internally by the `updateHour` function.

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed time field keyboard input so repeated "0" keystrokes correctly advance focus through hour, minute, second, and AM/PM segments.
* **Tests**
  * Added a test covering repeated "0" input to validate focus progression across time segments.


zomakus:
I reverted the previous fix to avoid the issues mentioned by CodeRabbit, and used a more localized approach instead. The `props.lastKeyZero.value` logic in both `updateMinuteOrSecond` and `updateHour` seems to be completely redundant since `prev === null` will never trigger while `props.lastKeyZero.value` is set to true (this is probably a leftover from copy/pasting the `updateDayOrMonth` function where internal value is in fact null when pressing 0). I left it as is just in case but it can probably be removed.

P.S. Please don't mind the typos in the commit messages, I've been a bit dyslexic today.

PR #2610, opened 2026-04-24
feat(Popper): add dir prop for `RTL/LTR` support

malik-jouda:
### ❓ Type of change

### 📚 Description
* **New Features**
  * Added support for specifying reading direction (LTR/RTL) for popper components.
  * Popper positioning, transform origin and arrow alignment now adjust dynamically based on text direction to ensure correct placement in bidirectional and multilingual layouts.
  

### 📝 Checklist

## Summary by CodeRabbit

* **New Features**
  * Popper now supports explicit reading-direction (RTL/LTR) control; content and wrapper reflect the direction.
  * Arrow alignment and transform-origin calculations adapt to the specified direction for correct horizontal positioning.

* **Tests**
  * Added RTL/LTR coverage and tests verifying direction propagation and direction-aware positioning behavior.

### DIFF
--- a/packages/core/src/Popper/Popper.test.ts
+++ b/packages/core/src/Popper/Popper.test.ts
@@ -1,7 +1,12 @@
 import type { VueWrapper } from '@vue/test-utils'
 import { mount } from '@vue/test-utils'
 import { beforeEach, describe, expect, it } from 'vitest'
+import { defineComponent, h } from 'vue'
 import Popper from './_Popper.vue'
+import PopperAnchor from './PopperAnchor.vue'
+import PopperContent from './PopperContent.vue'
+import PopperRoot from './PopperRoot.vue'
+import { transformOrigin } from './utils'
 
 describe('give default Popper', async () => {
   let wrapper: VueWrapper<InstanceType<typeof Popper>>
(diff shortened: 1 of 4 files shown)


zernonia (changes requested):
Thanks for working on RTL support for Popper 🙏 The X/Y split in `transformOrigin` is the right idea, but I think the feature is incomplete as-is and I'd like to request a change before merging.

The `dir` prop currently only influences the **animation origin**, not the actual **placement**. Floating UI decides RTL alignment-flipping from the computed `direction` of the *floating wrapper* element (`isRTL` → `getComputedStyle(floating).direction`, where `floating` is our `floatingRef` wrapper). But this PR sets `:dir` only on the inner `<Primitive>` and feeds the prop solely into `transformOrigin`.

So the two can disagree — e.g. `<PopperContent dir="rtl" />` inside an LTR document positions with LTR alignment (wrapper inherits `ltr`) but animates from the RTL origin, landing on one side while scaling from the opposite corner.

Could you also bind `:dir="dir"` on the `floatingRef` wrapper so Floating UI's `isRTL` and `transformOrigin` stay consistent and the prop actually drives placement?

```vue
<div
  ref="floatingRef"
  data-reka-popper-content-wrapper=""
  :dir="dir"
  ...
>
```

Minor follow-ups:
- Reuse the shared `Direction` type in `utils.ts` instead of inlining `'ltr' | 'rtl'`.
- A test covering the RTL flip (and placement/origin consistency) would be great, since the snapshot is the only thing exercising it right now.

malik-jouda:
@zernonia 
Thanks for the review! Requested changes have been applied.

PR #2628, opened 2026-05-06
feat(Number Field): add startingValue prop for increment/decrement

rikzwarthoff:
### 🔗 Linked issue
(https://github.com/unovue/reka-ui/issues/2627)

### ❓ Type of change

### 📚 Description

**Added new prop 'startingValue' to Number Field**
When no value is present and startingValue is given, increment/decrement will fill the startingValue. If startingValue is not present it will fallback to previous behavior, min value or zero. 

See the issue for an useful example.

### 📝 Checklist

## Summary by CodeRabbit

* **New Features**
  * Number fields gain an optional starting value. When empty, the first increment or decrement begins from this configured starting value instead of the minimum or zero.

* **Tests**
  * Added test coverage ensuring increment/decrement sequences respect the configured starting value, including clearing and subsequent operations.

### DIFF
--- a/packages/core/src/NumberField/NumberField.test.ts
+++ b/packages/core/src/NumberField/NumberField.test.ts
@@ -91,6 +91,22 @@ describe('numberField', () => {
     expect(input.value).toBe('0')
   })
 
+  it('should start at starting value on increment/decrement, when no value present', async () => {
+    const { input, increment, decrement } = setup({ step: 3, startingValue: 40 })
+    expect(input.value).toBe('')
+
+    await userEvent.click(increment)
+    expect(input.value).toBe('40')
+    await userEvent.click(increment)
+    expect(input.value).toBe('43')
+
+    await userEvent.clear(input)
+    await userEvent.click(decrement)
+    expect(input.value).toBe('40')
+    await userEvent.click(decrement)
+    expect(input.value).toBe('37')
+  })
+
   it('should increase and decrease based on keyboard navigation on input', async () => {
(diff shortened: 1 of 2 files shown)


zernonia:
The new test passes only because the test harness accidentally disables step-snapping — so it isn't actually validating the increment/decrement behavior under real defaults.

The story component used by the tests (`story/_NumberField.vue`) uses a bare `defineProps<NumberFieldRootProps>()` without `withDefaults`, so `stepSnapping` arrives as `undefined` and resolves falsy inside `clampInputValue`. With snapping effectively off, `clampInputValue(40)` returns `40`.

But the component's real default is `stepSnapping: true`. Under that default, `clampInputValue(40)` with `step: 3` snaps to the nearest step and returns **39**, not 40. I confirmed this locally — forcing `stepSnapping: true` makes the test fail:

```
AssertionError: expected '39' to be '40'
```

So as written, `startingValue` is silently altered by step-snapping and the test doesn't catch it.

Could we add an explicit case that pins this down, e.g.:

```ts
it('should snap startingValue to the nearest step when stepSnapping is on', async () => {
  const { input, increment } = setup({ step: 3, startingValue: 40, stepSnapping: true })
  await userEvent.click(increment)
  expect(input.value).toBe(/* whatever the intended behavior is */)
})
```

This forces a decision on the intended behavior:
- if `startingValue` should be used verbatim, clamp to min/max but skip the step-snap (and update the test to expect `40`);
- if snapping is intended, document it on the prop and assert `39`.

Either way the test should run with `stepSnapping: true` so it exercises the default code path.

PR #2636, opened 2026-05-12
fix(DatePicker): reset time after clearing value

### LINKED ISSUE #2599: [Bug]: <DatePickerRoot> overwrites time with the previously selected value
### Environment

```bash
Developement/Production OS: MacOS Tahoe 26.4 (25E246)
Node version: v20.12.0
Package manager: pnpm v8.15.6
Reka UI version: 2.9.6
Vue version: 3.5.32
Client OS: MacOS Tahoe 26.4
Browser: Chrome 146.0.7680.178 (arm64)
```

[CodeSandbox](https://codesandbox.io/p/devbox/rekaui-issue-pc4rfl)

### Steps to reproduce

1. Set the `DatePickerRoot` prop `granularity="minute"` (or any value that enables time selection)
2. Create a `ref` with a date object from `@internationalized/date` (`CalendarDateTime` or `ZonedDateTime`) with a non-zero time value
3. Bind it to `DatePickerRoot` using `v-model`
4. Clear the date programmatically (set it to `undefined` or `null`)
5. Select any date via the `DatePickerCalendar`
6. Observe that the time is restored to the initial value from step 1, instead of being reset to `00:00:00`

### Describe the bug

`DatePickerRoot` unexpectedly restores the previously selected time value instead of resetting it to 00:00:00

See reproduction link.

### Expected behavior

`DatePickerRoot` emits an `update:modelValue` event with a date object where the time is set to the minimum value (00:00:00)

_No response_

yan-ad:
### 🔗 Linked issue

Closes #2599

### ❓ Type of change

### 📚 Description

`DatePickerRoot` could preserve a previously selected time after its model value was cleared. When a user selected a new date through the calendar with time granularity enabled, the calendar emitted a value based on the placeholder that still contained the old time.

This resets the time fields to `00:00:00` when selecting a date while the current model value is empty, and adds a regression test covering the clear-then-select flow.

N/A

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * DatePicker now normalizes time to 00:00 when selecting a date after clearing the previous selection; zoned values keep their time zone while time components are reset.
* **Tests**
  * Added tests verifying time reset for calendar and zoned values after clearing, and that typing preserves user-entered hour/minute values instead of using stale placeholders.

[Review Change Stack](https://app.coderabbit.ai/change-stack/unovue/reka-ui/pull/2636?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/src/DatePicker/DatePicker.test.ts
+++ b/packages/core/src/DatePicker/DatePicker.test.ts
@@ -252,7 +252,7 @@ describe('datePicker', async () => {
     expect(calendar.querySelector('[data-selected]')).not.toBeInTheDocument()
   })
 
-  it('resets time when selecting a date after the model value is cleared', async () => {
+  it('resets stale placeholder time when selecting a date after the model value is cleared', async () => {
     const emittedValues: (DateValue | undefined)[] = []
     const { user, trigger, getByTestId, rerender } = setup({
       datePickerProps: {
@@ -285,7 +285,7 @@ describe('datePicker', async () => {
     expect((selectedValue as CalendarDateTime).millisecond).toBe(0)
   })
 
-  it('resets ZonedDateTime time when selecting a date after the model value is cleared', async () => {
+  it('resets stale ZonedDateTime placeholder time when selecting a date after the model value is cleared', async () => {
     const emittedValues: (DateValue | undefined)[] = []
     const zonedValue = toZoned(new CalendarDateTime(1980, 1, 20, 12, 30, 45, 123), 'America/New_York')
     const { user, trigger, getByTestId, rerender } = setup({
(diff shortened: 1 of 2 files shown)


zernonia:
Nice catch on the root cause. Two things to address before merging:

**1. Regression risk in `DatePickerField`** — `onDateChange` is also called from `DatePickerField.vue:35`. After this change, clearing `modelValue` and then typing a date+time into the field will silently zero out the user's typed time. Prefer fixing the stale `placeholder` directly (e.g. reset its time in the `watch(modelValue, …)` when value becomes `undefined`) so the calendar emits clean dates and the field path stays untouched. Please also add a test for the type-into-field-after-clear flow.

**2. Redundant `.copy()`** — `date.set(...)` already returns a new instance, so `resetTime(date).copy()` clones twice. Drop the `.copy()`.

yan-ad:
Addressed, thanks! Moved the time reset to the stale placeholder path when `modelValue` is cleared, so typed field input keeps the user’s time. Also removed the redundant `.copy()` and added a regression test for typing date+time after clearing.

PR #2639, opened 2026-05-12
fix(Popper): memoize content subtree updates

### LINKED ISSUE #2415: [Bug]: When using position 'popper' on Combobox, many rerenders of PopperContent happen on open, causing noticable lag.
### Environment

```bash
Node version: 16.0.0
Package manager: pnpm@8.6.0
Reka UI version: 2.7.0
Vue version: 3.5.26
Browser: Chrome 90.0.4430.212
```

https://stackblitz.com/edit/vitejs-vite-kntgpp6y?file=src%2FApp.vue

### Steps to reproduce

1. Create a Combobox.
2. Use prop position="popper" on the inner ComboboxContent. 
3. Record the action of opening popper/dropdown in chrome devtools performance tab.

### Describe the bug

When using the prop position="popper" for ComboboxContent, which utilizes floating ui under the hood, there are multiple useless rerenders of the content(PopperContent component to be exact). This is causing every single nested option in the dropdown to be rerendered as well. This causes a noticeable delay, the longer the list gets.

Screenshot of devtools with list of 60 items. Note each 'column' of events. Most of which on rerenders:

Image

Here is the view with app.config.devtools enabled. You can see the PopperContent rerendering multiple times, and every single individual nested item going through its entire render again as well:

Image

It looks like the popper is getting recalculated positions from floating ui, then not properly re-using the content.

### Expected behavior

I would expect any layout shift or recalculation to re-use the content of the popper in a more efficient manner.

_No response_

yan-ad:
### 🔗 Linked issue

Resolves #2415

### ❓ Type of change

### 📚 Description

This PR reduces unnecessary rerenders for `PopperContent` consumers, specifically the `ComboboxContent` `position="popper"` case described in #2415.

When Floating UI recalculates popper placement, the wrapper needs to update positioning/style metadata, but the inner content subtree does not need to rerender unless the props that affect it actually change. This memoizes the inner `Primitive` subtree so heavy slot content, such as large combobox option lists, is not repeatedly rerendered during popper position updates.

A regression test was added for `ComboboxContent position="popper"` with 60 options to ensure the option slot content does not rerender excessively after opening and running popper position updates.

Not applicable.

### 📝 Checklist

## Summary by CodeRabbit

* **Tests**
  * Added tests validating combobox popper behavior and slot-render stability, ensuring filtering updates visible options while keeping re-renders within expected bounds.

* **Refactor**
  * Introduced targeted memoization to popper-positioned content and combobox items to reduce unnecessary re-renders and improve responsiveness.


zernonia:
Nice fix direction! Two concerns though:

1. **`v-memo` caches the entire subtree, including slot vnodes.** So when deps don't change, dynamic slot content goes stale. e.g. typing in a Combobox to filter options doesn't change `placedSide`/`placedAlign`/`isPositioned`, so users would see stale options. The added test only opens the popper — could you add one that filters while open?

2. **`$attrs` in the deps is a no-op.** `instance.attrs` is a stable proxy reference, so `Object.is` never sees it change. Since `PopperContent` has `inheritAttrs: false` and forwards attrs to the inner Primitive, dynamic attrs go stale too — e.g. `ComboboxContentImpl`'s `:data-empty` won't update on filter. Either spread attr values (`...Object.values($attrs)`), or move the memo closer to the heavy consumer (e.g. inside Combobox).

Also `toBeLessThanOrEqual(2)` could be exact, and the fake `ResizeObserver` skips the auto-update loop this is actually optimizing 🙂

yan-ad:
Addressed, thank you! I moved the memoization deps so filtering while the popper is open updates the rendered options, included dynamic attr values in the memo dependencies, and added coverage for filtering while open. I also updated the ResizeObserver mock so the test exercises the auto-update path more closely.

PR #2665, opened 2026-05-28
fix: handle IME composition correctly across input components

benjamincanac:
### 🔗 Linked issue

Fixes IME composition handling for CJK users across all input components (related to #2541).

### ❓ Type of change

### 📚 Description

CJK users experienced broken input across multiple components because intermediate IME composition events were processed as final input. This caused premature focus shifts in PinInput, corrupted segments in DateField, bypassed validation in NumberField/ColorField, and incorrect filtering in Combobox/Listbox/DropdownMenu.

**Changes:**

- Introduces a shared `useComposing` composable (`packages/core/src/shared/useComposing.ts`) that encapsulates `isComposing` state with correct `nextTick` timing on `compositionend`
- **PinInput**: guards `handleInput` during composition, processes final value + advances focus only after commit
- **DateField/TimeField/DateRangeField/TimeRangeField**: guards `handleSegmentKeydown` against `isComposing`/`keyCode 229`, adds `handleSegmentCompositionEnd` that replays committed digits as synthetic keydowns
- **ColorField/NumberField**: guards `@beforeinput` handler so `preventDefault()` doesn't block IME intermediate characters
- **TagsInput**: refactored to `useComposing`, guards `handleInput` so delimiters during composition don't create tags
- **Autocomplete**: refactored to `useComposing` (previously had inline composition handling from #2541)
- **ComboboxInput**: adds composition guards to both `handleInput` and `handleKeyDown`
- **ListboxFilter/DropdownMenuFilter**: defers `highlightFirstItem()`/search updates until composition ends

All affected components have new IME composition test cases (202 tests passing).

### 📝 Checklist

## Summary by CodeRabbit

* **New Features**
  * Unified IME composition handling across inputs: composition keystrokes are ignored until commit; composed text is applied on commit and distributed across segmented inputs where applicable.
* **Bug Fixes**
  * Prevents premature validation, beforeinput blocking, list/filter openings, or focus moves during active composition.
* **Tests**
  * Added comprehensive IME composition tests covering many input components.


zernonia:
Thanks @benjamincanac, really solid work here 🙌 The `useComposing` extraction is clean and the test coverage is great. Nice job already knocking out the test + PinInput multi-char comments. Three things I'd still like to sort before merge:

1. **ComboboxInput** — `@keydown.down.up.prevent` runs `preventDefault()` before your `isComposing` early-return, so arrow keys (used for IME candidate nav) still get swallowed mid-composition. Mind dropping `.prevent` and calling `ev.preventDefault()` inside the handler only when `!isComposing.value`?

2. **ListboxFilter** — `modelValue` is set before the `isComposing` guard, so consumer-side filtering still reacts to intermediate text. Could we defer it to `compositionend` like the other two, or note why Listbox keeps it live?

3. **useDateField** — replaying digits via `handleSegmentKeydown` keeps hitting the original segment, so a multi-digit commit like `"34"` can land in the wrong slot. Dispatching to `document.activeElement` each loop would follow focus.

Also — did you get to test on real Safari + a CJK IME? The `nextTick` reset relies on `input` firing before `compositionend`, which some Safari versions flip. jsdom won't catch that.

Tiny nit: the new `describe` blocks say `'iME'` instead of `'IME'`.

Thanks again, happy to re-review once these are in 🙏

benjamincanac:
@zernonia I've applied the suggestions and also ran some manual testing on Safari and everything seems to be working fine now (I might have missed some test cases though).

zernonia (approved):

PR #2667, opened 2026-05-28
fix(NavigationMenu): behavior should match radix ui navigation menu w…

dbzx10299:
…hen entering another trigger while menu is open

### ❓ Type of change

- [ x] 🐞 Bug fix (a non-breaking change that fixes an issue)
- [ x] 👌 Enhancement (improving an existing functionality like performance)

### 📚 Description

In Radix UI, when entering a new trigger while the viewport is visible, there is no delay. In Reka UI, there is quite a visible delay which is especially apparent when going back and forth between triggers rapidly due to a hardcoded 150ms value.

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented accidental/flicker closes when switching between open navigation triggers so menus remain open and responsive while moving between items.
  * Improved timing so opening or switching menu items updates immediately while still honoring intended close delays when leaving content.

* **Tests**
  * Added timer-based tests to ensure debounce and close behavior are reliable in real scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/unovue/reka-ui/pull/2667?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/core/src/NavigationMenu/NavigationMenu.test.ts
+++ b/packages/core/src/NavigationMenu/NavigationMenu.test.ts
@@ -137,6 +137,11 @@ describe('given default NavigationMenu', () => {
 
     const findLinkContent = () => wrapper.find('[data-dismissable-layer]')
 
+    async function useRealDebounceFn() {
+      const { useDebounceFn: realUseDebounceFn } = await vi.importActual<typeof import('@vueuse/core')>('@vueuse/core')
+      vi.mocked(useDebounceFn).mockImplementation(realUseDebounceFn)
+    }
+
     it('should open menu on click by default', async () => {
       const button = findTriggerButton()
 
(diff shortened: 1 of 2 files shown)


zernonia:
Thanks for this! The approach works for the main hover-switch case, but a few things before merge:

1. **Lint is failing** — the new lines use double quotes and a same-line `if`/brace style that violate the repo's ESLint config (6 errors). Running `pnpm lint --fix` should sort it.

2. **`skipNextClose` can get stuck and prevent the menu from closing.** It's only cleared on consume or in `onTriggerLeave`, but `onContentLeave` also schedules a close without resetting it. Repro: open trigger A via keyboard (no trigger-leave fires) → hover trigger B (`skipNextClose = true`, never consumed) → move into B's content and back out → the content close gets swallowed and the menu stays open. Clearing the flag wherever a close is initiated (or scoping it to just the pending trigger-leave close) would fix it.

3. **Tests:** the suite passes, but only because `useDebounceFn` is mocked synchronously, so the new timing/`skipNextClose` logic isn't actually exercised. A regression test for "switching triggers keeps menu open" + "leaving content closes menu" would help lock this in.

Minor: worth keeping the `// passing 'undefined' resets the debounce timer` comment since `onContentEnter` still relies on that.

dbzx10299:
Alright, I did my best to implement your suggestions. I think I understand what you meant with point 2, I could not reproduce it though.

You will have to forgive me for the tests, I did my best but tests are something I am working on improving since they are not my strong point.

PR #2673, opened 2026-06-03
feat(Toast): allow action without dismissing toast

### LINKED ISSUE #2626: [Feature]: Action button should not always close Toast
### Describe the feature

As describe in https://github.com/nuxt/ui/issues/6424,
when we want to have an interaction with a toast using an action button, toast is automatically closed event preventing default behavior.

Currently, ToastAction wraps ToastClose which make the action button closing the toast.

I guess, we need to add an option to prevent the auto closing.

### Additional information

yan-ad:
### 🔗 Linked issue

Resolves #2626

### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)

### 📚 Description

Adds a `closeOnClick` prop to `ToastAction` so action buttons can run without always dismissing their toast. The default remains `true`, preserving the existing dismiss-on-click behavior.

This also documents the new prop and adds a regression test covering `closeOnClick: false` keeping the toast open after clicking the action.

N/A

### 📝 Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a closeOnClick prop for toast actions to control whether clicking an action dismisses the toast (default: true).

* **Changes**
  * Toast actions now render as buttons by default.

* **Documentation**
  * Updated Toast docs with a Vue example and clarified action dismissal and keyboard behavior.

* **Tests**
  * Added tests verifying clicks both dismiss and preserve the toast based on closeOnClick.

### DIFF
--- a/docs/content/docs/components/toast.md
+++ b/docs/content/docs/components/toast.md
@@ -339,16 +339,14 @@ When providing an icon (or font icon), remember to label it correctly for screen
       keys: ['Space'],
       description: `
         <span>
-          When focus is on a <Code>ToastAction</Code> or
-          <Code>ToastClose</Code>, closes the toast
+          When focus is on a <Code>ToastAction</Code>, closes the toast when <Code>closeOnClick</Code> is <Code>true</Code>. When focus is on <Code>ToastClose</Code>, closes the toast.
         </span>`
     },
     {
       keys: ['Enter'],
       description: `
         <span>
-          When focus is on a <Code>ToastAction</Code> or
-          <Code>ToastClose</Code>, closes the toast
+          When focus is on a <Code>ToastAction</Code>, closes the toast when <Code>closeOnClick</Code> is <Code>true</Code>. When focus is on <Code>ToastClose</Code>, closes the toast.
         </span>`
     },
     {
--- a/docs/content/docs/components/toast.md
+++ b/docs/content/docs/components/toast.md
@@ -139,6 +139,20 @@ An action that is safe to ignore to ensure users are not expected to complete ta
 
 When obtaining a user response is necessary, portal an ["AlertDialog"](/docs/components/alert-dialog) styled as a toast into the viewport instead.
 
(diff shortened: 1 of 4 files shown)


zernonia:
Thanks for the PR @yan-ad ! Two notes before merge:

**Add a default-path test.** The new test only covers `closeOnClick: false`. There's currently no test asserting that a `ToastAction` click closes the toast by default, and this PR forks the render into `v-if`/`v-else` — so the `ToastClose` branch is now untested. A sibling test (default `closeOnClick` → toast closes, `open` becomes `false`) would close that gap cheaply.

**Optional — collapse the render fork.** Instead of `ToastClose` vs `Primitive`, a single `Primitive` with a conditional click avoids duplicating the `:type` logic and the nested `ToastAnnounceExclude`:

```vue
<ToastAnnounceExclude v-if="altText" :alt-text="altText" as-child>
  <Primitive
    :ref="forwardRef"
    :as="as"
    :as-child="asChild"
    :type="as === 'button' ? 'button' : undefined"
    @click="closeOnClick ? rootContext.onClose() : undefined"
  >
    <slot />
  </Primitive>
</ToastAnnounceExclude>
```

This needs `injectToastRootContext()` in `ToastAction` (currently it leans on `ToastClose` for that). Trade-off is it stops reusing `ToastClose`, but removes the branch and the double-wrapping. Not blocking — your call on whether reuse or flatness reads better.

yan-ad:
Resolved both notes — added the default-path `ToastAction` test and also collapsed the render fork into a single `Primitive` with conditional close handling.

yan-ad:
Thanks for the notes! I added the default-path test for `ToastAction`, and also went ahead with the optional simplification so `ToastAction` now uses a single `Primitive` path with conditional close handling.

zernonia (approved):
Thanks for the quick turnaround @yan-ad, this looks good! Both earlier notes are addressed and the default close path is covered.

Only thing left is a small rebase: `Toast.test.ts` conflicts with the viewport-listener cleanup test from #2809 (just keep both tests). Optional nit: add a `<script setup>` with a `retry` stub to the docs snippet so it's copy-pasteable. Happy to merge after the rebase.

PR #2678, opened 2026-06-06
fix(DismissableLayer): keep body pointer-events locked when nested layer closes (#2674)

### LINKED ISSUE #2674: [Bug]: Dropdown inside dialog strips pointer-events style from body
### Environment

```bash
Reka UI version: 2.9.9
Vue version: latest
Browser: Chrome 149
```

https://stackblitz.com/edit/dkrbq9li-ztmjkhkr?file=src%2FApp.vue

### Steps to reproduce

- Add a dropdown menu to a dialog.
- Open dialog -> body receives "pointer-events:none" style
- Open menu inside dialog.
- Close menu -> body is stripped of  "pointer-events:none" style

### Describe the bug

Opening a dropdown menu inside a dialog then closing the menu again conflicts with the pointer-events styles set by the dialog.

### Expected behavior

The `pointer-events:none` style should remain on the body tag as long as the dialog remains open.

_No response_

zernonia:
Closes #2674

## Problem

Opening a `DropdownMenu` (or any modal layer) inside a `Dialog` and then **closing the menu** strips the `pointer-events: none` style from `<body>`, even though the Dialog is still open. This breaks the modal pointer-event lock while the dialog remains visible.

## Root cause

`DismissableLayer` manages the body lock in a `watchEffect`:

```ts
watchEffect((cleanupFn) => {
  if (props.disableOutsidePointerEvents) {
    if (context.layersWithOutsidePointerEventsDisabled.size === 0) { ... }
    context.layersWithOutsidePointerEventsDisabled.add(layerElement.value)
  }
  ...
})
```

Because `context` is `reactive()`, reading `context.layersWithOutsidePointerEventsDisabled.size` registers a reactive dependency, so the effect **re-runs whenever any layer is added/removed** from the set — across *all* mounted layers.

When the nested menu closes, the still-open Dialog's effect re-runs, and its cleanup sees `size === 1` and restores the body's original `pointer-events`, removing the lock prematurely.

Radix React avoids this by using `useEffect` with a fixed dependency list (`[node, ownerDocument, disableOutsidePointerEvents, context]`) — it does not re-run on set mutations.

## Fix

Convert the `watchEffect` to a `watch` with explicit sources (`layerElement`, `disableOutsidePointerEvents`). Reads of the set size inside the callback no longer create reactive dependencies, so the effect only re-runs when its real inputs change — matching the React behavior.

## Tests

Added two tests under `DismissableLayer.test.ts`:
- nested layer closing keeps `body` `pointer-events: none` while the outer layer stays open
- closing the last layer correctly restores `body` `pointer-events`

All existing Dialog / Menu / Popover / DropdownMenu / AlertDialog tests pass.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for nested dismissable layers, validating pointer-event behavior across multiple layers during open/close sequences and prop changes.

* **Refactor**
  * Improved pointer-event management for dismissable layers so body pointer-events are reliably set and restored as layers toggle, ensuring consistent interaction when multiple layers are present.

### DIFF
--- a/packages/core/src/DismissableLayer/DismissableLayer.test.ts
+++ b/packages/core/src/DismissableLayer/DismissableLayer.test.ts
@@ -31,6 +31,9 @@ describe('nested layers with disableOutsidePointerEvents (#2674)', () => {
   function mountNested() {
     const outerOpen = ref(true)
     const innerOpen = ref(false)
+    // Mirrors how a modal Menu drives the prop (`menuContext.open.value`):
+    // the layer stays mounted while the prop toggles back to `false`.
+    const innerDisable = ref(true)
 
     const wrapper = mount(defineComponent({
       setup() {
@@ -39,13 +42,13 @@ describe('nested layers with disableOutsidePointerEvents (#2674)', () => {
             ? h(DismissableLayerPrimitive, { 'disableOutsidePointerEvents': true, 'data-testid': 'outer' }, () => 'Outer')
             : null,
           innerOpen.value
-            ? h(DismissableLayerPrimitive, { 'disableOutsidePointerEvents': true, 'data-testid': 'inner' }, () => 'Inner')
+            ? h(DismissableLayerPrimitive, { 'disableOutsidePointerEvents': innerDisable.value, 'data-testid': 'inner' }, () => 'Inner')
             : null,
         ])
       },
     }), { attachTo: document.body })
 
-    return { wrapper, outerOpen, innerOpen }
+    return { wrapper, outerOpen, innerOpen, innerDisable }
   }
 
   it('should keep body pointer-events none after a nested layer closes while outer stays open', async () => {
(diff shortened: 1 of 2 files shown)


zernonia:
Thanks @coderabbitai — both points addressed in ba7cd63:

1. **Stale set entry on prop toggle** (valid): the watch cleanup now removes the layer from `layersWithOutsidePointerEventsDisabled` and restores the body when the set is empty *after* deletion (making the restore independent of cleanup ordering). This matters because a modal `Menu` binds `:disable-outside-pointer-events="menuContext.open.value"`, so the prop goes `true → false` while the layer is still mounted — previously leaving a stale entry that would prevent the body from re-locking.

2. **Toggle regression tests** (nitpick): added two tests — one for `true → false` while nested (body stays `none` for the still-open outer layer), and one full `true → false → true` cycle on a single layer asserting the body restores and then re-locks. Both fail on the previous implementation.

PR #2687, opened 2026-06-09
feat(HoverCard): add `enableTouch` prop for touch devices

### LINKED ISSUE #1969: [Bug]: Add touch support option for mobile devices
### Environment

```bash
- Operating System: Darwin
- Node Version:     v22.12.0
- Nuxt Version:     3.17.4
- CLI Version:      3.25.1
- Nitro Version:    2.11.12
- Package Manager:  pnpm@10.11.0
```

https://codesandbox.io/p/devbox/reka-ui-hovercard-forked-877wgd?file=/src/App.vue

### Steps to reproduce

- Click on the trigger of a HoverCard on mobile
- Observe that the card does not open

### Describe the bug

We're using HoverCard in Nuxt UI with a mode prop that switches between Popover and HoverCard. When using hover mode (HoverCard), it works perfectly on desktop but can't be triggered on mobile devices.
We understand this is by design since HoverCard is meant for mouse hover interactions, but it would be helpful to have an option to enable touch support so we can maintain our unified mode API while still providing mobile functionality.
Currently, users can't open hover-mode popovers on mobile devices, which breaks the user experience.
Would it be possible to add an optional touch fallback mechanism to HoverCard?

### Expected behavior

_No response_

Relate issues:
- https://github.com/nuxt/ui/issues/2346
- https://github.com/unovue/reka-ui/issues/802

benjamincanac:
### 🔗 Linked issue

Resolves #1969 (also covers nuxt/ui#2346)

### ❓ Type of change

### 📚 Description

`HoverCard` ignores touch input by design (ported from Radix), so it can't be opened on mobile. The trigger only reacts to real pointer hover (`pointerenter`/`pointerleave`, both wrapped in `excludeTouch`) and `focus`, neither of which reliably fires on a tap.

This adds an opt-in `enableTouch` prop on `HoverCardRoot` (defaults to `false`, so existing/Radix behavior is unchanged). When enabled, tapping the trigger toggles the card open/closed:

- Handled on `pointerup` and gated to `pointerType === 'touch'`, so mouse/hover behavior is untouched and a scroll gesture (which fires `pointercancel`, not `pointerup`) won't accidentally open it.
- Opens immediately via `onOpenChange(true)` / closes via `onDismiss()`. `DismissableLayer` defers its own dismiss to the `click` event on touch, which runs after our `pointerup` and resolves idempotently (no close-then-reopen race).

It's kept opt-in rather than always-on because always enabling it would override HoverCard's intentional "supplemental, non-interactive, not-for-touch" semantics and would conflict when the trigger is a real link (tap would both navigate and toggle).

Also includes a small correctness fix: `handleDismiss` now clears the pending open timer, so a dismiss can't be reopened by a lingering focus/hover open timer (a latent bug on desktop too).

### 📝 Checklist

## Summary by CodeRabbit

* **New Features**
  * HoverCard now supports touch interaction via the new `enableTouch` prop. When enabled, touch taps on the trigger toggle the card open/closed (disabled by default).

* **Documentation**
  * Updated HoverCardRoot documentation to describe the `enableTouch` prop and its behavior.

### DIFF
--- a/docs/content/meta/HoverCardRoot.md
+++ b/docs/content/meta/HoverCardRoot.md
@@ -17,6 +17,13 @@
     'default': 'false'
   },
   {
+    'name': 'enableTouch',
+    'description': '<p>When <code>true</code>, tapping the trigger on touch devices toggles the hover card open/closed. By default touch interactions are ignored to match pointer hover semantics.</p>\n',
+    'type': 'boolean',
+    'required': false,
+    'default': 'false'
+  },
+  {
     'name': 'open',
     'description': '<p>The controlled open state of the hover card. Can be binded as <code>v-model:open</code>.</p>\n',
     'type': 'boolean',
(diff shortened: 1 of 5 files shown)


nicooprat:
Could we add the same prop for Tooltips?

PR #2688, opened 2026-06-09
fix(SelectTrigger): prevent suppression of mousedown event

### LINKED ISSUE #1773: [Bug]: SelectTrigger captures clicks so that event listeners on window don't run, breaking other frameworks.
### Environment

```bash
Node version: 22.14.0
Package manager: npm@10.9.2
Reka UI version: 2.1.1
Vue version: 3.5.14
CSS framework: tailwindcss@4.0.0
Browser: All browsers impacted
```

https://stackblitz.com/edit/zug7xfs6-kdfcu2ef?file=src%2FApp.vue

### Steps to reproduce

1. Add a "mousedown" event listener to the window.
2. Click on a SelectTrigger to open the select.

### Describe the bug

Many frameworks add pointer event listeners to the window to manage things like closing popups when the user clicks away from them. If the user happens to open a popup, then click on a SelectTrigger to open a select, the popup that was previously opened will not close as expected, because the SelectTrigger stops further event execution. 

I was testing this framework out, I consider this a major usability bug. So I will be looking elsewhere.

### Expected behavior

EXPECTED: The "mousedown" event listener fires.
ACTUAL: The "mousedown" event listener does not fire.

_No response_

rastuhacode:
### 🔗 Linked issue

Resolves #1773

### ❓ Type of change

#### Issue
`SelectTrigger` opens the select on `pointerdown` and was calling `event.preventDefault()` there to stop the trigger from taking focus away from the highlighted item in the listbox.

Per the [Pointer Events spec](https://www.w3.org/TR/pointerevents3/), calling `preventDefault()` on `pointerdown` cancels the compatibility mouse events that follow — `mousedown`, `mouseup`, and `click`.

#### Fix
Focus prevention was moved from `pointerdown` to `mousedown`:

- `pointerdown` — still opens the select, but no longer calls `preventDefault()`, so the browser emits `mousedown`.
- `mousedown` — calls `preventDefault()` under the same conditions (left click, no Ctrl) to keep the trigger from stealing focus.

Global mousedown listeners (including capture-phase ones on window) now fire before the trigger’s own mousedown handler runs.

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Select trigger interactions to prevent unwanted focus and accidental openings (better handling of left-click vs touch and modifier-key clicks).
* **Tests**
  * Added mouse-interaction tests ensuring correct focus behavior, event propagation, and opening behavior across click/pointer sequences.

### DIFF
--- a/packages/core/src/Select/Select.test.ts
+++ b/packages/core/src/Select/Select.test.ts
@@ -38,6 +38,69 @@ describe('given default Select', () => {
     expect(selectTrigger.attributes('data-placeholder')).toBe('')
   })
 
+  describe('trigger mouse interop', () => {
+    async function openSelectWithMouseClick() {
+      const button = wrapper.find('button')
+      // Open on pointerdown, then emit the compatibility mouse events that follow in browsers.
+      await button.trigger('pointerdown', { button: 0, ctrlKey: false })
+      fireEvent.mouseDown(button.element, { button: 0, ctrlKey: false })
+      fireEvent.mouseUp(button.element, { button: 0, ctrlKey: false })
+      fireEvent.click(button.element, { button: 0, ctrlKey: false })
+      await nextTick()
+      await nextTick()
+    }
+
+    it('should not suppress window mousedown listeners when opening (#1773)', async () => {
+      const button = wrapper.find('button').element
+      const onWindowMousedown = vi.fn()
+      window.addEventListener('mousedown', onWindowMousedown, true)
+
+      fireEvent.pointerDown(button, { button: 0, ctrlKey: false, pointerType: 'mouse' })
+      const mousedownEvent = new MouseEvent('mousedown', {
+        bubbles: true,
+        cancelable: true,
+        button: 0,
+        ctrlKey: false,
+      })
+      button.dispatchEvent(mousedownEvent)
+
+      expect(onWindowMousedown).toHaveBeenCalled()
+      expect(mousedownEvent.defaultPrevented).toBe(true)
+
+      window.removeEventListener('mousedown', onWindowMousedown, true)
+    })
+
+    it('should focus the trigger on click without a preceding pointerdown', async () => {
+      const trigger = wrapper.find('[role="combobox"]').element as HTMLElement
+      const focusSpy = vi.spyOn(trigger, 'focus')
+
+      fireEvent.click(trigger, { button: 0, ctrlKey: false })
+
+      expect(focusSpy).toHaveBeenCalled()
+      focusSpy.mockRestore()
+    })
+
+    it('should not re-focus the trigger on click after opening via pointerdown', async () => {
+      const trigger = wrapper.find('[role="combobox"]').element as HTMLElement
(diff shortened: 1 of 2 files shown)


zernonia (changes requested):
Thanks for tracking this down — the direction is right. Rerouting `preventDefault()` from `pointerdown` to `mousedown` correctly lets the browser emit the compat `mousedown` so window-level listeners fire again (#1773), while still suppressing the trigger's focus. Reka's own outside-dismiss uses `pointerdown.capture`, so it's unaffected. A few things to address before merge:

### 1. The `@click` comment is now wrong — and the handler's behavior actually changes (main concern)

```js
// ...because we are preventing default in `onMouseDown` so effectively
// this only runs for a label 'click'
(event?.currentTarget as HTMLElement)?.focus();
```

The original reasoning relied on `preventDefault()` on `pointerdown` cancelling the `click` compat event — that's *why* `@click` only ran for label clicks. But `preventDefault()` on **`mousedown` does not cancel `click`**. So after this change `@click` fires on every direct mouse click on the trigger (not just label clicks) and calls `currentTarget.focus()` each time.

In practice this is probably benign (the native `click` fires synchronously during the gesture, while the content mounts and grabs focus on the next tick), but it's timing-dependent and the comment now asserts the opposite of what's true. Please correct the comment and verify (manually + ideally a test) that after a direct mouse click, focus lands on the content/active item and not the trigger — especially in Safari.

### 2. `isCtrlLeftClick` name is inverted

```js
function isCtrlLeftClick(event: MouseEvent) {
  return event.button === 0 && event.ctrlKey === false
}
```

The name reads as "Ctrl+Left click," but it returns `true` only when Ctrl is **not** held. Suggest `isPlainLeftClick` / `isLeftClickWithoutCtrl`.

### 3. No test coverage

This is a focus/event-handling change driven by a real interop bug. Please add a regression test — e.g. a `window` `mousedown` listener fires when the trigger is clicked (the #1773 repro), and focus does not remain on the trigger after opening via mouse.

### Minor
- Title typo: "supression" → "suppression" (also in the new `@mousedown` comment).

rastuhacode:
Hey, thanks for the quick and insightful review! I really appreciate it.

I've confirmed your comment about `@click` event. Didn't know before about such events quirks and I needed to make more research initially, my bad.

As for the fix:
1) After some research, test cases and thinking, I've decided to just make a flag for opening via `pointerdown` or by label:
	```ts
	// Tracks direct mouse presses handled in `pointerdown` so the Safari label
	// `click` workaround below does not re-focus the trigger after opening.
	let openedFromPointerDown = false
	```
	I think that's the most optimal and straightforward solution there is, but maybe you could point alternative better solution if you thought about it?
2) Fixed
3) Added test cases for initial issue with `mousedown` and for described focus issues. Also tested manually on a `Safari 26.5`
4) Fixed typo on the title (comments was rewritten containing more info, so mistake was removed)

zernonia (approved):

Issue #2677, opened 2026-06-05
[Bug]: DialogContentModal overrides disableOutsidePointerEvents


rafael-perini:
### Environment

```bash
Developement/Production OS: Windows 10 19043.1110
Node version: 16.0.0
Package manager: pnpm@8.6.0
Reka UI version: 2.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@3.3.3
Client OS: Windows 10 19043.1110
Browser: Chrome 90.0.4430.212
```

### Link to minimal reproduction

https://reka-ui.com/docs/components/dialog#content

### Steps to reproduce

Open any dialog with `<dialog-content :disableOutsidePointerEvents="false">` and you'll see that the body stills receive the `style="pointer-events: none;"`, even tought the docs says that the prop exists.

### Describe the bug

According to the docs bellow, the dialog content should be able to receive the `disableOutsidePointerEvents` prop.

<img width="753" height="475" alt="Image" src="https://github.com/user-attachments/assets/c9f25a27-bcaf-412f-8eac-fee249e85f2e" />

But what actually happens is that the `DialogContentModal`, used by the `DialogContent` overrides it.

My suggestion is to change the code bellow to use `withDefaults` (`const props = withDefaults( defineProps<DialogContentImplProps>(), { disableOutsidePointerEvents: true })`) and remove the `:disable-outside-pointer-events="true"` allowing the `DialogContent` to actually receive and pass forward the prop.

<img width="925" height="469" alt="Image" src="https://github.com/user-attachments/assets/626f760a-65d0-49f1-8e35-0ed0764d7e31" />

### Expected behavior

The `DialogContent` to receive  and actually pass the `disableOutsidePointerEvents` prop forward.

PR #2692, opened 2026-06-10
fix(AlertDialog): preserve default pointer events lock

Perdolique:
## Summary

This plugs the same boolean-default hole in `AlertDialogContent` that `DialogContent` already fixed in #2679 🧩✨ Without it, the wrapper accidentally forwards an absent `disableOutsidePointerEvents` prop as `false`, so alert dialogs with an overlay can open with the body locked while the content itself stays non-interactive 😬

- 🐛 Preserve `undefined` for absent `disableOutsidePointerEvents` on `AlertDialogContent`, letting `DialogContentModal` keep its default lock behavior
- ✅ Add regression coverage that confirms alert dialog content receives `pointer-events: auto` while the body remains locked
- 🧹 Clean up the alert dialog test mount between cases so teleported layers do not stack up and muddy pointer-event assertions

## Why

`DialogContent` now avoids Vue's boolean prop coercion by defaulting `disableOutsidePointerEvents` to `undefined`. `AlertDialogContent` extends the same props, but still used plain `defineProps`, so Vue coerced the missing prop to `false` before forwarding it downstream.

That means `DialogContentModal` never gets to apply its default `true`, and `DismissableLayer` does not mark the active alert dialog content as pointer-interactive. The body can still be locked by the overlay path, leaving the visible buttons looking fine but unable to receive real browser clicks 🧊

This is a small follow-up to #2677 / #2679.

## Summary by CodeRabbit

* **Tests**
  * Added cleanup hooks and pointer-event behavior validation to AlertDialog test suite.

* **Bug Fixes**
  * Improved AlertDialog pointer-event handling to properly respect component prop configurations while maintaining expected modal interaction behavior.

### DIFF
--- a/packages/core/src/AlertDialog/AlertDialog.test.ts
+++ b/packages/core/src/AlertDialog/AlertDialog.test.ts
@@ -1,7 +1,7 @@
 import type { VueWrapper } from '@vue/test-utils'
-import { findAllByText, findByText, fireEvent } from '@testing-library/vue'
+import { findAllByText, findByRole, findByText, fireEvent } from '@testing-library/vue'
 import { mount } from '@vue/test-utils'
-import { beforeEach, describe, expect, it } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
 import { axe } from 'vitest-axe'
 import { nextTick } from 'vue'
 import AlertDialog from './story/_AlertDialog.vue'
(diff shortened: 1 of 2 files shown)


Perdolique:
@zernonia, it is also reproducible here: https://reka-ui.com/docs/components/alert-dialog

Perdolique:
@zernonia, it's a kind of important fix as AlertDialog doesn't work on 2.9.10 (#2702).

PR #2740, opened 2026-06-23
fix(Presence): detect CSS transitions alongside animations to prevent premature unmount

### LINKED ISSUE #2403: [Bug]: Dialog shifts to the left during close animation when scrollbar appears
### Environment

```bash
Developement/Production OS: Linux/Manjaro
Node version: 22.14.0
Package manager: npm@10.9.2
Reka UI version: 2.6.1
Vue version: 3.5.13
Nuxt: None
CSS framework: tailwindcss@4.1.18
Client OS: Linux/Manjaro
Browser: Chromium 143.0.7499.169
```

https://stackblitz.com/edit/vitejs-vite-6qhyfkgw

### Steps to reproduce

add an animation to a dialog, open and close it while there is a scrollbar on the body element.

### Describe the bug

When I added fade animations to my dialogs, I noticed, that the dialogs would jump 8px to the left when closing them.

This seems to be due to some styles being added to the body and later removed:
`padding-right: 16px; margin-right: 0px; overflow: hidden; pointer-events: none;`

### Expected behavior

I would expect the dialog not to shift. Probably the removal of the body styles should be delayed until the dialog is done with the animation.

**edit:** This seems to be working correctly if the DialogOverlay has an animation as well. Is that expected behavior maybe?

_No response_

yan-ad:
## Description

Fixes #2403

### Problem
When a component using `Presence` (e.g. `DialogOverlay`) has a CSS `transition` instead of a CSS `animation`, `usePresence` would immediately dispatch `UNMOUNT` because it only checked `animationName`. This caused the body scroll lock to release before the visual transition completed, resulting in a leftward shift when the scrollbar reappeared.

### Root Cause
`usePresence` only detected CSS `animation` (via `animationName` computed style) and completely ignored CSS `transition`. When closing a dialog that uses CSS transitions (e.g., `transition: opacity 0.3s ease`), the component would unmount instantly.

### Fix
- Added `hasTransition()` helper to check if an element has a non-none/non-zero CSS transition
- Added `getTransitionDuration()` to calculate the longest transition duration
- When leaving and no animation is detected but a transition exists, dispatch `ANIMATION_OUT` instead of `UNMOUNT`, keeping `isPresent` true
- Listen for `transitionend` and `transitioncancel` events to dispatch `ANIMATION_END`
- Fallback timeout based on transition duration as a safety net

### Testing
- All existing Presence tests pass (9 tests)
- Typecheck passes cleanly

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved component unmount behavior to account for CSS exit transitions, ensuring animations complete before removal.
  * Enhanced handling of both transition completion and cancellation events.
  * Coordinated unmount timing using the longest applicable transition (including delay), with immediate fallback when no exit transition applies.

### DIFF
--- a/packages/core/src/Presence/usePresence.ts
+++ b/packages/core/src/Presence/usePresence.ts
@@ -14,6 +14,7 @@ export function usePresence(
   const initialState = present.value ? 'mounted' : 'unmounted'
   let timeoutId: number | undefined
   let transitionTimeout: number | undefined
+  let transitionPendingCount: number | undefined
   const ownerWindow = node.value?.ownerDocument.defaultView ?? defaultWindow
 
   const { state, dispatch } = useStateMachine(initialState, {
@@ -68,12 +69,14 @@ export function usePresence(
 
             const duration = getTransitionDuration(node.value)
             if (duration > 0) {
+              transitionPendingCount = getTransitionPropertyCount(node.value)
               transitionTimeout = ownerWindow?.setTimeout(() => {
                 if (state.value === 'unmountSuspended') {
                   dispatchCustomEvent('after-leave')
                   dispatch('ANIMATION_END')
                 }
                 transitionTimeout = undefined
+                transitionPendingCount = undefined
               }, duration + 50)
             }
             else {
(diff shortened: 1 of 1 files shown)


zernonia (changes requested):
Thanks for digging into #2403! A few concerns before this can land:

1. Any Presence element that declares a transition for unrelated reasons (e.g. Tailwind `transition-colors`) now stays mounted for `duration + 50ms` after close, where it unmounted immediately before.
2. close → open → close leaves the first `transitionTimeout` running, and it unmounts the second close early (the timer needs clearing on MOUNT / before re-arming).
3. Closing mid-enter fires `transitioncancel` for the reversed transition after we're already `unmountSuspended`, which ends the leave at once (the animation path guards this case).
4. Property counting doesn't work for `all`, long property lists, or shorthands.

If we do want transition support, `el.getAnimations()` and awaiting `.finished` handles all of this without parsing. Separately, #2403's repro seems to be the overlay releasing the scroll lock while the content is still animating out, which might be better fixed in Dialog itself. Could you also add tests (jsdom needs longhand `transition-property`/`transition-duration`)?

PR #2741, opened 2026-06-23
fix(Select): close open select when another opens

### LINKED ISSUE #2583: [Bug]: `<SelectContent>` doesn't close when opening an other select
### Environment

```bash
Developement/Production OS: Debian 12
Node version: 22.22.0
Package manager: yarn@1.22.22
Reka UI version: 2.9.4
Vue version: 3.5.29
Client OS: Debian 12
Browser: Chrome 146.0.7680.177
```

https://stackblitz.com/edit/fbvyauzr-yoqpxgmt?file=src%2FApp.vue

### Steps to reproduce

1. Open one of the select boxes
2. Open the other select box

The one that was opened first doesn't close

### Describe the bug

When there are two select boxes and the triggers (or any of their parents) have `pointer-events: auto` on them it overrides the `pointer-events: none` that is put on the body by Reka. When one of them is opened, clicking on the other trigger doesn't close the opened select's content.
This seems to be handled for dropdowns: https://stackblitz.com/edit/mp87vmpa?file=src%2FApp.vue

### Expected behavior

When opening a select, the opened one should close.

Image

yan-ad:
## Description

Resolves #2583.

When multiple Select components are on the same page, opening one Select should close any other open Select. This matches native `<select>` behavior where only one dropdown can be open at a time.

### Changes
- Added a module-level `openSelects` Set to track all open SelectRoot instances
- When a Select opens, it calls `onOpenChange(false)` on all other registered Selects
- Added regression test for two Selects with `pointer-events: auto` on triggers

### Testing
All 30 Select tests pass, including the new "should close the previously opened select when another opens" test.

## Release Notes

* **Bug Fixes**
  * Improved Select behavior so only one Select instance can be open at a time. Opening a Select now automatically closes any other open instances and keeps ARIA state in sync.

* **Tests**
  * Added coverage for multiple Select instances in the same view, including checks that only one listbox is present at a time, ARIA `aria-expanded` updates correctly, and an accessibility (axe) verification passes.

### DIFF
--- a/packages/core/src/Select/Select.test.ts
+++ b/packages/core/src/Select/Select.test.ts
@@ -38,14 +38,14 @@ const TwoSelects = defineComponent({
     SelectViewport,
   },
   setup() {
-    const first = ref('')
-    const second = ref('')
+    const first = ref()
+    const second = ref()
 
     return { first, second }
   },
   template: `
     <SelectRoot v-model="first">
-      <SelectTrigger style="pointer-events: auto">
+      <SelectTrigger aria-label="First" style="pointer-events: auto">
         <SelectValue placeholder="First" />
       </SelectTrigger>
       <SelectPortal>
@@ -60,7 +60,7 @@ const TwoSelects = defineComponent({
     </SelectRoot>
 
     <SelectRoot v-model="second">
-      <SelectTrigger style="pointer-events: auto">
+      <SelectTrigger aria-label="Second" style="pointer-events: auto">
         <SelectValue placeholder="Second" />
       </SelectTrigger>
       <SelectPortal>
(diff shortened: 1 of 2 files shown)


zernonia (changes requested):
Thanks for picking this up! I can reproduce #2583 in a real browser, but tracing it, the cause seems to be in DismissableLayer rather than Select: the trigger opens the new Select on `pointerdown`, Vue flushes in the microtask checkpoint before the first layer's document `pointerdown` listener runs, and the new layer is already registered as the top pointer-blocking layer. So the old layer treats the click as blocked (`isPointerEventsEnabled` is false) and never dismisses. The same thing happens with an open DropdownMenu and a Select trigger, which a Select-only registry can't cover.

The registry also emits `update:open(false)` to every other mounted Select, including controlled ones that are already closed, on each open. Would you be open to fixing this in DismissableLayer instead (e.g. checking whether a layer is blocked against the stack as it was *before* the event's handlers ran)?

Also, the new test fails on v2 only because it clicks before the layer's listener is registered (it's added on a `setTimeout`); if you wait a macrotask first, v2 passes it. eslint also flags `SelectRoot.vue` (indent) and the import order in `Select.test.ts`.

PR #2747, opened 2026-06-23
fix(Menu): allow Tab to move focus out of non-modal DropdownMenu

### LINKED ISSUE #2296: [Bug]: Unable to use Tab to move focus outside of open, non-modal DropdownMenu
### Environment

Not applicable. See description.

```bash
Development/Production OS: macOS 15.7
Node version: N/A
Package manager: N/A
Reka UI version: 2.6.0
Vue version: 3.5.24
Nuxt version: N/A
Nuxt mode: N/A
Nuxt target: N/A
CSS framework: N/A
Client OS: Any
Browser: Any
```

https://stackblitz.com/edit/afauw766?file=src%2FApp.vue

### Steps to reproduce

Given a page that has a DropdownMenu (with `:modal="false"`), a button/trigger, and at least one other focusable element:

1. Tab to focus the menu trigger
2. Press `Enter` to activate the menu
3. Press `Tab` to attempt to move focus out of the menu and on to the next item on the page, per the [keyboard interaction pattern for menubar](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/#keyboardinteraction). Note that the `Tab`/focus behavior is completely disabled.

### Describe the bug

When the user's focus is within an open, non-modal DropdownMenu and they press the `Tab` key, nothing happens. 

### Expected behavior

I expect focus to move from the DropdownMenu to the next focus-able element on the page, per the [keyboard interaction pattern for menubar](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/#keyboardinteraction).

### Context

[This section of code](https://github.com/unovue/reka-ui/blob/1e0670da1d9cf9652cb4cf89aa2015265dbd60d0/packages/core/src/Menu/MenuContentImpl.vue#L197-L199) seems to be what kills the `Tab` keypress event: 

```ts
if (isKeyDownInside) {
    // menus should not be navigated using tab key so we prevent it
    if (event.key === 'Tab')
      event.preventDefault()
```

It seems the intent was to prevent the `Tab` key from navigating focus between menu items. That _intent_ is correct, as navigation should only happen with Arrow keys. However, that should not be implemented by completely disabling the `Tab` key.

### Non-Solutions

These "solutions" were considered and are not acceptable:

1. Implementing another `Tab` keypress handler on my component that uses DropdownMenu, calling `preventDefault`, and either:
   a. Manually finding the next "focusable" DOM element on the page and calling .focus() on it. (Or a previous DOM element in the case of Shift+Tab)
   b. Re-emitting a `Tab` keypress event outside of DropdownMenu in hopes that it will trigger the browser's default `Tab`/focus handling behavior

### Possible Fixes

In all cases, remove the `preventDefault` call on the `Tab` keypress event, then one or more of:

1. Correctly implement roving tabindex by setting `tabindex="-1"` on all items except for the currently-selected one
2. Provide a configuration option that allows us to turn off the section of code that kills the `Tab` key to allow library consumers to (optionally) handle it ourselves.

yan-ad:
## Description

Fixes #2296 — Unable to use Tab to move focus outside of open, non-modal DropdownMenu.

### Root Cause

`MenuContentImpl.vue` unconditionally called `event.preventDefault()` on Tab keydown when inside the menu. This was intended to prevent Tab from navigating between menu items (menus use arrow keys), but it had the side effect of completely disabling Tab — focus could never leave the menu.

### Fix

Gate the Tab `preventDefault` on `trapFocus.value`. When the menu is modal (`trapFocus=true`), Tab is still prevented (FocusScope handles focus trapping). When non-modal (`trapFocus=false`, which is the default for DropdownMenu), Tab is allowed to naturally move focus to the next focusable element on the page, per the WAI-ARIA menubar keyboard interaction pattern.

### Testing

All 21 tests pass: Menu (6), DropdownMenu (6), ContextMenu (3), Menubar (6)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed keyboard `Tab` behavior in non-modal menus so users can move focus forward instead of being trapped.
  * Preserved `Tab` focus handling for modal menus, including nested submenus, preventing unintended focus changes.
* **Tests**
  * Added tab-navigation coverage validating `Tab` behavior for modal vs non-modal menus and nested submenu keydown prevention.

### DIFF
--- a/packages/core/src/DropdownMenu/DropdownMenu.test.ts
+++ b/packages/core/src/DropdownMenu/DropdownMenu.test.ts
@@ -1,9 +1,61 @@
 import type { VueWrapper } from '@vue/test-utils'
-import { findByRole } from '@testing-library/vue'
+import { fireEvent, findAllByRole, findByRole, render } from '@testing-library/vue'
 import { mount } from '@vue/test-utils'
 import { beforeEach, describe, expect, it } from 'vitest'
 import { axe } from 'vitest-axe'
+import { defineComponent } from 'vue'
+import {
+  DropdownMenuContent,
+  DropdownMenuItem,
+  DropdownMenuPortal,
+  DropdownMenuRoot,
+  DropdownMenuSub,
+  DropdownMenuSubContent,
+  DropdownMenuSubTrigger,
+  DropdownMenuTrigger,
+} from '.'
 import DropdownMenu from './story/_DropdownMenu.vue'
+
+const DropdownMenuTabTest = defineComponent({
+  components: {
+    DropdownMenuContent,
+    DropdownMenuItem,
+    DropdownMenuPortal,
+    DropdownMenuRoot,
+    DropdownMenuSub,
+    DropdownMenuSubContent,
+    DropdownMenuSubTrigger,
+    DropdownMenuTrigger,
+  },
+  props: {
+    modal: {
+      type: Boolean,
+      default: true,
+    },
+  },
+  template: `
+    <div>
+      <button>Before</button>
+      <DropdownMenuRoot :open="true" :modal="modal">
+        <DropdownMenuTrigger>Open</DropdownMenuTrigger>
+        <DropdownMenuPortal disabled>
+          <DropdownMenuContent @close-auto-focus.prevent>
+            <DropdownMenuItem>Item</DropdownMenuItem>
+            <DropdownMenuSub :open="true">
+              <DropdownMenuSubTrigger>Sub Trigger</DropdownMenuSubTrigger>
+              <DropdownMenuPortal disabled>
+                <DropdownMenuSubContent>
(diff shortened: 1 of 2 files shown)


zernonia:
Thanks for the fix — the gating on `trapFocus` looks correct for the non-modal case 👍

Could you add a regression test before we merge? There's currently no Tab-keydown coverage in the Menu/DropdownMenu suites, so "all tests pass" doesn't actually exercise this behavior. Ideally:

- **non-modal** menu: Tab is *not* `preventDefault`'d and focus moves to the next focusable element
- **modal** menu: Tab is still prevented

One more thing to confirm: submenus omit `trapFocus`, so `trapFocus.value` is `undefined` (falsy) for any `MenuSubContent`. That means Tab will now also escape an open submenu inside a *modal* menu. Is that intended? A submenu test case would be great to lock down whichever behavior we want.

Issue #2335, opened 2025-12-11
[Bug]: Combobox inside dialog cannot be focued


laserhybiz:
### Environment

```bash
Developement/Production OS: Apple M3 Pro / macOS Sequoia 15.6.1
Node version: 22.21.1
Package manager: npm@10.9.4
Reka UI version: 2.6.1
Vue version: 3.5.25
CSS framework: tailwindcss@4.1.18
Client OS: Apple M3 Pro / macOS Sequoia 15.6.1
Browser: Chrome 142.0.7444.176
```

### Link to minimal reproduction

https://stackblitz.com/edit/p1kyubba-rkoujers

### Steps to reproduce

Create a combobox with an input (shadcn) inside a dialog

### Describe the bug

the input of the combobox cannot be focused and cannot enter any value

rastuhacode:
@laserhybiz
It has pretty nice workaround with [FocusScope](https://reka-ui.com/docs/utilities/focus-scope)

In your example i've modified the `ComboboxList` like this:

``` vue
<script setup lang="ts">
import type { ComboboxContentEmits, ComboboxContentProps } from 'reka-ui';
import type { HTMLAttributes } from 'vue';
import { reactiveOmit } from '@vueuse/core';
import {
  ComboboxContent,
  ComboboxPortal,
  useForwardPropsEmits,
  // Import utiltity
  FocusScope,
} from 'reka-ui';

// Your code as follows
</script>

<template>
  <ComboboxPortal>
    <ComboboxContent
      data-slot="combobox-list"
      ...
      >
      <FocusScope>
        <slot />
      </FocusScope>
    </ComboboxContent>
  </ComboboxPortal>
</template>
```

With this change it seems like input focus works as expected

PR #2752, opened 2026-06-23
fix(FocusScope): default present to true so scopes without it still pause ancestor traps

### LINKED ISSUE #2749: [Bug]: Regression Bug #2335, Combobox inside dialog cannot be focued
### Environment

```bash
Developement/Production OS: MacOS 13.7.8 (22H730)
Node version: v24.17.0
Package manager: yarn@4.17.0
Reka UI version: 2.10.0
Vue version: 3.5.38
CSS framework: tailwindcss@4.3.1
Client OS: MacOS 13.7.8 (22H730)
Browser: Edge 149.0.4022.80
```

https://stackblitz.com/edit/p1kyubba-swv2kttq?file=package.json

### Steps to reproduce

Reproduction is the same as in #2335 but updated to reka-ui 2.10.0.

1. Open stackblitz repro, and wait until it load (takes a long time for some reason).
2. Click "Open Dialog" button
3. Click on "Select framework" ComboBox
4. Use arrows to select framework (doesn't work)
5. Try typing "Svelte" (doesn't work)

### Describe the bug

Same as #2335, Input inside ComboBox Content, inside Dialog Content cannot be focused not interested

### Expected behavior

To work same as in reka-ui V2.9.10

_No response_

### LINKED ISSUE #2756: [Bug]: Regression 2.10 : dropdown menu items not accessible with keyboard
### Environment

```bash
Developement/Production OS: macOS 26.5.1 (25F80)
Node version: 24.18.0
Package manager: pnpm@11.6.0
Reka UI version: 2.10.0
Vue version: 3.0.0
```

https://github.com/Ericlm/reka-dropdown-focus

### Steps to reproduce

1. Create a DropdownMenu with a trigger and some DropdownMenuItems
2. Click on the trigger
3. Try to access a DropdownMenuItem with keyboard (arrow keys)

### Describe the bug

When the DropdownMenu is opened, the items are not accessible by using the arrow up/down. This is a regression between 2.9 and 2.10.

### Expected behavior

We should be able to interact with the DropdownMenuItems using arrow keys.

_No response_

benjamincanac:
### 🔗 Linked issue

Resolves #2749 (regression of #2335), resolves #2756.

### ❓ Type of change

### 📚 Description

The 2.10.0 `FocusScope` refactor (#2662) added a `present` prop and gated **two** behaviors on `props.present !== false`:

1. **Stack membership** — registering in the global focus-scope stack, which **pauses an active ancestor trap** (e.g. a `Dialog`'s trapped `FocusScope`).
2. **Mount auto-focus** — dispatching `mountAutoFocus` and moving focus into the scope's content on mount.

It intended an absent prop to mean "present". But `present` is a **Boolean-typed prop**, so Vue's [boolean casting](https://vuejs.org/guide/components/props.html#boolean-casting) coerces an *absent* prop to `false`, not `undefined`. So `props.present !== false` evaluated to `false`, and **both** behaviors were silently disabled for every `FocusScope` consumer that doesn't pass `present` explicitly — `Combobox`, `Select`, `Popover`, `Menu` (DropdownMenu/Menubar/ContextMenu) and `Drawer` content. Only `Dialog*` (which forwards `present`) was unaffected.

That single gate produced two distinct regressions:

- **#2749 (re-introducing #2335)** — a `Combobox`/`Select` portaled out of an enclosing `Dialog` no longer registered in the stack, so it didn't pause the Dialog's trap and the Dialog kept yanking focus back. The input/items could not be focused or typed.
- **#2756** — a **standalone** `DropdownMenu` (no Dialog involved) no longer dispatched its mount auto-focus, so focus never moved into the menu content on open and arrow-key navigation between items stopped working.

The fix defaults `present` to `true` in `withDefaults`, matching the prop's own documented `@defaultValue true`. Because `withDefaults` supplies a default, Vue's boolean cast-to-`false` for an absent prop no longer applies, so `present` resolves to `true` as intended. Consumers that pass `present` explicitly (`Dialog` with `unmountOnHide`) are unaffected.

Verified in a real browser (`Dialog` → `ComboboxRoot` → `ComboboxPortal` → `ComboboxContent` with the input inside the content):

| Ref | Input focusable / typable |
| --- | --- |
| v2.9.10 | ✅ works |
| v2.10.0 | ❌ focus trapped back to Dialog |
| this PR | ✅ works |

### 🧪 Tests

Two regression tests in `FocusScope.test.ts`, each failing without the fix and passing with it:
- Combobox input inside Dialog (the reported #2749 case).
- Select content (portaled) inside Dialog (a second consumer, to document the fix is not Combobox-specific).

Both cover the stack-membership path (pausing the Dialog's trap); the Combobox test also asserts the mount auto-focus lands on the input, which is the same gate behind #2756.

Full `packages/core` suite (1946 tests) passes.

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a focus management regression so focus remains correctly inside Dialogs that contain portal-rendered Combobox or Select elements.

* **Tests**
  * Added regression coverage for Dialog + Combobox focus behavior, including preventing the Dialog focus trap from re-stealing focus immediately.
  * Added regression coverage for portal-rendered Select content within Dialogs to ensure focus lands inside the Select.

* **Documentation**
  * Clarified the `present` prop default for `FocusScope` and its relevance to focus-scope behavior.

### DIFF
--- a/docs/content/meta/FocusScope.md
+++ b/docs/content/meta/FocusScope.md
@@ -26,7 +26,8 @@
     'name': 'present',
     'description': '<p>Whether the scope is currently visible. Lets a consumer keep the scope\nmounted but hidden (e.g. <code>display: none</code>) and still get correct auto-focus:\nthe mount auto-focus is skipped while not present, and re-runs when it\nbecomes present again. Defaults to <code>true</code> so consumers that mount the scope\nonly while visible are unaffected.</p>\n',
     'type': 'boolean',
-    'required': false
+    'required': false,
+    'default': 'true'
   },
   {
     'name': 'trapped',
@@ -60,7 +61,7 @@
 | `as` | The element or component this component should render as. Can be overwritten by asChild. | `AsTag \| Component` | No | `"div"` |
 | `asChild` | Change the default rendered element for the one passed as a child, merging their props and behavior. Read our Composition guide for more details. | `boolean` | No | - |
 | `loop` | When true, tabbing from last item will focus first tabbable and shift+tab from first item will focus last tababble. | `boolean` | No | `false` |
-| `present` | Whether the scope is currently visible. Lets a consumer keep the scope mounted but hidden (e.g. display: none) and still get correct auto-focus: the mount auto-focus is skipped while not present, and re-runs when it becomes present again. Defaults to true so consumers that mount the scope only while visible are unaffected. | `boolean` | No | - |
+| `present` | Whether the scope is currently visible. Lets a consumer keep the scope mounted but hidden (e.g. display: none) and still get correct auto-focus: the mount auto-focus is skipped while not present, and re-runs when it becomes present again. Defaults to true so consumers that mount the scope only while visible are unaffected. | `boolean` | No | `true` |
 | `trapped` | When true, focus cannot escape the focus scope via keyboard, pointer, or a programmatic focus. | `boolean` | No | `false` |
 
 **Events**
--- a/packages/core/src/FocusScope/FocusScope.test.ts
+++ b/packages/core/src/FocusScope/FocusScope.test.ts
@@ -1,12 +1,12 @@
 import type { RenderResult } from '@testing-library/vue'
 import userEvent from '@testing-library/user-event'
 import { render } from '@testing-library/vue'
-import { beforeEach, describe, expect, it, vi } from 'vitest'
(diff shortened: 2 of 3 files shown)


zernonia:
LGTM! Thanks @benjamincanac

PR #2753, opened 2026-06-23
fix(Listbox): keep activedescendant highlight on focusout when filter-driven

benjamincanac:
### 🔗 Linked issue

2.10.0 regression: filter-driven listboxes (Combobox / Select, e.g. Nuxt UI's `USelectMenu`) no longer highlight the first item on open. Related: #2752.

### ❓ Type of change

### 📚 Description

When a `Listbox` is driven by a filter input (`ListboxFilter`, used by `ComboboxInput`), it sets `rootContext.focusable = false`: DOM focus deliberately stays on the input and items are highlighted via `aria-activedescendant`, never DOM-focused. In a `Combobox`/`Select` the content (input + items) is also portaled **out** of the `ListboxRoot` element, so `currentElement.contains(target)` in the root's `@focusout` handler is *always* `false`.

`packages/core/src/Listbox` is byte-identical between 2.9.10 and 2.10.0, but a 2.10.0 timing change moved the relevant `focusout` to land *after* the on-open highlight is set. The handler then sees focus "leave the listbox" (it can't contain the portaled input) and calls `onLeave`, which nulls `highlightedElement` and wipes the just-set highlight.

The fix gates the highlight-clearing branch on `focusable`: when `focusable === false`, a `focusout` from `currentElement` tells us nothing about whether the user left the widget (focus lives on the portaled input by design), so the highlight must not be cleared. Standalone (DOM-focusable) listboxes are unchanged.

```diff
- if (highlightedElement && currentElement && !currentElement.contains(target))
+ if (focusable && highlightedElement && currentElement && !currentElement.contains(target))
    onLeave(event)
```

### 🧪 Tests

Added a deterministic regression test in `Listbox.test.ts` ("filter-driven Listbox"): with a `ListboxFilter` present (`focusable === false`), a root `focusout` whose target sits outside the root must not clear the `aria-activedescendant` highlight. It exercises the exact code path without relying on focus-trap timing (which doesn't reproduce reliably in jsdom) — it fails without the gate and passes with it.

### 📝 Checklist

### DIFF
--- a/packages/core/src/Listbox/Listbox.test.ts
+++ b/packages/core/src/Listbox/Listbox.test.ts
@@ -5,7 +5,7 @@ import { axe } from 'vitest-axe'
 import { defineComponent, h, nextTick, ref } from 'vue'
 import { useKbd } from '@/shared'
 import { handleSubmit } from '@/test'
-import { ListboxContent, ListboxItem, ListboxRoot, ListboxVirtualizer } from '.'
+import { ListboxContent, ListboxFilter, ListboxItem, ListboxRoot, ListboxVirtualizer } from '.'
 import Listbox from './story/_Listbox.vue'
 
 describe('given default Listbox', () => {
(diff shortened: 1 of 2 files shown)


benjamincanac:
Closing in favor of https://github.com/unovue/reka-ui/pull/2752.

Issue #2760, opened 2026-06-24
[Feature]: NumberField — allowInvalid prop to keep typed input as-is (skip clamping & step snapping)


xerox0213:
### Describe the feature

By default `NumberField` normalizes a typed value on commit: it clamps it to `[min, max]` and snaps it to the step grid. There's no way to let the user enter a value the field would otherwise consider invalid.

I'd like an `allowInvalid` prop (default `false`, non-breaking):

- When `true`, a directly typed value is kept exactly as entered, even if it is invalid —  i.e. out of the `min`/`max` range **or** off the step grid (only its formatting precision  is normalized). It is no longer clamped or snapped.
- Step-based interactions (keyboard arrows, +/- buttons, wheel) **still clamp and snap**;  only free-form typing is allowed to produce an invalid value.

**Use case**

Forms where the user should be able to enter a value that violates the field's constraints (out of range, or a finer-grained value than the step) so the app / the consumer's own validation can surface it, instead of the field silently coercing it.

**Notes**

- "Invalid" maps to HTML constraint-validation states: `rangeUnderflow`/`rangeOverflow` (range) and `stepMismatch` (grid) — hence the single name covering both axes.
- Wiring native browser validation was intentionally left out: the form input is visually hidden, so a native validation bubble would be mis-anchored. The prop only controls whether typed input is coerced; validation UI is left to the consumer.

### Additional information

- [ ] I intend to submit a PR for this feature.
- [x] I have already implemented and/or tested this feature.

messenjer:
The same thing as `allowOutOfRange` prop of Base UI https://base-ui.com/react/components/number-field#NumberFieldRoot-allowOutOfRange

PR #2761, opened 2026-06-24
feat(NumberField): add allowInvalid prop

xerox0213:
### 🔗 Linked issue

#2760

### ❓ Type of change

- [x] 📖 Documentation (updates to the documentation, readme or JSdoc annotations)

### 📚 Description

Adds an `allowInvalid` prop to `NumberField`.

When `true`, a directly typed value is kept exactly as entered even if it is invalid — out of the `min`/`max` range or off the step grid (only its formatting precision is normalized). Step-based interactions (keyboard arrows, +/- buttons, wheel) still clamp and snap to the range/grid. Defaults to `false`, so existing behavior is unchanged.

"Invalid" maps to the HTML constraint-validation states (`rangeUnderflow`/`rangeOverflow` and `stepMismatch`), which is why a single prop covers both the range and the step-grid axes. I went with `allowInvalid` over `allowOutOfRange` for that reason — open to a different name if you'd prefer.

> ⚠️ **Stacked PR — contains the fix PR's commits.**
> Built on top of #2759 (the off-grid stepping fix), because `allowInvalid`
> reuses the same commit path. Targeting `v2` directly, so the diff currently also shows
> those fix commits; it will reduce to just the `allowInvalid` commit once #2759
> is merged. **Please review/merge #2759 first.**

### 📝 Checklist

## Summary by CodeRabbit

* **New Features**
  * Added an `allowInvalid` option to keep typed number field values as entered, even when out of range or off the step grid (while normal step interactions still snap/clamp).

* **Bug Fixes**
  * Refined increment/decrement behavior to snap to the correct step grid and properly clamp values near minimum/maximum bounds, including better handling of empty/NaN fallback behavior.

* **Documentation**
  * Updated the number field docs to describe the new typing behavior with `allowInvalid`.

* **Tests**
  * Expanded coverage for step snapping, boundary disabling, and `allowInvalid` interactions.

### DIFF
--- a/docs/content/meta/NumberFieldRoot.md
+++ b/docs/content/meta/NumberFieldRoot.md
@@ -3,6 +3,12 @@
 <llm-exclude>
 <PropsTable :data="[
   {
+    'name': 'allowInvalid',
+    'description': '<p>When <code>true</code>, a typed value is kept as-is even if invalid (out of <code>min</code>/<code>max</code> range or off the step grid); step interactions still clamp and snap.</p>\n',
+    'type': 'boolean',
+    'required': false
+  },
+  {
     'name': 'as',
     'description': '<p>The element or component this component should render as. Can be overwritten by <code>asChild</code>.</p>\n',
     'type': 'AsTag | Component',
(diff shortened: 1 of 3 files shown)


zernonia (changes requested):
Thanks @xerox0213! #2759 has landed (squashed), so could you rebase this onto v2 so it's just the `allowInvalid` commit? It cherry-picks cleanly.

One functional gap: `validate()` still passes `min`/`max` to `isValidPartialNumber`, which rejects a leading `-` whenever `min >= 0`, so with `allowInvalid` you still can't type a value below a non-negative min. Could you skip the bounds there when `allowInvalid` is true, and add a test that types through `beforeinput`? The current tests set `input.value` directly, so they don't catch this.

It would also help to mention in the prop docs that the field doesn't mark the value as invalid (no `aria-invalid`/`data-invalid`), so consumers need to validate it themselves. Otherwise this looks good!

PR #2763, opened 2026-06-25
fix(shared): create context with `Symbol.for` to fix duplicates

TheAlexLichter:
### 🔗 Linked issue

#229

### ❓ Type of change

### 📚 Description

`createContext` used a module-local `Symbol(description)` as the injection key. When reka-ui is installed more than once (e.g. a version mismatch between libraries that depend on it, duplicated by the package manager), each copy creates a *different* Symbol, so a child component from one copy cannot `inject` a context provided by a parent from another copy and triggers a `Injection \Symbol(...)\ not found. Component must be used within ...` error.

This PR uses the global symbol registry (`Symbol.for('reka-ui:<name>')`) instead, which resolves to the same identity across duplicate copies, so provide/inject keeps working regardless of how the dependency graph is deduplicated.

Every context already has a unique name, so there are no cross-context collisions; the array provider form (no explicit name) now derives a stable name from the provider component names.

Same approach other UI libraries like Vuetify and even React use.

### 👀 Additional notes

On a side note, this *likely* allows Nuxt UI to drop `transpile` for reka-ui, causing the first dev server request to be sped up by 0.8-1.2 **seconds**

### 📝 Checklist

### DIFF
--- a/packages/core/src/shared/createContext.test.ts
+++ b/packages/core/src/shared/createContext.test.ts
@@ -1,0 +1,54 @@
+import { describe, expect, it } from 'vitest'
+import { createSSRApp, defineComponent, h } from 'vue'
+import { renderToString } from 'vue/server-renderer'
+import { createContext } from './createContext'
+
+describe('createContext', () => {
+  // Two independent `createContext()` calls with the same name simulate the
+  // library being installed twice (e.g. a version mismatch duplicated by npm).
+  // With a module-local `Symbol()` the two keys differ and injection fails
+  // (`Injection Symbol(...) not found`, #229); with the global registry
+  // (`Symbol.for`) they share identity and provide/inject keeps working.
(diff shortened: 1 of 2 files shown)


zernonia (changes requested):
Thanks Alex, nice fix for the duplicate-instance case, and internally all context names are unique.

One concern: `createContext` is public API and used heavily downstream (Nuxt UI, shadcn-vue, …), so consumer contexts now land in the global `reka-ui:` namespace too. A consumer `createContext('DialogRoot')` inside a reka `DialogRoot` now silently injects reka's context (verified locally), and two libraries that both use e.g. `'Sidebar'` would now collide.

Could we apply `Symbol.for` only to reka's internal contexts (an internal variant or an opt-in flag) and keep the public helper module-local? It might also be worth putting the major version in the key (`reka-ui@2:`) so incompatible majors never share context shapes.

Issue #1865, opened 2025-04-30
[Bug]: Select - Close animation is not working


blro0319:
### Environment

```bash
Developement/Production OS: Windows 10 19045.5737
Node version: 22.13.1
Package manager: pnpm@10.9.0
Reka UI version: 2.2.0
Vue version: 3.5.13
Nuxt version: 3.17.0
Nuxt mode: universal
Nuxt target: server
CSS framework: tailwindcss@4.1.4
Client OS: Windows 10 19045.5737
Browser: Chrome 135.0.7049.115
```

### Link to minimal reproduction

https://stackblitz.com/edit/vda7rpmu?file=src%2FApp.vue

### Steps to reproduce

1. Open select menu. Opening animation is work.
2. Close select menu. Closing animation is not work.

### Describe the bug

The `data-state` attribute doesn't change to `closed` when it closes. This can be seen in the [debug reproduction](https://stackblitz.com/edit/vda7rpmu-kfwuq8dh?file=src%2FApp.vue).

It seems that the same issue exists in the [select](https://www.shadcn-vue.com/docs/components/select.html) component of shadcn/vue.

PR #2768, opened 2026-06-26
fix(Select): prevent placeholder flash when closing after selecting a value

### LINKED ISSUE #2767: [Bug]: Select placeholder briefly flashes when selecting a value (single-select)
### Environment

```bash
Node version: 16.0.0
Package manager: pnpm@8.6.0
Reka UI version: 2.0.0
Vue version: 3.0.0
Nuxt version: 3.0.0
Nuxt mode: universal
Nuxt target: server
Browser: Chrome 90.0.4430.212
```

https://reka-ui.com/docs/components/select#basic-example

### Steps to reproduce

1. Render a single-select `Select` with a `SelectValue` that has a `placeholder` (e.g. "Please select a fruit").
2. Click the trigger to open the content.
3. Click an item (e.g. "Banana") to select it — this also closes the content.
4. Watch the trigger label right as the content closes.

**Result:** the selected label ("Banana") is replaced by the placeholder for a single frame, then the label reappears.

### Describe the bug

When the content closes, `SelectContent` moves the `SelectItem`s from the popper (`Presence`) to a fallback `DocumentFragment` so they stay mounted and keep `optionsSet` populated. But `Presence` unmounts the popper items a few microtasks *before* the fragment remounts them (the fragment branch only activates on the `renderPresence` setTimeout / macrotask). During that handoff window `optionsSet` is transiently empty, so `SelectValue` cannot resolve the selected label and falls back to the placeholder — producing a one-frame flash.

### Expected behavior

The trigger should keep showing the selected value continuously while the content closes. The placeholder should only appear when no value is actually selected.

https://github.com/user-attachments/assets/6533ad64-e6d2-4efc-8780-ba5a53541ac0

xerox0213:
### 🔗 Linked issue

Resolves #2767

### ❓ Type of change

### 📚 Description

When a single-select `Select` closes after picking a value, the trigger label
briefly flashed the placeholder before showing the selected value again.

**Root cause:** on close, `SelectContent` moves the `SelectItem`s from the popper (`Presence`) to the fallback `DocumentFragment` so they stay mounted and keep `optionsSet` populated. But `Presence` unmounts the popper items a few microtasks *before* the fragment remounts them (the fragment branch only activates on the `renderPresence` `setTimeout` / macrotask). During that handoff window `optionsSet` is transiently empty, so `SelectValue` could not resolve the selected label and fell back to the placeholder — a one-frame flash.

**Fix:** `SelectValue` now keeps the last resolved label and only falls back to the placeholder when the model value is actually empty (`isEmptyModelValue`). This is a localized fix that doesn't touch `SelectContent`'s closing/animation timing (previously tuned in #1865 / #2638).

Added a self-contained regression test that stays inside the close handoff window (drains microtasks without firing the `setTimeout` macrotask) — it fails without the fix and passes with it.

### Fix in video

https://github.com/user-attachments/assets/699e970d-1b86-48bd-91d9-8bfb6ea72d1b

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented the selected value from briefly flashing back to the placeholder when closing a single-select dropdown.
  * Kept the chosen label visible during the close/unmount transition for a smoother experience.

* **Tests**
  * Added coverage for the close animation timing to verify the selected option remains displayed while the dropdown is shutting down.

### DIFF
--- a/packages/core/src/Select/Select.test.ts
+++ b/packages/core/src/Select/Select.test.ts
@@ -412,3 +412,29 @@ describe('given Select in a form', async () => {
     })
   })
 })
+
+describe('placeholder flash on close (#2638)', () => {
+  it('should keep showing the selected value while the content is closing', async () => {
+    document.body.innerHTML = ''
+    const wrapper = mount(Select, { attachTo: document.body })
+    const valueBox = wrapper.find('[aria-label="Customise options"]')
+
+    // Open then select an item, which closes the content in a single-select.
+    await wrapper.find('button').trigger('pointerdown', { button: 0, ctrlKey: false })
+    await nextTick()
+    const selection = wrapper.findAll('[role=option]')[1];
+    (selection.element as HTMLElement).focus()
+    await selection.trigger('pointerup')
+    // Needs 2 pointerup because SelectContentImpl prevents accidental pointerup's
+    await fireEvent.pointerUp(selection.element)
+
+    // While closing, `Presence` unmounts the popper items (clearing `optionsSet`)
+    // a few microtasks before the fallback fragment remounts them on the
+    // `renderPresence` setTimeout (macrotask). Draining only microtasks keeps us
+    // inside that handoff window, where the placeholder used to flash.
+    for (let i = 0; i < 5; i++) {
+      await nextTick()
+      expect(valueBox.text()).toBe('Banana')
+    }
+  })
+})
--- a/packages/core/src/Select/SelectValue.vue
+++ b/packages/core/src/Select/SelectValue.vue
@@ -10,7 +10,7 @@ export interface SelectValueProps extends PrimitiveProps {
 </script>
 
 <script setup lang="ts">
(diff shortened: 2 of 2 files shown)


zernonia (changes requested):
Thanks for the clear root-cause write-up and the regression test. I confirmed it fails on v2 and passes with your change.

One concern: because `selectedLabel` holds on to the last label whenever the lookup comes back empty, it can also show a stale label outside the close handoff. If a controlled `modelValue` changes to a value with no matching item, or the selected item is removed from the list, the trigger keeps the old text (both show the placeholder on v2).

Could we either close the gap in `SelectContent` (items unmount before the fragment remounts them on the `renderPresence` timeout), or only keep the previous label while the model value is unchanged and `optionsSet` is empty?

Small nits: the describe should reference #2767 rather than #2638, and the `ref(resolvedLabel.value)` seed isn't needed with `immediate: true`.

PR #2773, opened 2026-06-28
fix(useEmitAsProps): prevent TS2589 with void emit parameters

### LINKED ISSUE #2769: [Bug]: useForwardPropsEmits "Type instantiation is excessively deep and possibly infinite."
### Environment

```bash
Reka UI version: 2.10.1
```

-

### Steps to reproduce

Running `vue-tsc` on the following will trigger a typescript error for `useForwardPropsEmits`

```vue
<script setup lang="ts">
import { reactiveOmit } from '@vueuse/core'
import {
  useForwardPropsEmits,
  type CollapsibleContentEmits,
  type CollapsibleContentProps
} from 'reka-ui'

import { type HTMLAttributes } from 'vue'

const props = defineProps<
  CollapsibleContentProps & { class?: HTMLAttributes['class'] }
>()
const emits = defineEmits<CollapsibleContentEmits>()

const delegatedProps = reactiveOmit(props, 'class')

const forwarded = useForwardPropsEmits(delegatedProps, emits)
</script>
```

### Describe the bug

Running vue-tsc on `useForwardPropsEmits` causes this error:

`Type instantiation is excessively deep and possibly infinite.`

Most likely due to this commit: https://github.com/unovue/reka-ui/commit/f76c5e13

Also see this issue: https://github.com/unovue/reka-ui/issues/2766 where the author seems to have solved it by removing `void` from emitted events. Reka-UI uses void internally however, e.g.:

```ts
type CollapsibleContentEmits = {
    contentFound: [void];
}
```

### Expected behavior

No typescript errors.

_No response_

developers-universe-1:
Fixes #2769

The `OverloadUnionRecursive` type from `@vue/shared` enters infinite recursion when processing emit signatures that contain a `void` parameter (e.g. `defineEmits<{ contentFound: [void] }>()`). This triggers TS2589: "Type instantiation is excessively deep and possibly infinite."

This PR works around the upstream TypeScript limitation by adding a depth limit (10) to the recursive type. This caps the recursion before TypeScript bails out, while still preserving correct type inference for all practical emit shapes (single / overloaded / void args).

### Changes
- Added `Depth extends any[] = []` parameter to `OverloadUnionRecursive` with a depth cap of 10
- Added regression type tests for `void`-argument emits in `useEmitAsProps.test-d.ts`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TypeScript handling for event-forwarding helpers, including events with no payload value.
  * Reduced the chance of deep type recursion issues that could trigger compiler errors in complex projects.
  * Preserved correct optional handler parameter typing when forwarding emitted events.

### DIFF
--- a/packages/core/src/shared/useEmitAsProps.test-d.ts
+++ b/packages/core/src/shared/useEmitAsProps.test-d.ts
@@ -31,6 +31,10 @@ declare const singleEmit: SingleEmit
 type LooseEmit = (name: string, ...args: any[]) => void
 declare const looseEmit: LooseEmit
 
+// Void-argument emit (regression test for #2769)
+type VoidEmit = (e: 'contentFound', payload: void) => void
+declare const voidEmit: VoidEmit
+
 // useEmitAsProps — overloaded emit maps each event to its `onXxx` handler prop
 {
   const result = useEmitAsProps(overloadedEmit)
@@ -48,6 +52,16 @@ declare const looseEmit: LooseEmit
   type _args = Expect<Equal<Parameters<NonNullable<Result['onClose']>>, [reason: string]>>
 }
 
+// useEmitAsProps — void argument emit (regression test for #2769)
+// Note: OverloadUnion normalises `void` params to optional, but this is
+// still type-safe — calling the handler with or without args is valid.
+{
+  const result = useEmitAsProps(voidEmit)
+  type Result = typeof result
+  type _keys = Expect<Equal<keyof Result, 'onContentFound'>>
+  type _args = Expect<Equal<Parameters<NonNullable<Result['onContentFound']>>, [payload?: void | undefined]>>
+}
+
 // useEmitAsProps — loose signature stays accepted (backwards compatibility)
 {
   const result = useEmitAsProps(looseEmit)
@@ -56,6 +70,7 @@ declare const looseEmit: LooseEmit
 
 // EmitAsProps utility — emit signature to handler props
 type _emitAsProps = Expect<Equal<keyof EmitAsProps<OverloadedEmit>, 'onUpdate:modelValue' | 'onChange'>>
+type _emitAsPropsVoid = Expect<Equal<keyof EmitAsProps<VoidEmit>, 'onContentFound'>>
 
 // useForwardPropsEmits — fixtures
 interface DemoProps { foo: string, bar: boolean }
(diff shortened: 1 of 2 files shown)


JarvisH:
This PR fixes the problems I had with type checking 🙏

zernonia (changes requested):
Thanks for digging into this — the root-cause analysis is right, and a depth cap is a reasonable workaround for the upstream `OverloadUnionRecursive` behavior. One issue with the chosen value though.

**The cap of `10` fires earlier than TypeScript's own instantiation limit, so it silently truncates emit maps that work today.**

I extracted the type in isolation and compared patched vs. unpatched across increasing overload counts:

| Overloads | `v2` (unpatched) | This PR (cap 10) |
|---|---|---|
| ≤5 | ✅ correct | ✅ correct |
| 6–8 | ✅ correct | ❌ silently drops keys |
| ≥9 | ❌ TS2589 | ❌ silently drops keys |

Concretely, running the real `ToastRootImplEmits` (8 events) through `EmitFn` → `EmitAsProps` on this branch drops `onClose`, `onEscapeKeyDown` and `onPause` — recursion peels from the end, so the *first-declared* events lose their handler props. Since `EmitAsProps` is `Partial<…>`, nothing errors at the definition site; it surfaces downstream as a spurious excess-property error when a wrapper forwards `onClose`. That's harder to debug than the TS2589 we're fixing.

**`16` works.** I tested 12 / 16 / 20 / 24 against both the `void` case and the 8-emit Toast case:

- `12` — still truncates the 8-emit case
- `16` — ✅ `void` fixed *and* full 8-emit fidelity
- `20`+ — TS2589 returns (TS's own limit)

So 16 is the usable window, and it's strictly better than the status quo on overload count. Suggested, with a comment so the value doesn't look arbitrary:

```ts
type OverloadUnionRecursive<
  TOverload,
  TPartialOverload = unknown,
  // Depth cap: a `void` param makes the `TPartialOverload extends TOverload`
  // termination check unreachable (#2769), so bound the recursion. 16 is the
  // largest value that still fits TS's own instantiation budget (20+ re-triggers
  // TS2589); supports emit maps up to ~8 events, beyond which the union
  // truncates silently.
  Depth extends any[] = [],
> = Depth['length'] extends 16
```

**Test coverage:** the new fixtures are 1–2 overloads, which pass at any cap ≥ 4 — so the suite can't catch this truncation or guard a future change to the constant. Could you add something like:

```ts
type ManyEmit = EmitFn<{
  close: []
  escapeKeyDown: [event: KeyboardEvent]
  pause: []
  resume: []
  swipeStart: [event: SwipeEvent]
  swipeMove: [event: SwipeEvent]
  swipeCancel: [event: SwipeEvent]
  swipeEnd: [event: SwipeEvent]
}>
type _many = Expect<Equal<
  keyof EmitAsProps<ManyEmit>,
  'onClose' | 'onEscapeKeyDown' | 'onPause' | 'onResume'
  | 'onSwipeStart' | 'onSwipeMove' | 'onSwipeCancel' | 'onSwipeEnd'
>>
```

That's the assertion that fails at 10 and passes at 16.

**Separate follow-up (not for this PR):** we declare three `[void]` payloads that really mean "no payload" — `PopperContent` `placed`, `CollapsibleContent` `contentFound`, `Toast/FocusProxy` `focusFromOutsideViewport`. Switching those to `[]` removes the trigger for every reported repro, since all of them cite our own types. It's a signature change for anyone typing those handlers explicitly, so it deserves its own PR — but this depth cap is still worth having for user emits that are genuinely `void`.

Everything else looks good — style, `Expect`/`Equal` usage, and test placement all match the existing file, and there's no runtime change.

PR #2774, opened 2026-06-29
fix(aria): omit controls for unmounted popup content

### LINKED ISSUE #2597: [Bug]: aria-controls references unmounted content on Select/Combobox/DropdownMenu/Popover triggers (apply #2390 pattern)
### Summary

PR #2390 fixed `TabsTrigger` so it only renders `aria-controls` when the referenced `TabsContent` is mounted. The same pattern applies to other primitives whose content mounts lazily — currently `Select`, `Combobox`, `DropdownMenu`, and `Popover` triggers emit `aria-controls="reka-…"` pointing at content that isn't in the DOM until the popup opens.

### Why it matters

WAI-ARIA 1.2 §6.6.1 permits a collapsed combobox to reference a not-yet-rendered popup, so this isn't a screen-reader bug. But:

- HTML validators (e.g. `html-validate`'s `no-missing-references`) flag it as a dangling `idref` on every page load.
- The Tabs fix in #2390 establishes that Reka prefers to emit the attribute only when the target exists.
- Consumers currently have to either disable the validator rule project-wide or set `forceMount` (which has side effects — focus-scope can leak `aria-hidden="true"` onto sibling landmarks).

### Proposed change

Apply the #2390 approach to the triggers of:
- `SelectTrigger`
- `ComboboxTrigger` / `ComboboxAnchor`
- `DropdownMenuTrigger`
- `PopoverTrigger`

i.e., bind `aria-controls` only when the corresponding content component is mounted (track via context/provide as Tabs does).

### Alternatives considered

- `forceMount` — works, but trips focus-scope siblings with `aria-hidden` at rest.
- Disabling validator rules — hides real issues elsewhere.
- Rendering an empty `<template :id>` placeholder — requires consumer boilerplate per primitive.

yan-ad:
Closes #2597

## Description
Triggers and combobox inputs rendered `aria-controls` unconditionally, so it pointed at an id that either did not exist in the DOM yet or was still `""` because the content component had not minted it.

This aligns every remaining trigger with the pattern #2883 established for Popover and Select: **mint the content id where it is read, and only emit `aria-controls` while the popup is open.**

- **Combobox / Autocomplete** — input and trigger mint `contentId` and gate `aria-controls` on `open`. Also fixes a statically open `ComboboxRoot` rendering `aria-controls=""` on its input, since the non-reactive id was previously assigned by `ComboboxContent` after the input had already rendered.
- **Menu sub-trigger** — covers DropdownMenu, ContextMenu and Menubar submenus.
- **NavigationMenu** — trigger uses the id its item already computes instead of one assigned after mount.
- **Collapsible / Accordion** — content is always in the DOM, so `aria-controls` stays; the trigger now mints the id itself instead of rendering `""` until the first toggle.

Popover and Select are unchanged here (already fixed by #2883). Tabs is unchanged (already gated on registered content; its SSR id issue is #2866). Id minting order is preserved, so SSR ids and the Accordion hydration test are unaffected.

## Verification
- `pnpm --filter reka-ui exec vitest run` — 99 files, 2109 tests pass
- `pnpm --filter reka-ui type-check`
- `pnpm lint`
- The 7 new/updated `aria-controls` tests all fail against `v2` without the component changes.

## Summary by CodeRabbit

* **Accessibility**
  * Improved ARIA relationships across comboboxes, autocomplete, menus, navigation menus, accordions, and collapsible content.
  * Controls now reference their associated content with stable identifiers and expose `aria-controls` only while related content is open.
  * Improved submenu accessibility behavior.

* **Bug Fixes**
  * Improved keyboard, IME composition, autocorrect, and CJK interaction handling in comboboxes.
  * Improved automatic scrolling to highlighted items across virtualized, deferred, and force-mounted popup content.


princebhanwra-collationhub:
Confirming this also resolves a distinct manifestation from the closed-state dangling `aria-controls` in #2597: a **statically-open** `ComboboxRoot` (the command-palette / always-open combobox pattern) renders `aria-controls=""` **persistently on the open input**, which axe-core 4.12.1 flags as `aria-required-attr`, impact **critical** — not just a page-load validator warning.

Root cause on 2.10.1: `ComboboxRoot` provides `contentId` as a non-reactive string that `ComboboxContent` assigns lazily (`contentId ||= useId(...)`). With `open` static, nothing re-renders the already-rendered `ComboboxInput`, so it keeps the empty value it read before `ComboboxContent`'s setup ran. (Typing heals it — the id exists; the input just never re-read it.)

Minimal repro: `<ComboboxRoot open>` with `<ComboboxInput>` before `<ComboboxContent position="inline">`. Verified in Chromium 148. This PR's reactive `contentElement` gate should fix the static-open case too, since the ref change forces the input to re-render and re-read `contentId` — might be worth a test alongside the unmounted-content ones (the current coverage seems to exercise the open/close *transition*, which the always-open case never hits).

bpmore:
I came to #2597 from the other end — an accessibility audit of a production app (Statamic 6's control panel, reka-ui 2.9.5) — and was about to write this same fix before finding your PR. Rather than duplicate it, here is independent verification that it does what it claims, plus one thing that may be blocking it.

## It fixes the case I reported

I checked out `a9d2d4c` and ran the Combobox suite:

```
pnpm --filter reka-ui exec vitest run src/Combobox/Combobox.test.ts
Test Files  1 passed (1)
     Tests  43 passed (43)
```

Then, on the same checkout, I reverted **only** the two `aria-controls` bindings — `ComboboxInput.vue` and `ComboboxTrigger.vue` back to `:aria-controls="rootContext.contentId"` — and left everything else, including your test, in place:

```
× should only render aria-controls while content is mounted
  AssertionError: expected '' to be undefined
Tests  1 failed | 42 passed (43)
```

That empty string is exactly the symptom I measured in a real app. Your test catches it, and the fix is what makes it pass — it isn't passing incidentally.

## The `ComboboxInput` half matters more than the issue suggests

#2597 frames this as a validator complaint on the grounds that WAI-ARIA 1.2 §6.6.1 lets a *collapsed* combobox reference a not-yet-rendered popup. That holds for the trigger. It does not hold for `ComboboxInput`, which in 2.9.5 keeps `aria-controls=""` **even while the listbox is open and in the DOM**:

```json
{ "inputAriaControls": "", "inputAriaExpanded": "true",
  "listboxId": "reka-combobox-content-v-5", "match": false }
```

A `role="combobox"` advertising `aria-expanded="true"` with no resolvable `aria-controls` is a real screen reader problem, not a lint warning. Because `contentId` is a plain property rather than a ref, the input renders once with `""` and never updates; the trigger only appears to work because it re-renders when `open` flips and picks up the mutated string as a side effect. Your `contentElement` ref fixes both, and is a better shape than the `Ref<string>` route I had suggested on the issue.

## One likely blocker

This branch is behind `v2`. `ComboboxRoot.vue` on `v2` now imports `./useComboboxContentPositioning`, which doesn't exist at this PR's base — I hit it while trying to mix the two trees. That's probably why the mergeability shows as unknown. A rebase on `v2` may be all this needs.

## Also worth folding in

@eBrou noted on the issue that **Collapsible and Accordion** share the root cause and aren't covered here. Worth confirming whether they're in scope for this PR or a follow-up, since #2522 attempted Collapsible separately and was closed.

Happy to help with the rebase or extend coverage to Collapsible/Accordion if that's useful — say the word rather than me pushing at your PR uninvited.

zernonia:
Thanks @yan-ad for the fix and the tests, and @bpmore / @princebhanwra-collationhub for the verification — I've picked this up to get it landed.

Since this PR was opened, #2883 fixed Popover and Select with a smaller pattern (mint the content id on the trigger, gate `aria-controls` on `open`), so I merged `v2` in and reworked the remaining components to match rather than adding `contentElement` plumbing to each root context. Your tests are kept; a few new ones were added.

Now covered in this PR:
- Combobox / Autocomplete (including the statically-open `aria-controls=""` case — thanks for the repro)
- Menu sub-triggers (DropdownMenu / ContextMenu / Menubar)
- NavigationMenu
- Collapsible / Accordion — @bpmore to your question: yes, folded in here. Their content is always in the DOM so `aria-controls` stays on, but the trigger now mints the id itself instead of rendering `""` until first toggle.

Tabs is left alone (already gated on registered content; the SSR id issue is #2866).

PR #2778, opened 2026-07-02
fix(Stepper): remove stale trigger element from totalStepperItems on unmount

### LINKED ISSUE #2777: Stepper: StepperTrigger leaks stale elements in totalStepperItems on unmount, inflating totalSteps
### Description

`StepperTrigger` registers its DOM element in the root `totalStepperItems` `Set` on mount and removes it on unmount:

```ts
const { forwardRef, currentElement } = useForwardExpose()

onMounted(() => {
  rootContext.totalStepperItems.value.add(currentElement.value)
})

onUnmounted(() => {
  rootContext.totalStepperItems.value.delete(currentElement.value)
})
```

`currentElement` comes from `useForwardExpose()`, whose value is derived from the component instance's `vnode.el`. By the time `onUnmounted` runs, that ref no longer points to the same element that was added on mount (it can be `null`/stale after teardown). So `delete()` receives a value that isn't the one stored in the `Set`, the element is never removed, and the `Set` keeps growing across mount/unmount cycles.

Anything derived from `totalStepperItems.size` (e.g. `totalSteps`, the total step count shown on mobile) then drifts.

### Reproduction

1. Render a `StepperRoot` whose set of `StepperItem`/`StepperTrigger` changes at runtime (conditionally swapping which triggers are rendered, or `v-if`-toggling the whole stepper).
2. Each time a trigger unmounts and a new one mounts, `totalStepperItems.size` increases instead of staying stable (+2 per cycle in my case).

### Root cause

`onUnmounted` deletes `currentElement.value`, but that value is not guaranteed to still equal the element added on mount — `useForwardExpose`'s element ref reads a torn-down `vnode.el` at unmount time.

### Proposed fix

Capture the exact element added on mount in a local variable and delete that same reference on unmount (skipping nullish values so `null` is never inserted into the `Set<HTMLElement>`).

xerox0213:
### Summary

`StepperTrigger` deleted `currentElement.value` on unmount, but that ref can be `null`/stale by teardown, so the element added on mount was never removed and the `Set` kept growing across mount/unmount cycles — inflating `totalSteps`.

This captures the exact element added on mount and deletes that same reference on unmount, also skipping nullish values so `null` is never inserted into the `Set<HTMLElement>`.

### Changes

- `StepperTrigger.vue`: track the registered element in a local variable; add/delete that exact reference.
- Added a regression test that mounts/unmounts triggers (5 → 3 → 5 → 1 → 5) and asserts `totalSteps` stays in sync.
- Added a `_StepperDynamic.vue` fixture that varies the rendered triggers at runtime.

Fixes #2777

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Stepper step counts so the displayed total stays accurate when steps are shown, hidden, or remounted repeatedly.
  * Improved Stepper trigger registration to prevent the total from drifting upward after toggling.
* **Tests**
  * Added regression coverage for dynamic Stepper rendering and repeated visibility changes.

### DIFF
--- a/packages/core/src/Stepper/Stepper.test.ts
+++ b/packages/core/src/Stepper/Stepper.test.ts
@@ -3,8 +3,10 @@ import userEvent from '@testing-library/user-event'
 import { render } from '@testing-library/vue'
 import { describe, expect, it } from 'vitest'
 import { axe } from 'vitest-axe'
+import { nextTick } from 'vue'
 import { useTestKbd } from '@/shared'
 import Stepper from './story/_Stepper.vue'
+import StepperDynamic from './story/_StepperDynamic.vue'
 
 const steps = [{
   step: 1,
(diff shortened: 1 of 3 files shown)


zernonia:
Thanks for the PR @xerox0213

PR #2780, opened 2026-07-06
fix(focus-scope): trap focus inside a shadow root

qudsiamehmood:
FocusScope attaches its focusin/focusout listeners to `document`. When the scope is teleported into a shadow root, those events are retargeted to the host, so `container.contains(target)` never matches and focus is not trapped — Tab escapes to the background.

Resolve the scope's root node via `getRootNode()` (the shadow root when teleported, otherwise the document) and attach the listeners there.

Related: #1667

### ❓ Type of change

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved focus trapping and focus redirection when a focus scope is rendered inside a shadow root.
  * Focus listeners now attach to the appropriate document or shadow root, with matching cleanup.
  * Improved focus detection across shadow DOM boundaries.

* **Tests**
  * Added coverage for focus scopes inside open shadow roots, including focus restoration after focus moves to light DOM elements.

### DIFF
--- a/packages/core/src/FocusScope/FocusScope.test.ts
+++ b/packages/core/src/FocusScope/FocusScope.test.ts
@@ -1,7 +1,7 @@
 import type { RenderResult } from '@testing-library/vue'
 import userEvent from '@testing-library/user-event'
 import { render } from '@testing-library/vue'
-import { afterEach, beforeAll, beforeEach, describe, expect, it, vi } from 'vitest'
+import { beforeAll, beforeEach, describe, expect, it, vi } from 'vitest'
 import { defineComponent, nextTick } from 'vue'
 import { FocusScope } from '.'
 import { ComboboxAnchor, ComboboxContent, ComboboxInput, ComboboxItem, ComboboxPortal, ComboboxRoot, ComboboxTrigger, ComboboxViewport } from '../Combobox'
@@ -149,6 +149,64 @@ describe('focusScope', () => {
   })
 
   // https://github.com/unovue/reka-ui/issues/2550
+  // https://github.com/unovue/reka-ui/issues/1667
+  // A trapped FocusScope teleported into a shadow root must keep focus inside
+  // even when focus escapes to a light-DOM element outside the shadow tree.
+  // Listening only on `document` misses intra-shadow moves (retargeted to the
+  // host); listening only on the shadow root misses the light-DOM escape (which
+  // never propagates into the root). We assert the behaviour, not the wiring.
+  describe('given a trapped FocusScope rendered inside a shadow root (#1667)', () => {
+    const ShadowScope = defineComponent({
+      components: { FocusScope },
+      template: `<FocusScope asChild loop trapped>
+        <form>
+          <input data-testid="inner-first" />
+          <input data-testid="inner-second" />
+        </form>
+      </FocusScope>`,
+    })
+
+    // document.activeElement stops at the shadow host; drill down to the real one.
+    function deepActiveElement(): Element | null {
+      let active = document.activeElement
+      while (active?.shadowRoot?.activeElement)
+        active = active.shadowRoot.activeElement
+      return active
+    }
+
+    it('recalls focus into the scope when it escapes to a light-DOM element', async () => {
+      const host = document.createElement('div')
+      const outsideButton = document.createElement('button')
+      outsideButton.textContent = 'outside'
+      document.body.append(host, outsideButton)
+
+      try {
+        const shadowRoot = host.attachShadow({ mode: 'open' })
+        const container = document.createElement('div')
+        shadowRoot.appendChild(container)
+
+        render(ShadowScope, { container })
+        await nextTick()
+
+        const innerFirst = shadowRoot.querySelector<HTMLInputElement>('[data-testid="inner-first"]')!
+        innerFirst.focus()
+        await nextTick()
+        expect(container.contains(deepActiveElement())).toBe(true)
+
+        // Focus escapes to a light-DOM button outside the shadow tree.
+        outsideButton.focus()
+        await nextTick()
+
+        // The trap must pull focus back inside the scope.
+        expect(container.contains(deepActiveElement())).toBe(true)
+      }
+      finally {
+        host.remove()
+        outsideButton.remove()
+      }
+    })
+  })
+
   describe('given a FocusScope with SelectTrigger inside Dialog (#2550)', () => {
     const DialogWithSelect = defineComponent({
       components: { DialogRoot, DialogTrigger, DialogContent, DialogTitle, SelectRoot, SelectTrigger, SelectValue, SelectContent, SelectItem },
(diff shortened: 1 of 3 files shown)


zernonia:
Thanks for digging into this — the diagnosis is right, and we're happy to take the fix on `v2`. A couple of changes needed before merge:

**1. Keep a `document` `focusin` fallback.**
Moving the listeners entirely to the shadow root swaps one blind spot for another: a `focusin` on a light-DOM element *outside* the shadow tree never propagates into the shadow root, so `handleFocusIn` no longer fires for the exact escape being fixed. That leaves only `handleFocusOut` as the recall path, and in practice the `focusout` recall loses the race to the pending focus commit — the recall that actually wins is the document `focusin`.

So please attach both: root listeners for intra-shadow moves, plus a document-level `focusin` when `root !== ownerDocument`, using a composed (deep) target check rather than `container.contains()`:

```ts
function handleDocumentFocusIn(event: FocusEvent) {
  if (focusScope.paused || !container)
    return
  // composed path target, not event.target (which is retargeted to the host)
  const target = event.composedPath()[0] as HTMLElement | null
  if (!containsComposed(container, target))
    focus(lastFocusedElementRef.value, { select: true })
}
```

where `containsComposed` walks up through shadow hosts (`node.getRootNode().host`) instead of stopping at the shadow boundary. This is the shape we landed on in #2794 on `v3`, so aligning here keeps the two branches consistent.

**2. Make the test behavioural.**
The current test spies on `addEventListener`, which asserts an implementation detail — it passes whether or not focus is actually trapped. jsdom doesn't retarget, but it *can* cover the scenario that matters: focus an element inside the shadow scope, focus an outside light-DOM button, assert focus returns inside the scope. That's the regression we care about.

Also, minor: the `host` element is appended to `document.body` and never removed, and the spy isn't restored — both leak into later tests in the file.

**Nits:**
- `getRootNode()` never returns `null`, so `?? document` only covers `container === undefined` — clearer as an explicit ternary, and the fallback document should be `container?.ownerDocument ?? document` for iframe consistency.

Once the document fallback is in, could you also confirm it traps in a real browser (Chrome/Safari) with a scope teleported into a shadow root? That's the part the test can't prove.

qudsiamehmood:
Thanks for the detailed review @zernonia  — that's a great catch on the document blind spot, the reasoning about the recall race makes total sense. I've pushed an update addressing all the points:

1. **Document focusin fallback:** The scope now keeps its root listeners for intra-shadow moves and, when root !== ownerDocument, additionally attaches a focusin listener on the owner document. That handler uses event.composedPath()[0] plus a containsComposed() helper (walking up through shadow hosts) rather than container.contains(), so a light-DOM escape is caught even though the event is retargeted to the host at the document level.

2. **Behavioural test:** Replaced the addEventListener spy with a real assertion: focus an element inside the shadow-mounted scope, move focus to a light-DOM button outside the shadow tree, and assert focus is pulled back inside. Also cleaned up the leaks — the host and the outside button are created/removed via beforeEach/afterEach, and there's no longer a spy left un-restored.

**Nits.** Replaced the ?? document fallback with an explicit ternary (container ? container.getRootNode() : ownerDocument), and the fallback document is now container?.ownerDocument ?? document for iframe consistency.

**Real-browser check:** I verified this in the app where the bug originally surfaced — it's a microfrontend that mounts its entire UI inside a shadow root, and it uses reka's Dialog (which relies on FocusScope with trapped) as a normal component. To test this branch's changes I built the patched reka-ui and applied it to the app via a pnpm patch (over the built dist), then ran the app in Chrome:

- Opened a modal Dialog (rendered, as always, inside the shadow root).
- Tab through the dialog's fields/buttons → focus cycles and stays within the dialog; it never lands on the page behind it.
- Shift+Tab from the first focusable → wraps to the last.
- Focus attempting to move to a light-DOM element outside the shadow tree is recalled into the dialog.

Before this change, Tab escaped straight to the background. With the fix applied, focus stays trapped. (Clicking outside still dismisses the dialog, which is the expected DismissableLayer behavior and unrelated to the trap.)

Let me know if you'd like any further changes.

MILLERMARRU (commented):
The core diagnosis is right and matches how focus events cross shadow boundaries per spec. Listening only on `document` misses focus moves inside the shadow root because those never need to bubble past the root to be observed from inside it, and listening only on the root misses the light-DOM escape because that event never enters the shadow tree at all. Attaching to both, with `composedPath()[0]` for the document-level listener to undo retargeting, is the right shape for the fix.

One gap I traced through by hand rather than running: this only fully works for `mode: 'open'` shadow roots, which is also the only mode the new test exercises (`attachShadow({ mode: 'open' })`). For `mode: 'closed'`, `composedPath()` called from a listener outside the shadow tree truncates at the host, per MDN it does not reveal internal nodes of a closed root to an external listener. So when focus moves between two elements that are both inside a closed-mode `container` (a completely normal Tab press within the trap), the event still bubbles to `document` and fires `handleDocumentFocusIn`, but `composedPath()[0]` there resolves to the shadow host, not the actual focused element. `containsComposed(container, host)` walks up via `getRootNode().host` starting from the host itself, which is already outside the shadow tree, so it returns `false`, and the handler treats a normal in-scope Tab move as an escape and yanks focus back via `lastFocusedElementRef`. That would fire on every internal focus change when the scope lives in a closed shadow root, not just on genuine escapes.

Might be a non-issue in practice if reka-ui only ever teleports into open roots internally, but if this is meant to support arbitrary consumer-supplied shadow roots, closed mode is a legitimate and spec-supported choice, worth either testing that case explicitly or noting the open-mode assumption in the comment above `containsComposed`.

qudsiamehmood:
Thanks for looking into this so carefully — closed mode is a fair thing to raise, and since you'd traced it by hand I went and ran it: jsdom, real Chrome, and real `Tab` keypresses sent to Chrome over CDP.

The truncation itself is real. In a closed root a document-level listener does get the host from `composedPath()`, and `containsComposed(container, host)` does return `false`. But the trigger you describe doesn't happen. When focus moves between two elements inside the same shadow root, the document isn't notified at all — from outside the root, focus was inside the host before the move and inside the host after it, so there's nothing new to report and `handleDocumentFocusIn` never runs.

Real Tab presses inside a closed root:

Tab: in1 -> in2  (both inside the closed root)
  docEvents  : []                                  <- nothing reaches the document
  rootEvents : [{ target: input#in2 }]

Tab: in3 -> #after  (leaving the closed root)
  docEvents  : [{ target: button#after, path0: button#after }]
  rootEvents : []

`document.hasFocus()` was `true` and light-DOM moves fired a document event every time, so those empty lines aren't a dead listener. The second block is the important one: the only events that reach `handleDocumentFocusIn` are real escapes, and those land in the light DOM where nothing is truncated — so it gets the true element. jsdom behaves identically.

The truncated value does turn up in one case: focus *entering* the scope from the page into a closed root. There the recall fires when it shouldn't, but it has no effect — the root listener runs first, so `lastFocusedElementRef` already points at the field that just received focus and the recall simply re-focuses it.

Closed roots are genuinely broken, though — in existing code rather than in this diff. `getActiveElement()` resolves focus by looking inside shadow roots, can't look inside a closed one, and so returns the host. Two things depend on that: the Tab loop compares it against the tabbable edges, so `focusedElement === last` never matches and focus stops wrapping; and `focusFirst` uses it to check whether focus moved, so on mount it walks every candidate and falls back to `focus(container)`. I measured both — an open root wraps and focuses the first field, a closed root does neither. That helper has ~26 non-test callers across Menu, Select, Combobox, Dialog, Tree and others, all comparing it against specific elements, so this isn't FocusScope-specific.

So there's nothing to change in this fix for closed roots — the blocker is `getActiveElement()`, which is outside the scope of this PR. Happy to note the limitation in a comment if you'd like it on record.

PR #2786, opened 2026-07-10
fix(TagsInput): keep disabled tags out of keyboard removal index

### LINKED ISSUE #2785: [Bug]: TagsInput removes the wrong tag via keyboard when a disabled item is present
### Environment

```bash
Development OS: macOS (darwin 25.5.0)
Node version: 22.14.0
Package manager: npm
Reka UI version: 2.10.0 (also reproducible on 2.10.1 — source unchanged)
Vue version: 3.5.38
Nuxt version: none (Vite + Vue 3 SPA)
CSS framework: tailwindcss
Client OS: macOS
Browser: Chrome (Chromium 140)
```

https://stackblitz.com/edit/vitejs-vite-jbs2n75d?file=src%2FApp.vue

### Steps to reproduce

Model starts as `['admin@acme.com', 'user@gmail.com']`, where `admin@acme.com` is a disabled item and `user@gmail.com` is removable.

1. Click the text input to focus it (leave it empty).
2. Press `Backspace` → the last removable tag (`user@gmail.com`) becomes selected/active.
3. Press **Backspace** again to delete the selected tag.

What happens: `admin@acme.com` (the disabled tag) was removed, and `user@gmail.com` remains.

### Describe the bug

When a `disabled` `TagsInputItem` is mixed with removable items, keyboard deletion (Backspace/Delete on the selected tag) removes the wrong tag. With the disabled item positioned before the removable one, the disabled tag is removed, even though it can never be selected (keyboard nav correctly skips disabled items).

Root cause is an index mismatch in `TagsInputRoot`. Keyboard nav computes the selected index against a `disabled-filtered` collection:

    // onInputKeydown
    const collection = getItems().map(i => i.ref).filter(i => i.dataset.disabled !== ''); // removable-only
    const index = collection.findIndex(i => i === selectedElement.value);
    handleRemoveTag(index);

…but `handleRemoveTag` applies that index to the **full** model:

    function handleRemoveTag(index) {
      if (index !== -1) {
        const collection = getItems().filter(i => i.ref.dataset.disabled !== '');
        modelValue.value = modelValue.value.filter((_, i) => i !== index); // full-model index
        emits('removeTag', collection[index].value);
      }
    }

The filtered-collection index ≠ the full-model index whenever a disabled item precedes the target, so the wrong element is removed. The same mismatch removes the wrong *removable* tag when the disabled item comes after the target.

Suggested fix: resolve the selected item to its index in `modelValue` (unfiltered) before removing

I'm happy to open a PR for this if it helps.

### Expected behavior

Deleting the selected (removable) tag via keyboard removes that exact tag. Disabled items are never removed regardless of their position in the model.

For the repro: after step 3, the model should be `['admin@acme.com']` (removed `user@gmail.com`), not `['user@gmail.com']`.

Confirmed in reka-ui source `dist/TagsInput/TagsInputRoot.js` (`handleRemoveTag` + `onInputKeydown`), unchanged between 2.10.0 and 2.10.1. `TagsInputItem` correctly sets `data-disabled=""` (so the collection filter works and nav skips disabled items) — the defect is purely the index reuse in `handleRemoveTag`.

JoMarchant:
Closes #2785

## Description

`TagsInput` removed the **wrong** tag on keyboard delete when a `disabled` `TagsInputItem` was present.

The keyboard handler computes the selected tag's index against a **disabled-filtered** collection:

```ts
const collection = getItems().map(i => i.ref).filter(i => i.dataset.disabled !== '') // removable-only
const index = collection.findIndex(i => i === selectedElement.value)
handleRemoveTag(index)
```

…but `handleRemoveTag` applied that index to the **full** model:

```ts
modelValue.value = modelValue.value.filter((_, i) => i !== index)
```

Since `getItems()` already excludes disabled items, the filtered index doesn't match the model index whenever a disabled item exists. Symptoms:

- A disabled tag placed **before** removable ones gets deleted instead of the selected tag.
- With only disabled tags left, Backspace could still remove one.

`TagsInputItemDelete` was unaffected because it already resolves the model index by value (`modelValue.findIndex(isEqual(...))`).

## Fix

- Resolve the keyboard-selected tag to its index in `modelValue` **by value** (via `ohash`'s `isEqual`, mirroring `TagsInputItemDelete`) before removing.
- Make `handleRemoveTag` emit the value that was actually removed (`modelValue[index]`) instead of indexing the disabled-filtered collection — this also fixes an incorrect `removeTag` payload from the delete button when disabled items are present.

Navigation still uses the disabled-filtered collection, so disabled items remain unreachable/unselectable.

## Repro (before this PR)

```vue
<script setup lang="ts">
import { ref } from 'vue'
import { TagsInputRoot, TagsInputItem, TagsInputItemText, TagsInputItemDelete, TagsInputInput } from 'reka-ui'
const modelValue = ref(['admin@acme.com', 'user@gmail.com'])
const isDisabled = (v: string) => v === 'admin@acme.com'
</script>

<template>
  <TagsInputRoot v-model="modelValue">
    <TagsInputItem v-for="item in modelValue" :key="item" :value="item" :disabled="isDisabled(item)">
      <TagsInputItemText />
      <TagsInputItemDelete v-if="!isDisabled(item)" />
    </TagsInputItem>
    <TagsInputInput placeholder="Add..." />
  </TagsInputRoot>
</template>
```

Focus the input → Backspace (selects `user@gmail.com`) → Backspace again.
- Before: `admin@acme.com` (disabled) is removed.
- After: `user@gmail.com` is removed. ✅

## Tests

Added a colocated test fixture (`story/_TagsInputDisabled.vue`) and cases in `TagsInput.test.ts`:
- keyboard delete removes the selected removable tag (not the disabled one),
- the disabled tag is never removed when it's the only tag left.

All `TagsInput` tests pass; `type-check` and lint are green.

### DIFF
--- a/packages/core/src/TagsInput/TagsInput.test.ts
+++ b/packages/core/src/TagsInput/TagsInput.test.ts
@@ -5,6 +5,7 @@ import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { axe } from 'vitest-axe'
 import { nextTick } from 'vue'
 import TagsInput from './story/_TagsInput.vue'
+import TagsInputDisabled from './story/_TagsInputDisabled.vue'
 import TagsInputObject from './story/_TagsInputObject.vue'
 
 describe('given default TagsInput', () => {
@@ -349,5 +350,41 @@ describe('given a TagsInput with objects', async () => {
       expect(tags.length).toBe(2)
       expect(tags[1].text()).toBe('hello')
     })
+  })
+})
+
+describe('given TagsInput with a disabled item before a removable one', () => {
+  // @ts-expect-error we return empty object
+  window.getComputedStyle = () => ({})
+  let wrapper: VueWrapper<InstanceType<typeof TagsInputDisabled>>
+  let input: DOMWrapper<HTMLInputElement>
+  let rootComponent: Omit<VueWrapper, 'exists'>
+
+  beforeEach(() => {
+    wrapper = mount(TagsInputDisabled, { attachTo: document.body })
+    rootComponent = wrapper.getComponent({ name: 'TagsInputRoot' })
+    input = wrapper.find('input')
+    input.element.focus()
+  })
+
+  it('removes the selected removable tag, not the disabled one', async () => {
+    // First Backspace selects the last removable tag, second removes it.
+    await input.trigger('keydown', { key: 'Backspace' })
+    await input.trigger('keydown', { key: 'Backspace' })
+
+    const tags = wrapper.findAll('[data-reka-collection-item]')
+    expect(tags.map(tag => tag.text())).toEqual(['Disabled'])
+    expect(rootComponent.emitted('removeTag')?.[0]?.[0]).toEqual('Removable')
+  })
+
+  it('does not remove the disabled tag when it is the only remaining tag', async () => {
+    await input.trigger('keydown', { key: 'Backspace' })
+    await input.trigger('keydown', { key: 'Backspace' })
+    // Only the disabled tag remains; further backspaces must not remove it.
+    await input.trigger('keydown', { key: 'Backspace' })
+    await input.trigger('keydown', { key: 'Backspace' })
+
+    const tags = wrapper.findAll('[data-reka-collection-item]')
+    expect(tags.map(tag => tag.text())).toEqual(['Disabled'])
   })
 })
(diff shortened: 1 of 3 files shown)


zernonia:
LGTM! Thanks @JoMarchant !

Issue #2725, opened 2026-06-17
feat: full Shadow DOM support


zernonia:
> Part of the v3 roadmap — #2721
> **Phase 2 — Built on the foundation**

## Problem

Shadow DOM isn't fully supported:

- ~70 direct `document` / `window` references.
- 9 files resolve aria refs via `document.getElementById`.
- Scroll-lock, focus-guards, and Splitter mutate `document.body` / `head`.

## Goal

Every DOM lookup / listener respects the element's `getRootNode()` / `ownerDocument`; components work inside shadow roots.

## Approach

- Route aria / id queries through the root node.
- Scope focus-guards & scroll-lock per root.
- Use adopted stylesheets for injected CSS.
- Add a shadow-DOM test harness.

## Key files

- `shared/useBodyScrollLock.ts`
- `shared/useFocusGuards.ts`
- `shared/getActiveElement.ts` (already partially aware)
- `FocusScope/FocusScope.vue`
- `DismissableLayer/utils.ts`
- `Splitter/utils/style.ts`
- `ScrollArea/ScrollAreaScrollbarImpl.vue`
- plus the 9 `getElementById` aria-wiring sites.

## Why it matters

Unblocks web-component / micro-frontend / design-system embedding use cases.

AntonioPolaGTP:
For consideration, we found a related bug for the select.




# Title

Combobox/SelectMenu: selecting an item does nothing in Shadow DOM (handleBlur uses document.getElementById)

# Body

## Description

When a `Combobox` (and therefore `SelectMenu`, and @nuxt/ui v4's `USelectMenu`) is rendered inside an
**open Shadow DOM**, clicking a `ComboboxItem` with the mouse often does not commit the selection — the
listbox just closes and the value is unchanged. It's intermittent (sometimes works on the 2nd/3rd
click). The same component works perfectly in the light DOM.

## Reproduction

1. Mount a Vue app inside an open shadow root, and teleport the combobox content within that shadow root:

   ```js
   const host = document.createElement('div')
   document.body.append(host)
   const shadow = host.attachShadow({ mode: 'open' })
   const mount = document.createElement('div')
   shadow.append(mount)
   createApp(App).mount(mount) // App renders a <ComboboxRoot> / <SelectMenu>, content teleported into the shadow root
   ```

2. Open the combobox and click an item **with the mouse**.

_(Minimal StackBlitz: attach a reproduction here.)_

**Expected:** the clicked item is selected and the value updates.

[...]

Issue #2722, opened 2026-06-17
feat: `useRender` rendering primitive


zernonia:
> Part of the v3 roadmap — #2721
> **Phase 1 — Foundation (rendering & headless core)**

## Problem

Every part renders through a `Primitive` wrapper component (`as` / `asChild`) — an extra component instance per part, and composables can't render without a host component.

## Goal

A `useRender({ defaultTagName, as, props, state, ref })` composable returning `{ tag, renderProps, renderless, ... }`, rendered via `<component :is="tag">`:

- Renderless mode via a `Slot` sentinel.
- Integrated prop / ref / state merging.
- No wrapper component instance.

## Approach

- Build the `useRender` + `mergeProps` / `useMergedRefs` primitives (inspired by Base UI — https://baseui-vue.com/docs/utils/use-render).
- Keep `Primitive` as a thin compat shim during migration.

## Why it matters

Foundation for the headless logic composables; removes a wrapper layer (perf); cleaner ref / root-node forwarding (also helps Shadow DOM).

Issue #2721, opened 2026-06-17
[Epic] v3 Roadmap


zernonia:
Tracking issue for **reka-ui v3**. This lays out scope and sequencing; each item links to its own issue.

## The dependency spine

These items are not a flat list — there's a foundation, a layer built on it, and one independent track:

```
[Epic] reka-ui v3 Roadmap
│
├─ Phase 1 — Foundation (rendering & headless core)
│   ├─ useRender — render via composable, drop the Primitive wrapper
│   └─ Headless logic composables for every component (depends on useRender)
│
├─ Phase 2 — Built on the foundation
│   ├─ Performance overhaul (DismissableLayer shared stack; render overhead)
│   └─ Full Shadow DOM support
│
└─ Independent track (any time)
    ├─ Per-component imports / fast dev (subpath exports)
    ├─ ESM-only build (drop CJS)
    └─ Promote all Alpha components to Stable
```

`useRender` is the render primitive the logic composables need. Performance and Shadow DOM both build on the new ref/render plumbing. Per-component imports, ESM-only, and Alpha→Stable are decoupled and can land any time.

## Core architectural decision

**Components stay exported with their public API intact**, except for the breaking surface listed below. v3 re-architects them *internally*:

- Each component's logic moves into a `useX()` **logic composable**.
- The `.vue` SFC becomes a thin shell composing `useX()` + `useRender`.
- `useRender` replaces the `Primitive` / `asChild` wrapper internally.
- The logic composables become a **new additive public API** for programmatic control.

This is an internal teardown plus an additive API — **not** a breaking rewrite of the component-level surface.

## Migration path

Two docs pages carry the migration, and both grow as breaking changes land: the human-readable [Migration to v3](https://reka-ui.com/docs/guides/migration-v3) guide, and [Migrating with an AI agent](https://reka-ui.com/docs/guides/migration-v3-agent), the same content rewritten as instructions for a coding agent (search patterns, per-selector decision rules, when to stop and ask, verification, a closing report). The llms plugin serves the agent page raw at `…/migration-v3-agent.md`, so `claude "Read <that url> and migrate this project"` is the whole workflow. The static `@reka-ui/codemod` package was tried in #2915/#2917 and removed in #2918: the mechanical `data-state` rewrite was the only part a regex could do, and every other v3 change needs judgment. Every PR that adds a breaking change updates both pages.

## Breaking surface (v3)

Everything that changes for a consumer, in one place:

[...]

PR #2790, opened 2026-07-12
perf(DismissableLayer): centralized layer stack manager (#2724, Tasks 1-7)

### LINKED ISSUE #2724: perf: overlay & render performance overhaul
> Part of the v3 roadmap — #2721
> **Phase 2 — Built on the foundation**

## Problem

`DismissableLayer` uses a global `reactive` stack with per-layer `Array.from(...).indexOf()` and separate global `pointerdown` / `focusin` / `click` listeners **per layer** (`packages/core/src/DismissableLayer/DismissableLayer.vue:57`). Wrapper instances add further render overhead.

## Goal

- A **single shared listener** + **one centralized stack manager** (inspired by Base UI's approach).
- Remove wrapper render overhead via `useRender`.
- Measurable reductions in overlay interaction cost and render count.

## Approach

- Centralize the layer stack; consolidate the global listeners.
- Lean on the `Primitive` wrapper removal from the `useRender` work.
- Add benchmarks to quantify the wins.

## Why it matters

Overlays are the heaviest interactive surface — correctness + speed.

zernonia:
## Overlay & render performance overhaul (Tasks 1–7)

Implements #2724 — Phase 2 of the [reka-ui v3 roadmap #2721](https://github.com/unovue/reka-ui/issues/2721). Replaces `DismissableLayer`'s **per-layer** global listeners and per-read O(n) scans with a **single centralized stack manager**.

> **Scope:** Tasks 1–7. Task 8 (removing the `Primitive` wrapper via `useRender`) is deferred until #2722 merges into `v3` — it's a small mechanical follow-up.

### The problem
Each open `DismissableLayer` attached its **own** `pointerdown` + `focusin` (document) + `keydown` (window) listeners, and every read recomputed `Array.from(reactiveSet).indexOf(...)`. N open overlays = 3N listeners and O(n) work per interaction, plus reactive-Set churn.

### The change
- **`layerStack.ts` — a plain-module manager (transport only).** Two separate registries model two populations with different lifetimes:
  - `layers` (presence-driven ordered stack) → Escape routing, indexing, pointer-events accounting.
  - `outsideSubscribers` (setup/`enabled`-driven) → pointer/focus dispatch.
  It installs **exactly one** shared listener of each kind (torn down when its registry empties), builds **one** `querySelectorAll` snapshot per event, arms subscribers, and owns the touch-click deferral.
- **The composables become subscribers.** `usePointerDownOutside`/`useFocusOutside` keep their exact signatures (`Editable` compiles unchanged) but port their `handlePointerDown`/`handleFocus` bodies **branch-for-branch** into subscriber closures.
- **Escape** moves from a per-layer `window` keydown (`onKeyStroke`) to one shared listener routing to the top *present* layer.
- **Reactive `context` removed.** `index`/`isPointerEventsEnabled`/`isBodyPointerEventsDisabled` derive from the manager (array `indexOf`, no `Array.from`). The `disableOutsidePointerEvents` body lock is a **ref-counted** manager helper (first disabling layer sets `none`, last restores — #2674 order-independent). Branches move to the manager.
- **Shadow-ready** (baked in for #2725): target reads go through `composedPath()[0]` (captured synchronously), threaded into `handleAndDispatchCustomEvent` (which now guards a null target — the original event target is nulled post-dispatch for shadow-origin events in real browsers).

### Proof of the win
`DismissableLayer.listeners.test.ts` asserts structurally: **10 layers → exactly 1 `pointerdown` + 1 `focusin` + 1 `keydown` listener**, **1 `querySelectorAll` per outside pointerdown regardless of layer count**, and full teardown on unmount. (jsdom benchmarks in `DismissableLayer.bench.ts` are directional; render count is parity, not reduced.)

### Testing
- **Full suite green: 1972 tests / 97 files.** Type-check clean.
- Every existing `DismissableLayer.test.ts` case passes **unchanged** — including the #2674 `disableOutsidePointerEvents` suite, the `present:false` tests, and the `isLayerExist` unit tests (export kept). Consumer sweep (Dialog, Drawer, Menu, Popover, Tooltip, HoverCard, Combobox, Select, NavigationMenu, Toast, Editable) all green.
- New: manager unit tests, stacked-Escape routing, listener/query-count regressions.

### Reviewed by an advisor before building
The manager design was pressure-tested against the current `utils.ts` semantics before the migrations were ported. Findings incorporated: focus uses a **fresh** `isLayerExist` after its double `nextTick` (snapshot is pointerdown-only); the armed-guard is the literal first statement of the pointer closure (no flag writes on that path); focus never checks `armed` or writes its flag; touch deferral is a per-subscriber `Map`; arming timers are cleared on reset; a liveness guard covers synchronous mid-dispatch unregistration.

### Behavior notes (documented, not regressions)
- Escape stays on `window` (matching `onKeyStroke`'s default), so nothing that stops propagation at document level is affected.
- Two touch/interleaving micro-quirks of the old per-layer `{once}` click model are preserved bug-for-bug; one real-browser cascading-dismissal interleaving now matches what the jsdom oracle already exhibited.

Refs #2724, #2721 · Follow-up: Task 8 after #2722 merges.

## Summary by CodeRabbit

* **Performance**
  * Improved `DismissableLayer` handling for stacked overlays by consolidating shared event processing and reducing redundant DOM queries.
  * Added performance benchmarks covering layer mounting, outside interactions, and Escape-key handling.

* **Bug Fixes**
  * Escape now consistently targets the topmost active layer.
  * Improved coordination of outside pointer and focus interactions across nested layers.
  * Preserved correct pointer-event restoration for nested overlays.

* **Tests**
  * Added coverage for shared listener lifecycle, event routing, query efficiency, and stack cleanup.

* **Documentation**
  * Added an implementation plan for the overlay rendering and interaction performance improvements.

### DIFF
--- a/docs/superpowers/plans/2026-07-11-2724-overlay-render-perf.md
+++ b/docs/superpowers/plans/2026-07-11-2724-overlay-render-perf.md
@@ -2,215 +2,237 @@
 
 > **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
 
-> Tracks GitHub issue **#2724** (Phase 2 — Built on the foundation), part of the reka-ui v3 roadmap **#2721**. The stack-manager work is independent of #2722 (`useRender`); only the final wrapper-removal step (Task 8) depends on it.
+> Tracks GitHub issue **#2724** (Phase 2), part of the reka-ui v3 roadmap **#2721**. Tasks 1–7 are independent of #2722; the wrapper-removal (Task 8) depends on `useRender`. **This plan lands before #2725 (Shadow DOM) and builds the shadow-safe composed-target reads into `layerStack.ts` from day one** (so #2725 does not re-patch `DismissableLayer/utils.ts`).
 
-**Goal:** Replace `DismissableLayer`'s per-layer global listeners (3+ `document`/`window` listeners **per open layer**) and per-read O(n) `Array.from(...).indexOf()` scans with a **single centralized stack manager** exposing exactly one shared set of `document` listeners regardless of layer count, plus benchmarks that quantify the wins. Then remove the wrapper render instance via `useRender`.
+**Goal:** Replace `DismissableLayer`'s per-layer global listeners (3 `document`/`window` listeners **per open layer**) and per-read O(n) `Array.from(...).indexOf()` scans with a **single centralized stack manager** exposing exactly one shared listener of each kind regardless of layer count, one `querySelectorAll` snapshot per event, plus benchmarks that quantify the wins. Then remove the wrapper render instance via `useRender`.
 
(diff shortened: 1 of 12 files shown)


zernonia:
@coderabbit review

PR #2792, opened 2026-07-12
feat(shadow-dom): root helpers + aria-id/label resolution (#2725)

zernonia:
## Shadow DOM support — root helpers + aria-id/label resolution

Implements #2725 — Phase 2 of the [reka-ui v3 roadmap #2721](https://github.com/unovue/reka-ui/issues/2721). Delivers the shadow-DOM **foundation** + the issue's headline fix: **every aria-id / label lookup respects the element's root node** so components work inside shadow roots.

> **Sequenced after #2724** (overlay perf): that PR already makes `DismissableLayer` shadow-safe (composed-target reads in `layerStack`), so this PR does **not** touch `DismissableLayer`/`utils.ts`. **Scope here:** the shared helpers + the id/label migrations. The deeper per-root restructures (FocusScope root-attachment, per-root scroll-lock & focus-guards, Splitter adopted-cursor) are cleanly separable follow-ups (see below).

### Foundation (`@/shared`)
- **`getRootNode` / `getOwnerDocument` / `getOwnerWindow`** — resolve an element's root, cross-realm-safe (nodeType/host duck-typing), SSR-safe (global-document fallback).
- **`getElementByIdFrom` — dual-root:** the anchor's root first (in-shadow content), then the owner document (content Teleported out to `body`). Light-DOM behavior is unchanged; shadow behavior finds both. This is what keeps Combobox/NavigationMenu working when their content portals out.
- **`getEventTarget`** — `composedPath()[0]`, the un-retargeted deep target for open roots.
- **`injectStyle`** — `adoptedStyleSheets` (lazy-detected) with a nonce-threaded `<style>` fallback (older Safari + the branch jsdom exercises).
- **`createShadowHost`** test harness.

### Migrations (light-DOM behavior preserved)
The ~11 aria-id / label lookups now route through the root node:
- **Dialog / Drawer** title & description warnings → `getElementByIdFrom(contentElement, id)` (no more false "missing title" warnings inside a shadow root).
- **Combobox** content-focus containment → dual-root `getElementByIdFrom` + `getActiveElement(anchor)` (stays open on inner focus, in-root **or** body-portalled).
- **NavigationMenu** content entry/exit → root-scoped.
- **Radio / Checkbox / Switch** `[for]` label lookups → `getRootNode(el).querySelector(...)` (root-scoped is also semantically correct — `for`/`id` never cross shadow boundaries).
- **`getActiveElement`** gains an optional `anchor` (starts from the anchor's owner document).

### Testing
[...]

PR #2794, opened 2026-07-12
fix(shadow-dom): shadow-safe FocusScope, per-root focus guards, retarget-safe listeners

zernonia:
## Summary

Second and final part of shadow-DOM support (`Refs #2725`). Part 1 (#2792) landed the shared helpers (`getRootNode`/`getOwnerDocument`/`getEventTarget`/`getElementByIdFrom`/`injectStyle`/…). This PR wires the remaining components that still assumed the light DOM, so an open shadow-rooted component actually works — most importantly the focus trap.

## What changed

**FocusScope — shadow-safe trap** (`FocusScope.vue`, `getRootNode.ts`)
- Attaches `focusin`/`focusout` to the container's **root** (the `ShadowRoot` in a web component) so event targets aren't retargeted to the host, **plus** a `document` `focusin` fallback: a `focusout`-driven recall loses the race to the pending focus commit, and the recall that actually wins fires as a `document` `focusin` — which a shadow-root listener never sees.
- Adds `containsComposed()` for the composed-target scope-membership check across shadow boundaries.
- Light DOM (`root === ownerDocument`) behavior is unchanged.

**Focus guards — per-root** (`useFocusGuards.ts` + 4 callers)
- `useFocusGuards(element?)` refcounts guards **per root** (`WeakMap`) and places them at the `ShadowRoot` edges when scoped; the default (no arg) stays document-body for back-compat.
- Combobox/Popover/Menu/Select content pass their `currentElement` so a shadow-mounted trap's edge guards live inside the same root.

**Retarget-safe listeners + injected styles**
- `ScrollAreaScrollbarImpl` / `SelectContentImpl`: document-level wheel/pointer handlers read the **composed** target so a shadow-mounted content isn't mistaken for "outside" itself; body user-select toggles via `getOwnerDocument`.
- `Splitter` cursor: global cursor style now goes through `injectStyle` (per-document, nonce-aware) instead of a module-global `<style>`.

**Body scroll lock — intentionally unchanged.** A shadow root shares the host `document.body`, which is the correct scroll surface, so `useBodyScrollLock` needs no scoping.

## Known limitations (documented in CONTRIBUTING)

Portalled content still defaults to `body` (outside the root) unless `ConfigProvider`'s `teleportTo` points inside the shadow root; closed shadow roots, cross-boundary aria idrefs for teleported content, `HoverCard` selection/scroll (`composed: false`), `useHideOthers`, and the iOS `touchmove` overflow walk remain browser-only / genuinely unfixable.

## Testing

- New shadow-root tests: FocusScope (intra-shadow moves don't yank; escapes recall into the scope; removed element recovers to container) and `useFocusGuards` (back-compat + per-root refcount at shadow edges).
- Full suite green: **2033 tests / 105 files**; `type-check` clean.

Refs #2725

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved focus trapping and restoration for components rendered inside Shadow DOM.
  - Enhanced focus guards, outside-click detection, scrolling, and text selection behavior across shadow boundaries.
  - Improved focus handling when nested elements are removed or focus moves outside a component.
  - Added safer support for components hosted in different documents.

- **Documentation**
  - Expanded Shadow DOM guidance, including Teleport behavior and known browser limitations.


zernonia:
@coderabbit review

Issue #2723, opened 2026-06-17
feat: headless logic composables for every component


zernonia:
> Part of the v3 roadmap — #2721
> **Phase 1 — Foundation (rendering & headless core)** · depends on #2722 (`useRender`)

## Problem

Component logic is locked inside `.vue` SFCs; users can't drive behavior programmatically.

## Goal

Every component family exposes a `useX()` composable holding all logic/state. The `.vue` SFC becomes a thin shell composing `useX()` + `useRender`.

- **Components keep their current public API** — this is an internal re-architecture.
- The logic composables become a **new additive public API** for programmatic control (the headless-of-the-headless layer).

## Approach

- Establish the pattern on one family end-to-end first, then convert all ~60 families.
- Full commitment: every component ships a composable by v3.

## Why it matters

Programmatic control; testability; smaller component shells.

zernonia:
## Proposal for decision: public export policy for `useX()` composables

Drafted against the `v3` head. The review of the three converted families found the additive API inconsistent: `useSwitch`/`useTabs` are public, Menu's composables are exported from nowhere, and `stateToDataAttrs` is not exported even though the recipe tells consumers to call it. This settles the policy before more families convert. Recommendations are marked; amend freely.

### Current state

| Symbol | Family index | Package root | Notes |
|---|---|---|---|
| `useSwitch`, `UseSwitchProps/Return`, `SwitchState`, **`PartSurface`** | `Switch/index.ts:11-17` | yes | `PartSurface` reaches the root only by riding on Switch |
| `useTabs`, `UseTabsProps/Return`, `getTabsTriggerSurface`, `getTabsContentSurface`, `Tabs*State` | `Tabs/index.ts:23-31` | yes | `baseId` defaults to the literal `'reka-tabs'` (`useTabs.ts:135`) |
| `useMenuRoot`, `useMenuContent`, `getMenuItemBaseSurface`, `getMenuItemSelectSurface` | not in `Menu/index.ts` | no | Menu is only reachable via `reka-ui/internal` (`internal.ts:7`) |
| `PartState`, `stateToDataAttrs` | `shared/index.ts:17` | **no**, the `./shared` block in `index.ts` omits them | recipe step 4 tells consumers to call it |
| `useRender`, `PrimitiveState`, `StateAttributesMapping` | | yes | `getStateAttributes` is private (`useRender.ts:17`) and duplicates `stateToDataAttrs` |
[...]

PR #2795, opened 2026-07-12
feat(Tabs): headless useTabs composable (recipe-validation gate for #2723)

zernonia:
## Summary

Second pilot of the headless-composables effort (`Refs #2723`) and the **recipe-validation gate**: converts **Tabs** — the collection / per-item / roving-focus archetype — after Switch (the form-control archetype, already merged). Tabs forces the questions Switch dodged, so the migration recipe is now validated on two structurally-different families and ready to batch-apply.

Components keep their **exact public API** — this is an internal re-architecture that additively exposes the logic as a new public API. `pnpm docs:gen` reports **zero** change to `Tabs*.md` (no prop/emit/slot/context diff).

## What changed

**`useTabs()` composable** (`Tabs/useTabs.ts`) — owns selection + content registration, returns per-part surfaces:
- a fixed `root` / `list` surface, plus
- `getTriggerSurface(value)` / `getContentSurface(value)` — per-item **getters**, because each item's id/aria/state derive from `(context, value)` and a fixed root blob can't express them.

**Per-item builders are exported, context-PURE functions** (`getTabsTriggerSurface`/`getTabsContentSurface`). The crux of the collection archetype: a descendant part SFC injects the **context**, not the `useTabs()` return (exactly like `SwitchThumb`). So both the SFC and `useTabs().getTriggerSurface` call the **same** builder — one derivation, zero drift.

**The SFCs become thin shells** and keep exactly what belongs to them:
- `TabsRoot` — `useVModel` (controlled/uncontrolled), `useDirection`, `useId` (SSR-stable base id), and `provideTabsRootContext(context)`.
- `TabsTrigger` — the `<RovingFocusItem as-child>` wrapper (arrow-key nav) + tag-dependent `type`.
- `TabsContent` — the `<Presence>` wrapper, the registration lifecycle, `hidden`, and the mount-animation `style`.

All surfaces bind via `mergeProps(...)` so a consumer's listeners chain with the internal handlers.

**`PartSurface` relocated to `@/shared`** (the shared headless contract), re-exported from `useSwitch` so Switch's public path stays byte-identical.

## Bounded "headless" (documented)

Component-wrapper behaviors stay wrappers in v1: a standalone `useTabs()` gets ids/aria/selection, but **not** roving-focus keyboard nav (`RovingFocus`) or content mount/unmount (`Presence`) unless the consumer also composes those components.

## Testing

- **Characterization first** (committed against unmodified Tabs, green pre-refactor): id/aria-controls/aria-labelledby pairing, `data-state` active/inactive, automatic vs. manual activation, **consumer mousedown-listener chaining**, orientation, disabled.
- **`useTabs` unit tests** (15): selection, per-item surfaces, handler bodies (left/right/ctrl mousedown, Enter/Space, focus-activation), no `data-*` in `props`, context shape, public export.
- Full suite **2051 / 105 files** green; `type-check` clean; `docs:gen` diff empty.

## Recipe

`docs/superpowers/recipes/2723-headless-composable-recipe.md` is amended and marked **validated on Switch + Tabs**, with the per-item builder pattern, the `MaybeRefOrGetter` value rule, the `inheritAttrs`-by-part-count note, and the `useId`-stays-in-the-shell rule.

Follow-up (mirrors the Switch pilot, which shipped no user-facing composable docs yet): a public docs page + standalone story for the `useX()` API, best done as one batch across families.

Refs #2723

## Summary by CodeRabbit

* **New Features**
  * Added a headless Tabs composable with reusable trigger and content surface builders.
  * Exposed Tabs state, context, and surface APIs through the package entry point.
  * Improved Tabs accessibility behavior, including ARIA relationships, keyboard activation, orientation, disabled states, and stable IDs.

* **Bug Fixes**
  * Improved consistency of rendered Tabs attributes and state indicators.

* **Tests**
  * Added comprehensive coverage for Tabs selection, activation modes, accessibility, registration, and interaction behavior.

* **Documentation**
  * Updated headless composable migration guidance and collection/per-item patterns.


zernonia:
@coderabbit review

PR #2806, opened 2026-07-17
fix: process Android soft-keyboard composition live for filters

benjamincanac:
## Summary
- Follow-up to #2665: Android soft keyboards (Gboard, Samsung, …) keep the current Latin word in composition for autocorrect, so deferring all `input` until `compositionend` froze live filtering in Autocomplete, Combobox, ListboxFilter, and DropdownMenuFilter ([nuxt/ui#6717](https://github.com/nuxt/ui/issues/6717)).
- `useComposing` now exposes `shouldDeferInput`: on Android, plain-text composition updates filter live; IME-script composition (Han/Hiragana/Katakana/Hangul/Bopomofo) and desktop Pinyin preedit still wait for commit.
- The verdict is **sticky per composition session**: romaji-mode Japanese IMEs compose Latin before converting to kana in a single session, so classifying each `compositionupdate` independently would let the session flip back to live mid-conversion.
- Keydown still gates on `isComposing` so IME candidate navigation is unchanged.

## Scope
- `shouldDeferInput` is applied to the four filter inputs only (Autocomplete, Combobox, ListboxFilter, DropdownMenuFilter) — those are the ones that filter on every `input`. TagsInput, NumberField, ColorField and PinInput keep gating on raw `isComposing`: they act on keydown, not on live input, so the Android autocorrect freeze doesn't apply to them. Deliberately out of scope.

## Known limitation
- On Android, the *leading* plain-text updates of a transliterating IME are indistinguishable from soft-keyboard autocorrect — romaji-mode Japanese (`ka` → `か`), Vietnamese Telex, Thai and Indic transliteration all emit Latin before any IME script appears — so those first keystrokes filter live. No content-based heuristic can separate them. Once IME script appears the session defers and stays deferred; the committed value is always correct, only transient popup state is wrong.

## Test plan
- [x] Unit tests for `useComposing` (Android Latin live, desktop Latin defer, CJK defer, empty update)
- [x] Mid-session `romaji → kana → romaji` transition stays deferred once IME script is seen, and the verdict does not leak into the next session
- [x] Autocomplete / Combobox / Listbox / DropdownMenuFilter coverage for the same matrix
- [ ] Smoke on Android (Chrome/Firefox/Samsung): type `Br` in a Combobox/InputMenu and confirm Broccoli appears before Return
- [ ] Smoke CJK IME on Android: confirm filtering still waits until composition commit
- [ ] Smoke desktop Pinyin: confirm Latin preedit still does not filter early

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved autocomplete, combobox, dropdown filter, and listbox filtering behavior during IME composition.
  * Android plain-text autocorrect (e.g., desktop Pinyin preedit) now updates suggestions/results live when appropriate.
  * CJK/“real IME” composition is deferred until `compositionend`, preventing premature filtering and value updates.
* **Tests**
  * Expanded IME composition regression coverage for Android vs non-Android across autocomplete, combobox, dropdown filtering, and listbox filtering.
  * Added dedicated tests for `useComposing`’s new defer/update composition behavior and updated test lifecycle imports.

### DIFF
--- a/packages/core/src/Autocomplete/Autocomplete.test.ts
+++ b/packages/core/src/Autocomplete/Autocomplete.test.ts
@@ -1,6 +1,6 @@
 import type { DOMWrapper, VueWrapper } from '@vue/test-utils'
 import { mount } from '@vue/test-utils'
-import { beforeEach, describe, expect, it, vi } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { axe } from 'vitest-axe'
 import { nextTick } from 'vue'
 import { handleSubmit, sleep } from '@/test'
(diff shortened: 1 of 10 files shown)


zernonia:
Nice fix, and the test matrix is solid — full suite, `vue-tsc` and lint are all green on my end. One change I'd like before merging.

**The IME classification isn't sticky, so it can flip back mid-session.** Japanese IMEs (Gboard / Google Japanese Input in romaji mode) compose Latin *first*, then convert to kana within the same composition session:

```
compositionupdate 'k'   → shouldDeferInput: false   ← live
compositionupdate 'ka'  → shouldDeferInput: false   ← live
compositionupdate 'か'  → shouldDeferInput: true    ← deferral resumes, too late
```

End-to-end in a Combobox on Android UA, that's not just a stale value — the popup opens on the partial romaji and stays stuck on "no results" through the whole conversion phase:

```
after romaji 'ka':  listbox OPEN, data-empty=""
after kana  'かん':  listbox OPEN, data-empty=""
```

Backspacing out of kana back into romaji flips it live again too (`'かんじ'` → `'kan'` → `shouldDeferInput: false`).

Suggested fix — makes the IME verdict sticky per session:

```ts
const sawImeScript = ref(false) // reset alongside isImeComposition in handleCompositionStart

if (imeScriptRE.test(event.data)) {
  isImeComposition.value = true
  sawImeScript.value = true
}
else if (isAndroid() && !sawImeScript.value) {
  isImeComposition.value = false
}
```

I applied this locally: it fixes the flip-back and all 279 tests across the affected suites still pass. Worth adding a test for the mid-session `romaji → kana` transition, since the current `'Br'` → `'香'` case only asserts the final state.

Note this **doesn't** fix the initial romaji leak — the first keystrokes of a romaji-mode IME are genuinely indistinguishable from Android autocorrect, so no content-based heuristic separates them. I think accepting that is defensible (Latin users get the fix; Android CJK users get a transient wrong popup that self-corrects on commit), but could you call it out explicitly in the code comment? Right now the comment reads as if genuine IME composition always defers, and it narrowly doesn't.

Same reasoning applies to Vietnamese Telex / Thai / Indic transliteration on Android — all compose non-CJK scripts and will now filter live on partial input.

Minor: `shouldDeferInput` is only applied to the four filter inputs — TagsInput/NumberField/ColorField/PinInput still gate on raw `isComposing`. Fine to scope out, just worth saying so in the description so it reads as deliberate.

gavgrego:
Any thought when this will be merged in?  Having some issues with Combobox input detection on some Android devices with predictive text turned on.

gavgrego:
👍

PR #2807, opened 2026-07-18
fix(rating): reset hover state on mouse leave in case of hoverable

al1maher:
### 🔗 Linked issue
https://github.com/nuxt/ui/issues/6746

### ❓ Type of change

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Hover previews now clear automatically when the pointer leaves the rating control.
  * Hovering a rating restores the selected value when the pointer leaves.
  * Disabled ratings remain unaffected by hover reset behavior.

### DIFF
--- a/packages/core/src/Rating/Rating.test.ts
+++ b/packages/core/src/Rating/Rating.test.ts
@@ -56,6 +56,34 @@ describe('given a default Rating', () => {
   })
 })
 
+describe('given a hoverable Rating', () => {
+  let wrapper: VueWrapper<InstanceType<typeof Rating>>
+  let root: DOMWrapper<HTMLElement>
+  let radios: DOMWrapper<HTMLElement>[]
+
+  beforeEach(() => {
+    document.body.innerHTML = ''
+    wrapper = mount(Rating, { attachTo: document.body, props: { defaultValue: 1, hoverable: true, length: 3 } })
+    root = wrapper.find('[role=radiogroup]')
+    radios = wrapper.findAll('[role=radio]')
+  })
+
+  it('should preview the hovered rating', async () => {
+    await fireEvent.mouseEnter(radios[2].element)
+    expect(radios[1].attributes('data-state')).toBe('active')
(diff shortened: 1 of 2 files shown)


zernonia:
Thanks for the fix — the bug is real and the approach is right. A few things to adjust before merge:

### 1. Remove the unused context additions

`hoverable` and `resetHoveredRating` are added to `RatingRootContext` and provided, but nothing consumes them: `resetHoveredRating` closes over `hoverable` directly and is only called from `RatingRoot`'s own template. Since `injectRatingRootContext` is publicly exported from `Rating/index.ts`, this becomes permanent public surface with no consumer.

Please revert both context entries (the `RatingRootContext` fields and the `provideRatingRootContext` keys) and keep `resetHoveredRating` as a plain local function.

### 2. Drop the guards in `resetHoveredRating`

```js
if (disabled.value || !hoverable.value)
  return
```

These are the same conditions that prevent `hoveredRating` from ever being set, so in the normal path they're redundant — and in the edge cases they hurt: if `hoverable` or `disabled` is toggled *while* a preview is showing, the reset is permanently short-circuited and the stale preview can never be cleared. Cleanup should be unconditional:

```js
function resetHoveredRating() {
  hoveredRating.value = 0
}
```

### 3. Please add a regression test

`Rating.test.ts` has no hover coverage today, which is why this regressed. Something small covering:

- `hoverable` + `mouseenter` on an indicator → `data-state="active"` for steps ≤ hovered
- `mouseleave` on the root → indicators fall back to reflecting `modelValue`

### 4. Optional: `pointerleave` over `mouseleave`

Every other hover-exit in the library uses `pointerleave` (usually via `whenMouse`) — see `TooltipTrigger`, `ListboxRoot`, `NavigationMenu*`, `useGraceArea`. `mouseleave` is defensible here since the enter side already uses `@mouseenter`, so this is a nice-to-have rather than a blocker.

zernonia:
@coderabbit review

PR #2809, opened 2026-07-22
fix(Toast): remove viewport listeners when a toast is dismissed

### LINKED ISSUE #2782: [Bug]: Toast viewport listeners are never removed (watchEffect cleanup returned instead of registered) — every dismissed toast leaks its DOM subtree
### Environment

- reka-ui: 2.10.1 (bug still present on `main`)
- Vue: 3.5.x

### Bug description

`ToastRootImpl.vue` registers two listeners on the toast viewport inside a `watchEffect` and tries to clean them up by **returning** a cleanup function from the effect callback:

```ts
watchEffect((cleanupFn) => {
  const viewport = providerContext.viewport.value
  if (viewport) {
    // ...
    viewport.addEventListener(VIEWPORT_PAUSE, handlePause)
    viewport.addEventListener(VIEWPORT_RESUME, handleResume)
    return () => {
      viewport.removeEventListener(VIEWPORT_PAUSE, handlePause)
      viewport.removeEventListener(VIEWPORT_RESUME, handleResume)
    }
  }
})
```

Vue ignores the return value of a `watchEffect` callback — cleanup must be registered through the `onCleanup` argument (here declared as `cleanupFn` but never called). This looks like a React `useEffect` habit surviving the port from Radix.

As a result, the `VIEWPORT_PAUSE`/`VIEWPORT_RESUME` listeners added for **every toast are never removed**. Because the viewport `<ol>` is long-lived, each leaked listener closure retains the dismissed toast's component context, including its `currentElement` — i.e. **every toast ever shown leaks its detached DOM subtree and 2 listeners for the lifetime of the provider**.

Same class as #2420 (ScrollArea) and #2367 (Select), fixed by #2744 / #2638.

### Evidence

Measured on a production build with CDP forced-GC loops: a page showing one toast per navigation cycle grew ~+21 DOM nodes and +3.5 listeners per cycle sustained over 12 cycles. Heap-snapshot retainer paths (ephemeron-aware BFS to GC roots) anchor every detached toast `<li>` at:

```
toast-viewport <ol> (live) -> EventListener (VIEWPORT_PAUSE/RESUME)
  -> handlePause/handleResume closure -> component context
  -> currentElement computed ref _value -> detached toast <li> subtree
```

After patching the two lines below, the same measurement is flat (net 0 per cycle).

### Minimal reproduction

Any app with `ToastProvider`: show + dismiss N toasts, force GC, inspect the viewport element's listeners (`getEventListeners($0)` in DevTools) — 2×N listeners remain; heap snapshot shows N detached toast subtrees.

### Suggested fix

Register the cleanup through the effect's `onCleanup` argument instead of returning it:

```ts
watchEffect((onCleanup) => {
  const viewport = providerContext.viewport.value
  if (viewport) {
    // ...
    viewport.addEventListener(VIEWPORT_PAUSE, handlePause)
    viewport.addEventListener(VIEWPORT_RESUME, handleResume)
    onCleanup(() => {
      viewport.removeEventListener(VIEWPORT_PAUSE, handlePause)
      viewport.removeEventListener(VIEWPORT_RESUME, handleResume)
    })
  }
})
```

Happy to open a PR with this change if useful.

lazerg:
Each toast registers pause and resume listeners on the shared viewport inside a `watchEffect`, but the teardown was returned from the effect callback instead of being registered through its cleanup argument, so `watchEffect` silently discarded it. Every dismissed toast therefore left its listeners and detached `<li>` subtree attached to the viewport, leaking on each open/close cycle. This registers the cleanup through the `watchEffect` cleanup callback so the listeners are removed when the toast unmounts.

Fixes #2782

## Summary by CodeRabbit

* **Bug Fixes**
  * Toast notifications now properly remove viewport event listeners when dismissed.
  * Improved cleanup of pause and resume behavior to prevent lingering event handlers.

### DIFF
--- a/packages/core/src/Toast/Toast.test.ts
+++ b/packages/core/src/Toast/Toast.test.ts
@@ -1,10 +1,11 @@
 import type { DOMWrapper, VueWrapper } from '@vue/test-utils'
 import { findByText, fireEvent } from '@testing-library/vue'
 import { mount } from '@vue/test-utils'
-import { beforeEach, describe, expect, it } from 'vitest'
+import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { axe } from 'vitest-axe'
 import { nextTick } from 'vue'
 import Toast from './story/_Toast.vue'
+import { VIEWPORT_PAUSE, VIEWPORT_RESUME } from './utils'
 
 const CLOSE_TEXT = 'Close'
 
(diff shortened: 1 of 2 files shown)


zernonia:
Thanks for the PR @lazerg !

PR #2810, opened 2026-07-22
fix(Toast): remove aria-hidden from tabbable focus proxy

### LINKED ISSUE #2776: [Bug]: Toast focus proxies still render `aria-hidden="true"` + `tabindex="0"` (aria-hidden-focus, incomplete fix of #2486)
### Description

Follow-up to #2486 / #2488.

The `aria-hidden-focus` part of #2486 is **still reproducible on the latest release (`2.10.1`)**. PR #2488 removed the *explicit* `aria-hidden="true"` from `FocusProxy.vue`, but the Toast viewport focus sentinels are still rendered with `aria-hidden="true"`, because of `VisuallyHidden`.

`Toast/FocusProxy.vue` renders a focusable sentinel:

```vue
<VisuallyHidden tabindex="0" ...>
```

and `VisuallyHidden.vue` defaults `feature` to `'focusable'`, which unconditionally sets `aria-hidden`:

```vue
withDefaults(defineProps<VisuallyHiddenProps>(), { as: 'span', feature: 'focusable' })
// ...
:aria-hidden="feature === 'focusable' || feature === 'fully-hidden' ? 'true' : undefined"
```

So the head/tail viewport sentinels are still emitted as:

```html
<span aria-hidden="true" tabindex="0" style="position: fixed; ..."></span>
```

which is exactly the `aria-hidden-focus` violation (a focusable element must not be `aria-hidden`). `feature="focusable"` is inherently contradictory here: the element is visually hidden **and** focusable (`tabindex="0"`) **and** `aria-hidden`.

The sibling fixes from #2488 (removing `role="alert"` / `aria-atomic` from the toast `<li>`) did land correctly — only the focus-proxy case remains.

### Suggested direction

The `FocusProxy` needs a visually-hidden-but-focusable sentinel that does **not** set `aria-hidden`. Since the sentinel is empty there is nothing to announce anyway — same rationale as radix-ui/primitives#2584. Options:

- a new `VisuallyHidden` `feature` value that is focusable without `aria-hidden`, or
- have `FocusProxy` render without `aria-hidden` (e.g. explicit `:aria-hidden="undefined"`).

### Reproduction

Render an open `<ToastRoot>` inside `<ToastProvider>` + `<ToastViewport>` and run axe-core (or Storybook's `@storybook/addon-a11y`). `aria-hidden-focus` (impact: serious) is reported on the two focus proxies.

Confirmed against `reka-ui@2.10.1` by inspecting the published `dist/VisuallyHidden/VisuallyHidden.js` and `dist/Toast/FocusProxy.js`, both of which still produce the behavior above.

### System Info

- reka-ui: 2.10.1 (also reproduced on 2.10.0)

lazerg:
The Toast focus proxies render with both `aria-hidden="true"` and `tabindex="0"`, which axe flags as an `aria-hidden-focus` violation since a focusable element must not be hidden from the accessibility tree. The `aria-hidden` comes from `VisuallyHidden`'s default `feature="focusable"` — #2488 removed the explicit attribute from the proxy but left this one in place. Passing `:aria-hidden="undefined"` on the proxy keeps it a tabbable focus sentinel without being hidden.

Fixes #2776

## Summary by CodeRabbit

* **Accessibility**
  * Improved hidden form field behavior by excluding non-interactive inputs from assistive technology.
  * Preserved accessibility and focusability for visually hidden content using the default `focusable` mode.
  * Added explicit `fully-hidden` behavior for content that should be excluded from the accessibility tree and tab order.
  * Improved focus handling for toast notifications and navigation menus.

* **Documentation**
  * Clarified the available visually hidden content modes and their accessibility behavior.

### DIFF
--- a/packages/core/src/Toast/FocusProxy.vue
+++ b/packages/core/src/Toast/FocusProxy.vue
@@ -10,7 +10,14 @@ const providerContext = injectToastProviderContext()
 </script>
 
 <template>
+  <!--
+    `VisuallyHidden` defaults to `aria-hidden="true"` (feature="focusable"), but this
+    proxy is a deliberately tabbable (`tabindex="0"`) focus sentinel — a focusable
+    element must not be `aria-hidden` (axe `aria-hidden-focus`). Override it back to
+    `undefined` so the attribute is not rendered.
+  -->
   <VisuallyHidden
+    :aria-hidden="undefined"
     tabindex="0"
     style="position: fixed"
     @focus="(event: FocusEvent) => {
--- a/packages/core/src/Toast/FocusProxy.vue
+++ b/packages/core/src/Toast/FocusProxy.vue
@@ -7,15 +7,14 @@ const emits = defineEmits<{
 }>()
 
 const providerContext = injectToastProviderContext()
+
+// `VisuallyHidden` defaults to `aria-hidden="true"` (feature="focusable"), but this
+// proxy is a deliberately tabbable (`tabindex="0"`) focus sentinel, and a focusable
+// element must not be `aria-hidden` (axe `aria-hidden-focus`). Override it back to
+// `undefined` so the attribute is not rendered.
 </script>
 
 <template>
-  <!--
-    `VisuallyHidden` defaults to `aria-hidden="true"` (feature="focusable"), but this
-    proxy is a deliberately tabbable (`tabindex="0"`) focus sentinel — a focusable
-    element must not be `aria-hidden` (axe `aria-hidden-focus`). Override it back to
-    `undefined` so the attribute is not rendered.
-  -->
   <VisuallyHidden
     :aria-hidden="undefined"
     tabindex="0"
--- a/packages/core/src/Toast/Toast.test.ts
+++ b/packages/core/src/Toast/Toast.test.ts
@@ -49,6 +49,21 @@ describe('given a default Toast', () => {
       expect(proxy.getAttribute('aria-hidden')).toBeNull()
   })
 
+  it('should proxy focus out of the viewport on shift+Tab', async () => {
+    await fireEvent.click(trigger.element)
+    await findByText(document.body, 'Scheduled: Catch up')
+
+    // Shift+Tab on the viewport hands focus to the head proxy, which bounces it
+    // onto the first toast. The viewport only reaches the proxy through its
+    // template ref, so this covers the ref resolving to the proxy element
+    // itself rather than to some other node.
+    const viewport = document.querySelector<HTMLElement>('ol[tabindex="-1"]')
+    expect(viewport).toBeTruthy()
+    await fireEvent.keyDown(viewport!, { key: 'Tab', shiftKey: true })
+
+    expect(document.activeElement?.closest('li')).toBeTruthy()
+  })
+
   it('should pass axe accessibility tests', async () => {
     expect(await axe(document.body)).toHaveNoViolations()
 
--- a/packages/core/src/Toast/Toast.test.ts
+++ b/packages/core/src/Toast/Toast.test.ts
@@ -32,6 +32,23 @@ describe('given a default Toast', () => {
     expect(toastElement?.getAttribute('tabindex')).toBe('0')
   })
 
+  it('should have focus proxies that are focusable but not aria-hidden', async () => {
+    await fireEvent.click(trigger.element)
+    await findByText(document.body, 'Scheduled: Catch up')
+
+    // The viewport renders head/tail focus proxies: tabbable sentinels
+    // (tabindex="0") that catch focus entering or leaving the viewport and
+    // redirect it onto a toast. A tabbable element must not be aria-hidden,
+    // otherwise axe reports an `aria-hidden-focus` violation.
+    const proxies = Array.from(
+      document.querySelectorAll<HTMLElement>('[tabindex="0"]'),
+    ).filter(el => el.style.position === 'fixed')
+
+    expect(proxies).toHaveLength(2)
+    for (const proxy of proxies)
+      expect(proxy.getAttribute('aria-hidden')).toBeNull()
+  })
+
   it('should pass axe accessibility tests', async () => {
     expect(await axe(document.body)).toHaveNoViolations()
 


MILLERMARRU (commented):
Checked `VisuallyHidden.vue` itself to confirm the comment's claim isn't overstated. Its `aria-hidden` binding is `feature === 'focusable' || feature === 'fully-hidden' ? 'true' : undefined`, and since `feature` is typed as exactly `'focusable' | 'fully-hidden'` with no other option, that condition is true for every valid value of the prop, there's no way to use `VisuallyHidden` and end up without `aria-hidden="true"` short of overriding it from the outside the way this PR does. So this isn't a narrow edge case in a rarely-used mode, it's the component's only behavior.

The actual bug this fixes is a real one and not just a lint nag either, `aria-hidden` on something with `tabindex="0"` means a sighted keyboard user can still tab onto the element while assistive tech is told to skip it entirely, so a screen reader user tabbing through the page would land on an element that announces nothing, silently confusing about where focus went. That's exactly what axe's `aria-hidden-focus` rule catches, and it's the right rule to be catching this on.

The new test doesn't just check the fix exists, it locates the proxies by their actual runtime characteristics (`tabindex="0"` plus `position: fixed`, matching how `FocusProxy` renders) rather than a class name or test id, so it's asserting against the real behavior a screen reader would encounter rather than an implementation detail that could drift out of sync with the actual markup.

zernonia:
Thanks for digging into this! The intent is right and the attribute change itself works — `:aria-hidden="undefined"` does override `VisuallyHidden`'s own binding through fallthrough-attr merge. I confirmed the new test fails on `v2` and passes here, and type-check/lint are clean.

However, the explanatory HTML comment breaks the component, so this needs a change before merging.

---

### 🔴 The template comment turns `FocusProxy` into a fragment root

`packages/core/src/Toast/FocusProxy.vue:13-18` places an HTML comment *before* the root element. Vue does not strip template comments here — the built output (`dist/Toast/FocusProxy.js`) contains:

```js
return openBlock(), createElementBlock(Fragment, null, [
  createCommentVNode("\n    `VisuallyHidden` defaults to ..."),
  createVNode(unref(VisuallyHidden_default), { "aria-hidden": void 0, tabindex: "0", ... })
], 2112)
```

So the component now has two root nodes instead of one.

**In dev:** `$el` resolves to the fragment's text anchor rather than the span. `ToastViewport.vue:178` assigns `headFocusProxyRef` via `unrefElement(node)`, so `ToastViewport.vue:109` throws:

```
TypeError: headFocusProxyRef.value?.focus is not a function
```

Reproduced by shift+Tab from the focused viewport with a toast open — passes on `v2`, throws on this branch.

**In production it's worse.** Vue's dev-root-fragment handling is entirely behind `process.env.NODE_ENV !== "production"` (`runtime-core.esm-bundler.js:6601`), so `root` stays the `Fragment`, `shapeFlag & (1|6)` is `0`, and **no fallthrough attrs are applied at all** — silently, since the warning branch is also dev-only. That drops `tabindex="0"`, `style="position: fixed"`, the `@focus` handler, *and* the `:aria-hidden="undefined"` override. The focus sentinel stops working and the original bug isn't actually fixed.

**Fix:** move the comment into `<script setup>`. I verified this locally — template stays single-root, all 7 Toast tests pass, and the `$el` / shift+Tab path works again:

```vue
<script setup lang="ts">
const providerContext = injectToastProviderContext()

// `VisuallyHidden` defaults to `aria-hidden="true"` (feature="focusable"), but this
// proxy is a deliberately tabbable (`tabindex="0"`) focus sentinel — a focusable
// element must not be `aria-hidden` (axe `aria-hidden-focus`). Override it back to
// `undefined` so the attribute is not rendered.
</script>

<template>
  <VisuallyHidden
    :aria-hidden="undefined"
    tabindex="0"
    ...
```

---

### 🟡 Neither the new test nor the axe test catches the above

All 7 tests pass with the component broken. Two gaps:

- The new test asserts on DOM attributes via `document.querySelectorAll('[tabindex="0"]')`, which never exercises the `headFocusProxyRef` / `focus()` path — the behaviour the proxies exist for. Worth adding an assertion that shift+Tab on the viewport doesn't throw.
- The existing axe test (`Toast.test.ts:51-58`) *does* open the toast, yet passed on `v2` with the `aria-hidden` + `tabindex="0"` violation present. jsdom + vitest-axe isn't detecting `aria-hidden-focus` here, so that test gives false confidence for this class of issue. Not this PR's job to fix, but it explains why the explicit attribute assertion is needed.

### 🟡 Same pattern still present in NavigationMenu

`packages/core/src/NavigationMenu/NavigationMenuTrigger.vue:167-172` is identical — `<VisuallyHidden aria-hidden="true" :tabindex="0" @focus="...">`. It will trip the same axe rule whenever a nav menu is open. Fine to leave out of scope, but #2776 isn't fully closed without it.

### 🔵 Minor: root cause worth considering

Arguably the real defect is that `feature="focusable"` sets `aria-hidden="true"` at all — the name says focusable, and a focusable element can't be aria-hidden. Fixing it in `VisuallyHidden.vue` would cover NavigationMenu too, but it has a wide blast radius (Tooltip, DateField/TimeField, BubbleSelect, ToastAnnounce all depend on the current default). The local override is the safer call for a bugfix — just noting the tradeoff so it's deliberate.

---

Everything else in the diff is sound — just move that comment into the script block. 🙏

lazerg:
You're right, and it turned out worse than a style nit. Reproduced locally: compiling the template gives `createElementBlock(Fragment, ...)` with a `createCommentVNode` sitting in front of the `VisuallyHidden` vnode, and the viewport's `headFocusProxyRef` then lands on the fragment's text anchor. Shift+Tab from the viewport throws `TypeError: headFocusProxyRef.value?.focus is not a function` and focus never leaves `body`. Moving the note into `<script setup>` puts the template back to a plain `createBlock(VisuallyHidden)`, and focus reaches the toast again.

Took the test suggestion too. There's now a case that fires shift+Tab on the viewport and asserts focus ends up inside a toast. It fails on the fragment-root version and passes on this one, so the ref path is covered rather than just the rendered attributes. One thing worth flagging: jsdom swallows the exception thrown inside the listener, so an assertion phrased as "doesn't throw" passes either way. Checking where focus actually landed was the only reliable signal I found.

Full suite is green (1964 tests), `type-check` is clean, and eslint is clean over `src/Toast`.

Leaving `NavigationMenuTrigger` where it is, as you suggested. Happy to send a separate PR for it if you want #2776 closed out completely.

On the `VisuallyHidden` default: that skip was deliberate for the reason you gave. Tooltip, DateField/TimeField, BubbleSelect and ToastAnnounce all rely on the current behaviour, so changing it reads as its own discussion rather than something to fold into a bugfix.

zernonia:
Thanks for tracking this down. I pushed a commit on top of yours that moves the fix to the root: `VisuallyHidden`'s default `feature="focusable"` was rendering `aria-hidden="true"`, which is what forced the override here and left the same `aria-hidden-focus` violation on `NavigationMenuTrigger`'s focus proxy. It also broke the documented use of hidden text as a label (a button with only `<VisuallyHidden>Settings</VisuallyHidden>` had no accessible name).

With the root fix, only `fully-hidden` removes content from the accessibility tree (matching Radix), so the per-callsite override in `FocusProxy.vue` is no longer needed and I dropped it. Hidden native inputs that relied on the implicit attribute (DateField, TimeField, their range variants, PinInput) now set it explicitly, so their behavior is unchanged. Your Toast tests are kept as-is, and I added the equivalent assertion for NavigationMenu.

lazerg:
Agreed, the root cause fix is the right call. Dropping the FocusProxy override makes sense once VisuallyHidden itself no longer forces aria-hidden. Good catch on the hidden-text-as-label case too, I hadn't checked for that.

lazerg:
Closing this due to no activity.

PR #2813, opened 2026-07-23
fix(DateField): advance segment focus in DOM order in RTL

### LINKED ISSUE #2812: [Bug]: DateField/TimeField segment auto-advance moves backwards (or not at all) in RTL
### Environment

```
reka-ui: 2.10.1 (also present on main)
vue: 3.5.x
any browser / OS
```

Reproducible with any `DateFieldRoot` + `DateFieldInput` setup with `dir="rtl"` (steps below) — the linked PR adds regression tests that fail on main and reproduce it exactly.

### Steps to reproduce

1. Render a `DateFieldRoot` with `dir="rtl"` (the `dir` prop or `ConfigProvider`), any locale — `en-US` works (segments: month / day / year).
2. Click the first segment (month).
3. Type `2` (completes the month segment).

### Describe the bug

Focus does not advance to the day segment — it stays on month. Starting from a *middle* segment instead (e.g. day), completing it moves focus **backwards** (day → month) instead of forwards (day → year).

Root cause: `DateFieldRoot`'s `focusNext()` reuses `nextFocusableSegment`, whose index sign flips for RTL (`const sign = dir.value === 'rtl' ? -1 : 1`). That flip is correct for `ArrowLeft`/`ArrowRight` (spatial navigation), but auto-advance after filling a segment is *logical* navigation: the next segment to type is always the next one in DOM order (the locale's format order), exactly like Tab — which is why Tab behaves correctly while typing doesn't. From the first segment, `focusNext()` computes index `0 + (-1)` → `undefined` → no move; from a middle segment it moves backwards.

`DateRangeFieldRoot` and `TimeFieldRoot` duplicate the same pattern and have the same bug.

### Expected behavior

After completing a segment, focus advances to the next segment in the locale's format order (DOM order) regardless of writing direction — same order as Tab, same behavior as LTR.

Downstream report with video, via @nuxt/ui's `UInputDate` with the `fa-IR` locale: nuxt/ui#6414. Arrow-key navigation is unaffected (it should stay direction-aware).

61021:
### 🔗 Linked issue

Resolves #2812

### ❓ Type of change

### 📚 Description

`focusNext()` — the auto-advance that runs after a segment is completed — reused `nextFocusableSegment`, whose index sign flips for RTL (`const sign = dir.value === 'rtl' ? -1 : 1`). The flip is correct for `ArrowLeft`/`ArrowRight` spatial navigation, but auto-advance is *logical* navigation: the next segment to type is always the next one in DOM order (the locale's format order), exactly like Tab. In RTL this made completing the first segment move focus nowhere (index `-1`), and completing a middle segment move focus backwards.

`focusNext()` now advances `currentSegmentIndex + 1` in DOM order directly, in `DateFieldRoot`, `DateRangeFieldRoot` and `TimeFieldRoot` (all three duplicated the pattern). Arrow-key navigation is untouched and stays direction-aware.

Each component gets a regression test (default `en-US` locale with `dir="rtl"`, mirroring the existing typing tests); all three fail on `main` and pass with the fix. Full `DateField`/`DateRangeField`/`TimeField` suites pass (142 tests).

Downstream context: nuxt/ui#6414 (fa-IR `UInputDate`, report with video).

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic focus advancement for date, date-range, and time fields in right-to-left layouts.
  * Auto-advance now follows the locale/format DOM segment order, while arrow-key navigation remains direction-aware.
* **Tests**
  * Added RTL keyboard focus traversal coverage for DateField, DateRangeField, TimeField, and TimeRangeField, including year/month/day and hour/minute progression.

### DIFF
--- a/packages/core/src/DateField/DateField.test.ts
+++ b/packages/core/src/DateField/DateField.test.ts
@@ -443,6 +443,23 @@ describe('dateField', async () => {
     }
   })
 
+  it('advances focus through segments in DOM order when typing in RTL', async () => {
+    const { user, month, day, year } = setup({
+      dateFieldProps: {
+        dir: 'rtl',
+      },
+    })
+
+    await user.click(month)
+    expect(month).toHaveFocus()
+    await user.keyboard('{2}')
+    expect(day).toHaveFocus()
+    await user.keyboard('{19}')
+    expect(year).toHaveFocus()
+    await user.keyboard('{1980}')
+    expect(year).toHaveTextContent('1980')
+  })
+
   it('adjusts the hour cycle with the `hourCycle` prop', async () => {
     const { getByTestId, queryByTestId, user } = setup({
       dateFieldProps: {
--- a/packages/core/src/DateField/DateFieldRoot.vue
+++ b/packages/core/src/DateField/DateFieldRoot.vue
@@ -287,7 +287,10 @@ provideDateFieldRootContext({
   elements: segmentElements,
   setFocusedElement,
   focusNext() {
-    nextFocusableSegment.value?.focus()
+    // Auto-advance follows the segments' DOM order (the locale's format
+    // order) regardless of writing direction; only arrow-key navigation is
+    // direction-aware via nextFocusableSegment/prevFocusableSegment.
+    Array.from(segmentElements.value)[currentSegmentIndex.value + 1]?.focus()
   },
 })
 
--- a/packages/core/src/DateRangeField/DateRangeField.test.ts
+++ b/packages/core/src/DateRangeField/DateRangeField.test.ts
@@ -53,6 +53,21 @@ it('should pass axe accessibility tests', async () => {
 })
 
 describe('dateField', async () => {
+  it('advances focus through segments in DOM order when typing in RTL', async () => {
+    const { user, start } = setup({
+      dateFieldProps: {
+        dir: 'rtl',
+      },
+    })
+
+    await user.click(start.month)
+    expect(start.month).toHaveFocus()
+    await user.keyboard('{2}')
+    expect(start.day).toHaveFocus()
+    await user.keyboard('{19}')
+    expect(start.year).toHaveFocus()
+  })
+
   it('populates segment with value - `CalendarDate`', async () => {
     const { start, end } = setup({
       dateFieldProps: { modelValue: calendarDate },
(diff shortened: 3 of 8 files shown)


zernonia:
Thanks — the diagnosis is right, and the distinction you drew (auto-advance is logical/DOM order, arrow keys are spatial) is the correct one. Verified locally: the three new tests fail on `v2` and pass here, full suites green.

One gap: `TimeRangeField` has the identical bug and isn't covered. [`TimeRangeFieldRoot.vue:435`](https://github.com/unovue/reka-ui/blob/v2/packages/core/src/TimeRangeField/TimeRangeFieldRoot.vue#L435) still calls `nextFocusableSegment.value?.focus()`, and I confirmed on this branch that typing into `start-hour` with `dir="rtl"` leaves focus where it is. Same one-liner applies — its `currentSegmentIndex` already disambiguates start/end via `data-reka-time-range-field-segment-type`, so the plain `+1` is safe:

```ts
focusNext() {
  Array.from(segmentElements.value)[currentSegmentIndex.value + 1]?.focus()
},
```

Plus a test mirroring the `TimeField` one.

Minor: the `DateRangeField` test only walks `start.*` — extending it to `start.year → end.month` would cover the range-boundary hop, which is where the index matching actually does work.

Non-blocking: this block (and the `currentSegmentIndex` / `next` / `prevFocusableSegment` trio) is now duplicated across four roots — the fact that this bug lived in four places is a decent argument for extracting a shared composable in `shared/date/`, but that's a follow-up, not this PR.

61021:
Thanks for the careful review @zernonia — all three points addressed:

- **TimeRangeField covered**: same one-liner in `TimeRangeFieldRoot.vue`, plus a test mirroring the TimeField one (`en-GB` locale so the segment set is stable), which also exercises the `start.minute → end.hour` boundary hop.
- **DateRangeField test extended**: now walks `start.year → end.month` (typing the year as four separate keystrokes — a bracketed `{1980}` arrives as a single key event and the year segment only auto-advances after four digits).
- All four suites green: 164 tests.

On the non-blocking point: agreed — the `currentSegmentIndex` / `nextFocusableSegment` / `prevFocusableSegment` / `focusNext` block is now copy-pasted across four roots and this bug is exactly the kind of thing that breeds. Happy to do the `shared/date/` composable extraction as a follow-up PR once this lands, if you want it.

PR #2815, opened 2026-07-25
fix(PinInput): reflect undefined values in numeric v-model type

### LINKED ISSUE #2811: [Bug]: Numeric pin input should reflect possible `undefined` values in v-model type
### Environment

```bash
n/a
```

n/a

### Steps to reproduce

Enter values into numeric pin input and remove either of them. Observe how a number is replaced with undefined, e.g.:

before: `[1, 2, 3]`
after: `[1, undefined, 3]`

### Describe the bug

Current v-model type does not reflect this scenario (should account for `undefined` too).

### Expected behavior

v-model type accurately reflecting the type of numeric pin input

_No response_

lazerg:
Clearing a numeric PinInput uses `delete` on the model array, which leaves a gap, so a v-model like `[1, 2, 3]` comes back as `[1, undefined, 3]`. The `PinInputValue` type for `'number'` was `number[]`, so those `undefined` entries weren't reflected. This widens the numeric case to `(number | undefined)[]` to match what the component actually emits.

Fixes #2811

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated numeric PIN handling so cleared or empty positions are represented as explicit gaps, keeping other digits intact.
  * Empty or invalid numeric input now results in that slot becoming empty (instead of being removed), improving `modelValue`/completion accuracy and contextual state.
* **Tests**
  * Added coverage for clearing a middle slot via Backspace, verifying the emitted `update:modelValue` includes an explicit empty gap at the cleared index.

### DIFF
--- a/packages/core/src/PinInput/PinInput.test.ts
+++ b/packages/core/src/PinInput/PinInput.test.ts
@@ -334,6 +334,22 @@ describe('give PinInput type=number', async () => {
     })
   })
 
+  describe('after clearing a middle input', () => {
+    beforeEach(async () => {
+      await userEvent.keyboard('12345')
+      inputs[2].element.focus()
+      await inputs[2].trigger('keydown', { key: 'Backspace' })
+    })
+
+    it('should clear only the targeted box', () => {
+      expect(inputs.map(i => i.element.value)).toStrictEqual(['1', '2', '', '4', '5'])
+    })
+
+    it('should emit \'update:modelValue\' with an explicit undefined at the gap', () => {
+      expect(wrapper.emitted('update:modelValue')?.at(-1)?.[0]).toStrictEqual([1, 2, undefined, 4, 5])
+    })
+  })
+
   describe('after user input numeric word consisting only of zeros', () => {
     beforeEach(async () => {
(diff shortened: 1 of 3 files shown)


DamianGlowala:
@lazerg what do you think about changing this line too https://github.com/lazerg/reka-ui/blob/c101b78a2a5371fd8e27c0a2ae57456f604ad186/packages/core/src/PinInput/PinInputInput.vue#L221 so that instead of the delete statement, we would explicitly assign `undefined` to express our intention more clearly?

lazerg:
Agreed, that reads better than the delete. It also matches the `(number | undefined)[]` type nicely, since you get back a proper dense array instead of a sparse one with a hole in it. Nothing else changes internally, since every consumer reads through `[...currentModelValue]` first, so the only real difference is the emitted v-model. Pushed it, along with a test that pins the emitted value to `[1, 2, undefined, 4, 5]` after clearing a middle box.

lazerg:
Hey @zernonia, this one's been sitting a few weeks since the discussion with Damian wrapped up, could you take a look when you have a chance?

PR #2835, opened 2026-08-02
fix(Splitter): validate empty group silently

### LINKED ISSUE #2834: [Bug]: SplitterGroup warns "Invalid layout total size: ." when its last panel is unregistered
### Environment

```bash
Development OS: macOS 26.5.2
Node version: 26.5.0
Package manager: bun@1.3.14
Reka UI version: 2.10.1
Vue version: 3.5.29
Nuxt: not used (plain Vite + Vue SPA)
CSS framework: tailwindcss@4
Client OS: macOS 26.5.2
Browser: Chrome 150
```

https://gist.github.com/rcoenen/0a755251a464f78ecdea2037c53a7bfb

`npm install && npm run dev` — six flat files, no framework beyond Vite + Vue. The whole component is reproduced below too, so there is nothing to click through if you would rather read it here.

### Steps to reproduce

1. Render a `SplitterGroup` whose panels sit behind a `v-if`, so the group can outlive its panels:

```vue
<script setup>
import { SplitterGroup, SplitterPanel, SplitterResizeHandle } from 'reka-ui'
import { ref } from 'vue'

const show = ref(true)
</script>

<template>
  <button @click="show = !show">toggle panels</button>

  <!-- The group outlives its panels: the v-if is inside it, so the group stays
       mounted with zero registered panels. -->
  <SplitterGroup direction="horizontal" style="height: 200px; display: flex">
    <template v-if="show">
      <SplitterPanel :default-size="50">A</SplitterPanel>
      <SplitterResizeHandle style="width: 6px; background: #888" />
      <SplitterPanel :default-size="50">B</SplitterPanel>
    </template>
  </SplitterGroup>
</template>
```

2. Open the console and click **toggle panels** to hide them.

Observed, every time the last panel unregisters:

```
WARNING: Invalid layout total size: . Layout normalization will be applied.
```

Note the empty size list — that is the whole tell. Showing the panels again is silent; hiding them again produces another warning.

### Describe the bug

When a `SplitterGroup` outlives its panels — panels behind a `v-if`, a tab switch, a route change, one rail swapped for another — unregistering the last panel leaves `panelDataArray` empty. The registration watcher still runs and validates that empty group:

`SplitterGroup.vue:320` (the *"(Re)calculate group layout whenever panels are registered or unregistered"* watcher) → `calculateUnsafeDefaultLayout({ panelDataArray: [] })` returns `[]` → `validatePanelGroupLayout({ layout: [], panelConstraints: [] })` at `SplitterGroup.vue:356`.

In `utils/validation.ts` an empty layout then falls through to the total-size check:

```ts
const nextLayoutTotalSize = nextLayout.reduce((acc, cur) => acc + cur, 0)  // 0

if (nextLayout.length !== panelConstraints.length) {      // 0 !== 0 → false, no throw
  ...
}
else if (!fuzzyNumbersEqual(nextLayoutTotalSize, 100)) {  // 0 ≠ 100 → warns
  console.warn(`WARNING: Invalid layout total size: ${...}. Layout normalization will be applied.`)
  ...
}
```

An empty group has no sizes to total, so `0 !== 100` is not a misconfiguration — it is the only value it could have. The normalization the message promises then iterates zero times, so nothing is normalized either.

Stack from the app where we hit it:

```
at validatePanelGroupLayout (reka-ui.js:38219)
at reka-ui.js:38603                       // the panelDataArrayChanged watcher
at callWithErrorHandling (vue.runtime.esm-bundler.js)
```

Two things make this more than a stray log:

- The message names neither the group nor the component, and its size list is empty, so there is nothing in it to search for. With more than one splitter on the page you have to bisect your layout to find out which group is talking — and the answer turns out to be "none of them, in any way you can fix."
- It reads like a real misconfiguration, so it invites you to go hunting for defaults that do not add up to 100. That is exactly what we did, twice, before hooking `console.warn` and noticing the size list was empty.

### Expected behavior

A group with no registered panels validates silently. There is no layout, so there is nothing that can be invalid.

### Context

An early return in `validatePanelGroupLayout` covers both call sites and skips the two normalization loops that would iterate zero times anyway:

```ts
const nextLayout = [...prevLayout]

// An empty group has no sizes to total; 0 !== 100 here is not a misconfiguration.
if (panelConstraints.length === 0)
  return nextLayout

const nextLayoutTotalSize = nextLayout.reduce(...)
```

The watcher at `SplitterGroup.vue:320` could bail on `panelDataArray.length === 0` instead, but the guard in `validation.ts` seems the safer of the two, since `validatePanelGroupLayout` is also reached from the pixel-resize watcher at `SplitterGroup.vue:420`.

Happy to open a PR with the guard and a test if that would help.

lazerg:
When a `SplitterGroup` outlives its panels, say the last one sits behind a `v-if` that gets toggled off, the registration watcher still validates a group that now has zero registered panels. `validatePanelGroupLayout` compares the empty layout's total of 0 against 100 and logs `Invalid layout total size: .`, even though an empty group has no sizes to total and nothing to normalize. The fix returns early when there are no panel constraints, so an empty group validates silently.

Fixes #2834

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of empty panel groups by bypassing unnecessary layout validation and normalization.
  * Prevented irrelevant warnings when panel constraints are unavailable.
  * Added clearer validation for mismatched layout and panel-constraint lengths.

* **Tests**
  * Expanded coverage for empty panel groups and layout validation behavior.
  * Improved test isolation by restoring mocks after each test.

### DIFF
--- a/packages/core/src/Splitter/utils/validation.test.ts
+++ b/packages/core/src/Splitter/utils/validation.test.ts
@@ -1,0 +1,15 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { validatePanelGroupLayout } from './validation'
+
+afterEach(() => {
+  vi.restoreAllMocks()
+})
+
+describe('validatePanelGroupLayout', () => {
+  it('should validate an empty group silently', () => {
+    const warn = vi.spyOn(console, 'warn').mockImplementation(() => {})
+    const result = validatePanelGroupLayout({ layout: [], panelConstraints: [] })
+    expect(result).toEqual([])
+    expect(warn).not.toHaveBeenCalled()
+  })
+})
--- a/packages/core/src/Splitter/utils/validation.test.ts
+++ b/packages/core/src/Splitter/utils/validation.test.ts
@@ -12,4 +12,10 @@ describe('validatePanelGroupLayout', () => {
     expect(result).toEqual([])
     expect(warn).not.toHaveBeenCalled()
   })
+
+  it('should throw when the layout length does not match the panel constraints', () => {
+    expect(() =>
+      validatePanelGroupLayout({ layout: [50, 50], panelConstraints: [] }),
+    ).toThrow('Invalid 0 panel layout')
+  })
 })
--- a/packages/core/src/Splitter/utils/validation.ts
+++ b/packages/core/src/Splitter/utils/validation.ts
@@ -14,7 +14,7 @@ export function validatePanelGroupLayout({
   const nextLayout = [...prevLayout]
 
   // An empty group has no sizes to total, so 0 !== 100 here is not a misconfiguration
-  if (panelConstraints.length === 0)
+  if (nextLayout.length === 0 && panelConstraints.length === 0)
     return nextLayout
 
   const nextLayoutTotalSize = nextLayout.reduce(
--- a/packages/core/src/Splitter/utils/validation.ts
+++ b/packages/core/src/Splitter/utils/validation.ts
@@ -12,6 +12,11 @@ export function validatePanelGroupLayout({
   panelConstraints: PanelConstraints[]
 }): number[] {
   const nextLayout = [...prevLayout]
+
+  // An empty group has no sizes to total, so 0 !== 100 here is not a misconfiguration
+  if (panelConstraints.length === 0)
+    return nextLayout
+
   const nextLayoutTotalSize = nextLayout.reduce(
     (accumulated, current) => accumulated + current,
     0,


zernonia:
Thanks for the fix! Verified locally — it does silence the spurious warning, and the test fails without the change, so it's a proper regression test.

One thing to adjust: the early return sits above the length-mismatch check, so `validatePanelGroupLayout({ layout: [50, 50], panelConstraints: [] })` now silently returns `[50, 50]` instead of throwing `Invalid 0 panel layout`. Could you keep that invariant intact? Smallest change:

```ts
// An empty group has no sizes to total, so 0 !== 100 here is not a misconfiguration
if (nextLayout.length === 0 && panelConstraints.length === 0)
  return nextLayout
```

(Or move the guard below the `throw` and split the `else if` into its own `if`.)

Optional nice-to-have: a test asserting the mismatch still throws, so this doesn't regress later.

lazerg:
Good spot. Tightened the guard to `nextLayout.length === 0 && panelConstraints.length === 0`, so only a genuinely empty group short-circuits and the length-mismatch throw fires again for every other case. Added a test asserting `{ layout: [50, 50], panelConstraints: [] }` still throws `Invalid 0 panel layout`. Pushed to the branch.

PR #2852, opened 2026-08-06
fix(NumberField): allow deleting through formatted literals

### LINKED ISSUE #2850: [Bug]: NumberField backspace gets stuck mid-unit when using style: 'unit' format options
### Environment

```bash
Development/Production OS: macOS 26.3.1 (Darwin 25.3.0, arm64)
Node version: v24.11.1
Package manager: pnpm@11.4.0
Reka UI version: 2.10.1
Vue version: 3.5.27
Nuxt: N/A (not a Nuxt project)
CSS framework: tailwindcss@4.1.18
Client OS: macOS 26.3.1
Browser: Chrome 90.0.4430.212
```

https://stackblitz.com/edit/vitejs-vite-y5tkypjc?file=src%2FApp.vue

### Steps to reproduce

Try backspacing from "13 min", the input will stop at "13 mi"

### Describe the bug

Using NumberField (reka-ui 2.10.1) with:

:format-options="{
style: 'unit',
unit: 'minute',
unitDisplay: 'short',
}"

When the input shows "13 min" and you press Backspace repeatedly, it stops at "13 mi" — deleting the i would produce "13 m", which NumberParser.isValidPartialNumber rejects, so @beforeinput in NumberFieldInput.vue calls preventDefault() and the keystroke is silently swallowed. There's no way to backspace through the unit suffix one character at a time.

Selecting all and deleting does work (the resulting empty string passes validation), so this seems specific to incremental backspacing through the unit literal.

Is there a recommended workaround, or is this considered a bug in the partial-validation logic for unit-formatted values?

### Expected behavior

_No response_

_No response_

61021:
Fixes #2850

With `formatOptions: { style: 'unit', unit: 'minute', unitDisplay: 'short' }`, backspacing `13 min` stops dead at `13 mi`. Any format whose text contains literals (currency codes, unit names) can wedge the same way once a deletion lands inside the literal.

Two layers stack up to cause it:

1. `NumberParser.isValidPartialNumber` strips only complete literals. Its regex is built from full formatted parts (`min` and the space), so `13 mi` leaves `mi` behind and fails; every state between `13 min` and `13 ` is invalid.
2. The `beforeinput` guard's `slice(selectionStart) + data + slice(selectionEnd)` models insertions. A collapsed-cursor backspace removes nothing from that computation, so the guard validates the pre-edit string: it approves `13 min`, the browser deletes to `13 mi`, and the next backspace gets rejected for the state the previous one created.

Which means the deletion gate never prevented invalid intermediate states; it only trapped the caret in them one keystroke late.

The fix exempts deletions (`inputType` starting with `delete`) from the guard. They can only remove characters, and blur/enter still run the remaining text through `parse` exactly as before. Insertions keep the existing validation. I considered mirroring React Aria's per-inputType `nextValue` computation instead (the same slice math appears in their `useFormattedTextField`), but theirs computes the true post-delete string and rejects it, so it blocks the very first backspace rather than the second and makes the suffix fully undeletable. Their NumberField wedges the same way; I didn't find it on their tracker.

Tests: red/green on backspacing `13 min` down to empty (v2 wedges at `13 mi`), plus a test pinning that invalid insertions still get rejected — the `beforeinput` guard had no coverage before these. Full core suite 2006/2006, vue-tsc and eslint clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved deletion behavior in unit-formatted number fields.
  - Backspacing through a unit suffix now correctly clears the associated value.
  - Invalid characters remain blocked while the unit suffix stays visible.
  - Partial values are supported during editing and validated when the field loses focus or is submitted.

### DIFF
--- a/packages/core/src/NumberField/NumberField.test.ts
+++ b/packages/core/src/NumberField/NumberField.test.ts
@@ -237,6 +237,43 @@ describe('numberField', () => {
       expect(input.value).toBe('5 inches')
     })
 
+    it('should allow backspacing through the unit suffix', async () => {
+      const { input, user } = setup({
+        defaultValue: 13,
+        formatOptions: {
+          style: 'unit',
+          unit: 'minute',
+          unitDisplay: 'short',
+        },
+      })
+      expect(input.value).toBe('13 min')
+
+      await user.click(input)
+      input.setSelectionRange(input.value.length, input.value.length)
+      const initialLength = input.value.length
+      for (let i = 0; i < initialLength; i++)
+        await user.keyboard('{Backspace}')
+
+      expect(input.value).toBe('')
+    })
+
+    it('should still reject invalid insertions while showing units', async () => {
+      const { input, user } = setup({
+        defaultValue: 13,
+        formatOptions: {
+          style: 'unit',
+          unit: 'minute',
+          unitDisplay: 'short',
+        },
+      })
+
+      await user.click(input)
+      input.setSelectionRange(2, 2)
+      await user.keyboard('x')
+
+      expect(input.value).toBe('13 min')
+    })
+
     it('should change format based on reactive options', async () => {
       const { input, rerender } = setup({
         defaultValue: 5,
--- a/packages/core/src/NumberField/NumberFieldInput.vue
+++ b/packages/core/src/NumberField/NumberFieldInput.vue
@@ -122,9 +122,10 @@ function handleChange() {
     @beforeinput="(event: InputEvent) => {
       if (event.isComposing) return
       // Deletions are exempt: partially deleted literals ('13 mi' while
-      // backspacing through '13 min') never pass partial validation, and
-      // blur/enter re-parse whatever remains.
-      if (event.inputType.startsWith('delete')) return
+      // backspacing through '13 min') never pass partial validation, so
+      // validating them would trap the caret mid-literal. Deletions only
+      // ever remove characters, so they cannot introduce invalid input.
+      if (event.inputType?.startsWith('delete')) return
       const target = event.target as HTMLInputElement
       let nextValue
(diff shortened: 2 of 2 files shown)


MILLERMARRU (commented):
Checked the InputEvent spec on MDN for `inputType` values, deletions there are always one of `deleteContentBackward`, `deleteContentForward`, `deleteWordBackward`, `deleteByCut`, and a handful more, all prefixed with `delete` per the Input Events Level 2 spec, so `startsWith('delete')` catches the full set rather than just the common backspace case, which matters since this also needs to work for forward-delete and select-then-cut.

The reasoning for exempting deletions entirely holds up once you look at what happens after, `@blur` still calls `rootContext.applyInputValue(...)`, which is untouched by this PR and is what actually re-parses and commits the final value. So letting a deletion through unvalidated during `beforeinput` doesn't skip validation altogether, it just defers it to the point where the user is done typing rather than trying to validate every single intermediate keystroke of a multi-character deletion, which is exactly the case that was broken, `"13 mi"` mid-backspace through `"13 min"` was never going to pass partial validation no matter what, since it's not a complete number-plus-unit shape and never will be until the whole unit is gone.

The second test is the one that actually proves this isn't just loosening validation across the board, typing `x` (a real insertion, not a delete) at a mid-string position still gets rejected and the value stays `"13 min"`, so the fix is scoped to the direction of the character count that genuinely can't produce new invalid content.

61021:
Rebased on current v2. #2851 already landed the deletion fix, so what's left here is the `inputType?.` guard from your commit and two tests: one for a plain `beforeinput` event, which fails without the guard, and one for an invalid insertion with a unit suffix. CI passes. Should I keep it at that, or close it?

61021:
Closing, #2851 already covers this.

PR #2854, opened 2026-08-07
fix(Drawer): track drags in non-dismissable directions so the drawer pulls elastically

zernonia:
## Problem

The Drawer has no elastic "pull" feedback. Dragging a bottom drawer **upward** does nothing at all — no transform, no `data-swiping` — where [Base UI's Drawer](https://base-ui.com/react/components/drawer) rubber-bands.

## Root cause

`applyDirectionalDamping()` sqrt-damps any axis moving in a direction that can't dismiss the drawer. That's exactly the elastic pull, and it was a faithful port — but it was **unreachable**. `processMove` bailed out first:

```js
if (!dir)   // no allowed dismiss direction in this drag
  return
```

Base UI starts tracking the gesture on pointerdown regardless of direction (`startSwipeAtPosition` calls `setSwiping(true)` unconditionally) and only *later* adopts an intended direction, once the drag first moves in a dismissable one. This port folded direction-adoption into the promotion step, so a pure pull-up never became an active swipe.

## Fix

1. Promote to an active swipe on any movement past `MIN_DRAG_THRESHOLD`, not just movement in an allowed direction.
2. Adopt `intendedDirection` lazily on each move until one is found — otherwise a pull-up-then-push-down gesture would be permanently stuck without a dismiss candidate.
3. Guard the reversal-cancel branch on having an intended direction, and fall back to `directions[0]` for progress (a pull away clamps to 0), matching Base UI's `progressDirection = primaryDirection ?? intendedSwipeDirection`.

## Verification

Drove both implementations in Chromium via Playwright — 150px upward drag on a bottom drawer, reading the computed transform:

| | Base UI (base-ui.com) | before | after |
|---|---|---|---|
| `transform` | `translateY(-11.83px)` | `translateY(0)` | `translateY(-12.25px)` |
| `--drawer-swipe-movement-y` | `-11.83px` | `0px` | `-12.25px` |
| `data-swiping` | present | absent | present |

−12.25 = −√150, the expected sqrt damping. Base UI's −11.83 = −√140 — it absorbs the first move into the drag origin, hence the 10px difference.

Also browser-verified, no regressions:
- pull-up-then-push-down past threshold still dismisses
- sub-threshold down-drag still cancels and springs back
- Snap Points variant unchanged (both vertical directions are allowed there, so nothing is damped)
- Top / Right drawers now rubber-band on their non-dismiss axis too (`-12px` for a 144px drag)

Left alone deliberately: on touch, if the drawer body is scrollable and not at the relevant edge, an upward drag still yields to scrolling instead of rubber-banding. That matches Base UI, which refuses to start the swipe on a scrollable touch target.

## Checks

- 3 regression tests added to `useSwipeDismiss.test.ts`
- Full suite: 2007 tests pass
- `pnpm lint` and `pnpm --filter reka-ui type-check` clean

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved drawer swipe-to-dismiss gestures when dragging opposite to the configured dismiss direction.
  * Added elastic resistance and smoother progress tracking for opposing movement.
  * Swipes can now reverse into the dismiss direction without losing gesture state.
  * Improved recovery when pointer releases, capture is lost, or the window blurs.
  * Gestures are correctly canceled and reset between interrupted or incomplete swipes.
  * Improved drawer stretching during overscroll to keep content connected to its anchored edge.

* **Documentation**
  * Added guidance for configuring overscroll bleed and safe-area spacing.

### DIFF
--- a/docs/components/demo/Drawer/css/styles.css
+++ b/docs/components/demo/Drawer/css/styles.css
@@ -24,12 +24,19 @@ input {
 }
 
 .DrawerContent {
+  /* Extra drawer area that hangs below the viewport edge, kept off-screen by an
+     equal negative margin. Pulling the drawer away from its edge slides the
+     bleed into view instead of exposing the overlay behind it, so the drawer
+     stretches rather than visually detaching while it rubber-bands. */
+  --bleed: 48px;
   background-color: white;
   border-radius: 16px 16px 0 0;
   position: fixed;
   inset-inline: 0;
   bottom: 0;
   margin-inline: auto;
+  margin-bottom: calc(-1 * var(--bleed));
+  padding-bottom: calc(env(safe-area-inset-bottom, 0px) + var(--bleed));
   max-width: 500px;
   display: flex;
   flex-direction: column;
@@ -159,13 +166,16 @@ input {
     opacity: 0;
   }
 }
+/* The bleed already sits below the viewport edge, so the drawer only has to
+   travel `height - bleed` to clear it. Translating a full 100% would overshoot
+   by the bleed and make the slide look faster than it is. */
 @keyframes drawerSlideBottomIn {
   from {
-    translate: 0 100%;
+    translate: 0 calc(100% - var(--bleed));
   }
 }
 @keyframes drawerSlideBottomOut {
   to {
-    translate: 0 100%;
+    translate: 0 calc(100% - var(--bleed));
   }
 }
--- a/docs/components/demo/Drawer/css/styles.css
+++ b/docs/components/demo/Drawer/css/styles.css
@@ -29,6 +29,7 @@ input {
      bleed into view instead of exposing the overlay behind it, so the drawer
      stretches rather than visually detaching while it rubber-bands. */
   --bleed: 48px;
+
   background-color: white;
   border-radius: 16px 16px 0 0;
   position: fixed;
--- a/docs/components/demo/Drawer/tailwind/index.vue
+++ b/docs/components/demo/Drawer/tailwind/index.vue
@@ -21,8 +21,12 @@ import {
     </DrawerTrigger>
     <DrawerPortal>
       <DrawerOverlay class="DrawerOverlay fixed inset-0 z-30 bg-black/40" />
+      <!-- `--bleed` is extra drawer area hanging below the viewport edge, kept
+           off-screen by an equal negative margin. Pulling the drawer away from
+           its edge slides the bleed into view instead of exposing the overlay,
+           so the drawer stretches rather than detaching while it rubber-bands. -->
       <DrawerContent
-        class="DrawerContent fixed inset-x-0 bottom-0 z-[100] mx-auto flex max-w-[500px] flex-col rounded-t-[16px] bg-white outline-none"
+        class="DrawerContent fixed inset-x-0 bottom-0 z-[100] mx-auto flex max-w-[500px] flex-col rounded-t-[16px] bg-white outline-none [--bleed:48px] mb-[calc(-1*var(--bleed))] pb-[calc(env(safe-area-inset-bottom,0px)+var(--bleed))]"
       >
         <DrawerHandle class="mx-auto mt-3 h-1.5 w-12 shrink-0 rounded-full bg-mauve6" />
         <div class="p-6">
--- a/docs/components/demo/Drawer/tailwind/index.vue
+++ b/docs/components/demo/Drawer/tailwind/index.vue
@@ -99,6 +99,9 @@ import {
 
 @keyframes drawer-overlay-in { from { opacity: 0; } }
 @keyframes drawer-overlay-out { to { opacity: 0; } }
-@keyframes drawer-slide-bottom-in { from { translate: 0 100%; } }
-@keyframes drawer-slide-bottom-out { to { translate: 0 100%; } }
+/* The bleed already sits below the viewport edge, so the drawer only has to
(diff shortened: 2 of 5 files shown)


zernonia:
Follow-up: added the `--bleed` pattern to the docs demo (`d3eb9d8`).

**On "pull up seems stuck":** the motion itself is already correct — I compared the damping curve against base-ui.com across a 400px drag and they match:

| finger travel | Base UI | reka (this branch) |
|---|---|---|
| 100px | -8.94px | -10.00px |
| 200px | -13.42px | -14.14px |
| 300px | -16.73px | -17.32px |
| 400px | -19.49px | -20.00px |

(Both are −√travel. Base UI's are one step behind because it absorbs the first move into the drag origin.) Dragging from `DrawerHandle` measured identical to dragging from the body, and a down-drag from the handle still dismisses.

What actually looked broken was the demo CSS. Pulling up lifted the drawer clean off the bottom edge and the overlay showed through the gap:

**Before** — 13px gap of overlay under the drawer
**After** — drawer stays flush, matching Base UI's demo

Base UI hides this with `--bleed`: extra popup area past the anchored edge, held off-screen by an equal negative margin, so pulling slides the bleed into view instead of the overlay.

```css
.DrawerContent {
  --bleed: 48px;
  padding-bottom: calc(env(safe-area-inset-bottom, 0px) + var(--bleed));
  margin-bottom: calc(-1 * var(--bleed));
}

@keyframes slideIn { from { translate: 0 calc(100% - var(--bleed)); } }
@keyframes slideOut { to { translate: 0 calc(100% - var(--bleed)); } }
```

The keyframes travel `100% - bleed` rather than a full `100%` — the bleed already sits below the edge, so a full 100% overshoots and makes the slide look faster than it is. Verified the enter/exit still land exactly flush with the viewport edge at both extremes.

Applied to both demo variants (CSS + Tailwind) and documented as a "Bleeding past the edge" subsection so the copy-pasteable snippet matches what the demo does.

zernonia:
Reproduced and fixed in `195bb40` — and it was worse than stuck at position: the gesture wedged **permanently**.

### What happens

The listeners live on the popup. `setPointerCapture` keeps an ordinary in-window drag working (I verified tracking continues and release-outside settles correctly), but nothing covered a release the popup never receives — outside the window, over another application, or after the browser drops capture. Simulating that:

```
during drag        : { sy: '-7.74597px', swiping: true }
after phantom release: { sy: '-7.74597px', swiping: true }   <- stuck
subsequent real drag : { sy: '-7.74597px', swiping: true }   <- ignored entirely
```

`data-swiping` stays set, which pins `transition-duration: 0ms`, so the drawer can never animate back — and because `isSwiping`/`cancelledSwipe` never clear, every later drag is dead too.

### Root cause

Base UI hangs the gesture off `Drawer.Viewport`, which is `position: fixed; inset: 0` — a full-screen element, so the pointer can never leave the thing that's listening. This port attaches to `DrawerContent` (the popup), which the pointer leaves the moment the drag passes the drawer's bounds. `DrawerViewport` here is a documented passthrough.

### Fix

Rather than restructure where the gesture lives, make the drag recoverable from any release:

- mirror `pointerup`/`pointercancel` on the owner document
- end the drag on `lostpointercapture` and on window `blur` (the off-window release case)
- treat a `pointermove` with no button held as the missing `pointerup` — the only signal the browser gives when the release happened off-window
- reset per-gesture state on every press (Base UI parity with `startSwipeAtPosition`)

That last one matters independently: a gesture whose end was missed leaked its latched `cancelledSwipe` into the next one, leaving the drawer silently refusing to close however far you dragged it.

After the fix, the same simulation recovers to `0px` with `swiping: false`, and the next drag dismisses normally. Normal drags, pull-up rubber-banding, reversal, and snap points all re-verified unchanged.

4 regression tests added — all four fail without the fix. Full suite 2011 passing, lint and type-check clean.

### Worth flagging

`DrawerViewport` being a passthrough is the underlying divergence here. Moving the gesture onto a full-screen viewport would match Base UI's architecture and remove this whole class of problem, but it's an API-shape change (consumers would have to render `DrawerViewport` and style it `fixed; inset: 0`) so I kept it out of this PR.

zernonia:
Addressed CodeRabbit's three comments in `e474de4`.

**1. Tailwind demo keyframes still at `100%` — valid, and it corrects something I said earlier.**

I claimed above that the Tailwind variant has no `.DrawerContent` rules of its own and inherits them from the CSS demo's global `styles.css`. That was wrong — I'd only read the first 60 lines of the file. It has its own unscoped `<style>` block with its own `drawer-slide-bottom-{in,out}` keyframes, which were still translating a full `100%`.

Measured before and after, viewport height 700:

| | enter starts at | exit ends at |
|---|---|---|
| before | `top: 721` (21px past flush) | `top: 747` |
| after | `top: 689` | `top: 700` — exactly flush |

So it really was overshooting the closed position by the bleed. Fixed to `calc(100% - var(--bleed))` and re-verified the pull-up still holds the drawer against the edge.

**2. Left/right bleed guidance — valid.** My one-liner was only correct for a right-anchored drawer; a left-anchored one bleeds the other way. Split into two explicit variants with `padding/margin-inline-start` and negative X translation.

**3. Stylelint blank line — applied.** Worth noting this repo has no stylelint config and doesn't run it (lint is ESLint, which passes either way), so this is a CodeRabbit-side default rather than a build failure. Applied anyway since it's cosmetic and zero-risk.

PR #2863, opened 2026-08-10
fix(DismissableLayer): ignore outside pointerdown while not present

benjamincanac:
### 🔗 Linked issue

Reported downstream in nuxt/ui#6797.

### ❓ Type of change

### 📚 Description

On touch, `usePointerDownOutside` doesn't dispatch on `pointerdown`, it defers to the next `click` so the ~350ms tap delay can't replay events that should have been prevented.

A layer that stays mounted while hidden (`unmountOnHide: false`) was still listening while `present` was `false`, so it captured the `pointerdown` of the tap that opens it. By the time that tap's `click` fired the layer was present, the deferred handler ran, and it dismissed itself on the very interaction that opened it. On mouse there's no deferral, so the event is dispatched while the layer is still not present and the `!props.present` guard in the handler catches it. Touch never hits that guard.

Passing `present` as the composable's `enabled` means the listener isn't registered at all while hidden, and the existing `setTimeout(0)` before attaching covers the tick where the layer becomes present.

`useFocusOutside` takes the same `enabled` argument but is deliberately left alone here: it also gates `onBlurCapture`, so `isFocusInsideDOMTree` would stay stuck at `true` when focus leaves a hidden layer, and a later outside focus would be read as inside.

### 📝 Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue where showing a dismissible layer after a touch interaction could incorrectly trigger outside dismissal.
  - Touch interactions outside the layer now correctly emit outside-pointer and dismissal events.
  - Hidden but mounted layers no longer respond to deferred touch events.

### DIFF
--- a/packages/core/src/DismissableLayer/DismissableLayer.test.ts
+++ b/packages/core/src/DismissableLayer/DismissableLayer.test.ts
@@ -441,8 +441,8 @@ describe('given a not-present DismissableLayer (e.g. unmountOnHide hidden)', () 
     expect(wrapper.emitted('dismiss')).toBeUndefined()
   })
 
-  // Regression (nuxt/ui#6797): on touch, `pointerDownOutside` is deferred to the `click`
-  // event. A layer listening while not present captures the `pointerdown` of the tap that
+  // Regression: on touch, `pointerDownOutside` is deferred to the `click` event.
+  // A layer listening while not present captures the `pointerdown` of the tap that
   // opens it, and dismisses itself when that tap's `click` arrives.
   it('should not dismiss on the tap that made it present', async () => {
     // jsdom has no `PointerEvent`, so `fireEvent.pointerDown` loses `pointerType`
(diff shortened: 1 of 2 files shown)


MILLERMARRU (commented):
Checked the actual mechanism in `utils.ts` rather than taking the deferred-touch claim on faith, and it matches exactly: `pointerType === 'touch'` skips `handleAndDispatchPointerDownOutsideEvent()` and instead registers a one-time `click` listener that fires it later, while the mouse branch calls it immediately. That confirms why the existing `!props.present` guard inside the handler catches the mouse case (event dispatched while still hidden) but not touch (event dispatched later, after `present` flipped).

Also checked the reasoning for deliberately not touching `useFocusOutside` the same way, since that's the part that would be easy to get wrong. `onBlurCapture` there is gated by the same `enabled` param and sets `isFocusInsideDOMTree.value = false`. If it were gated by `present` too, a blur that happens while the layer is hidden would be silently skipped, so `isFocusInsideDOMTree` would stay stuck at whatever it was before hiding. When the layer becomes present again and focus genuinely moves outside afterward, `handleFocus`'s `!isFocusInsideDOMTree.value` check would still read the stale `true` and never dispatch. So leaving that composable alone is the correct call given the shape of its own state tracking, not an inconsistency.

Good regression test, the touch-specific `pointerType` property injection on a plain `MouseEvent` for jsdom is a reasonable way to hit the touch branch without a real `PointerEvent`.

zernonia:
LGTM @benjamincanac ! Nicely done 😁

Issue #2682, opened 2026-06-08
[Bug]: ConfigProvider useId is ignored on Vue 3.5+, causing SSR hydration ID mismatches in Nuxt prerender


Mat4m0:
### Environment

```bash
- Reka UI: 2.9.9
- Vue: 3.5.35
- Nuxt: 4.4.6
- Package manager: pnpm 10.27.0
- Rendering mode: Nuxt prerender / static output
- Browser: Chrome
```

### Link to minimal reproduction

https://github.com/Mat4m0/hydration

### Steps to reproduce

1. Install dependencies:

```bash
pnpm install
```

2. Generate the static site:
```bash
pnpm generate
```

3. Preview the generated static site:

```bash
npx serve .output/public  
```

4. Open the local URL printed by Nuxt, usually:

```txt
http://localhost:3000/
```

5. Check the browser console.

### Describe the bug

Vue reports hydration attribute mismatches for Reka-generated IDs.

Examples:

```txt
[Vue warn]: Hydration attribute mismatch

- rendered on server: id="reka-accordion-trigger-v-1-0"
- expected on client: id="reka-accordion-trigger-v-0-0"
```

```txt
[Vue warn]: Hydration attribute mismatch

- rendered on server: id="reka-tabs-v-2-0-trigger-account"
- expected on client: id="reka-tabs-v-1-0-trigger-account"
```

This also produces:

```txt
Hydration completed but contains mismatches.
```

### Root cause

Reka's shared `useId` helper currently prefers Vue's `useId` when available.

On Vue 3.5+, this means `ConfigProvider.useId` is effectively bypassed. That makes the provider unable to supply a framework-specific SSR-stable ID source.

The current behavior appears to be:

```ts
Vue useId -> ConfigProvider.useId -> fallback counter
```

But the provider should take precedence:

```ts
explicit deterministic ID -> ConfigProvider.useId -> Vue useId -> fallback counter
```

### Why this matters

`ConfigProvider.useId` is the public escape hatch for frameworks/integrations that need to provide their own stable ID generation strategy.

If Reka ignores the provider whenever Vue `useId` exists, the provider is no longer authoritative on Vue 3.5+.

This affects multiple primitives because they all use the shared Reka `useId` helper. In the repro, both Accordion and Tabs show the issue:

- Accordion: independent trigger/content IDs
- Tabs: base ID with derived trigger/content IDs

### Additional notes

[...]

Issue #2601, opened 2026-04-17
TabsTrigger sets aria-controls when no TabsContent exists


helloneele:
## Description

`TabsTrigger` unconditionally sets `aria-controls` to a generated content panel ID (`reka-tabs-...-content-{value}`), even when no corresponding `TabsContent` is rendered. This causes an `aria-valid-attr-value` axe-core violation because `aria-controls` references a DOM ID that doesn't exist.

This is a common pattern when using Tabs for navigation (route switching) rather than content toggling — the tabs control which component/route is shown, not which content panel is visible.

## ARIA Spec Reference

Per [WAI-ARIA 1.2 §6.6.2](https://www.w3.org/TR/wai-aria-1.2/#aria-controls):

> If an element with `aria-controls` references an ID that does not exist in the DOM, the attribute is invalid.

`aria-controls` is optional on `role="tab"` — omitting it is valid and preferred over referencing a non-existent element.

## Reproduction

```vue
<TabsRoot default-value="tab-1">
  <TabsList>
    <TabsTrigger value="tab-1">First</TabsTrigger>
    <TabsTrigger value="tab-2">Second</TabsTrigger>
  </TabsList>
  <!-- No TabsContent — tabs used for navigation -->
</TabsRoot>
```

Inspecting the rendered trigger:
```html
<button role="tab" aria-controls="reka-tabs-v-...-content-tab-1" ...>
```

The ID `reka-tabs-v-...-content-tab-1` does not exist in the DOM → axe-core reports `aria-valid-attr-value` (critical).

## Expected Behavior

`aria-controls` should only be set when the corresponding `TabsContent` element is actually registered/mounted. When no content panel exists, the attribute should be omitted.

## Suggested Fix

The `TabsRoot` context already tracks registered content panels. `TabsTrigger` could check whether a content panel with the matching value is registered before setting `aria-controls`:

```ts
// In TabsTrigger
const contentId = computed(() => {
  const id = makeContentId(rootContext.baseId, props.value)
  // Only set aria-controls if the content panel exists
  return rootContext.contentPanels?.has(props.value) ? id : undefined
})
```

Alternatively, if content panel tracking isn't available in the context, a DOM check on mount would work.

## Environment

- reka-ui version: latest
- Detected via axe-core 4.10.3 accessibility testing

helloneele:
[...]

PR #2866, opened 2026-08-10
fix(tabs): keep aria-controls resolvable when SSR and client ids diverge

TerrorSquad:
## Problem

`TabsTrigger` rebuilds the content id from `rootContext.baseId` instead of using the id the panel actually rendered with:

```ts
const contentId = computed(() => rootContext.contentIds.value.has(props.value)
  ? makeContentId(rootContext.baseId, props.value)
  : undefined)
```

`baseId` comes from `useId()`, and that counter can land on a different value during hydration than it did during SSR. The panel keeps its server id, the trigger recomputes `aria-controls` from the client one, and it points at nothing:

```
trigger id:    reka-tabs-v-0-0-16-0-trigger-queue
aria-controls: reka-tabs-v-0-0-2-0-content-queue   <- no such element
panel id:      reka-tabs-v-0-0-16-0-content-queue
```

One `TabsRoot` per side, so this is not a double mount. axe reports it as a critical `aria-valid-attr-value` violation.

It only shows up after hydration in a production build, never in `nuxt dev`, because `aria-controls` is unset until `TabsContent` registers in `onMounted`. Related: #2682 shows the same id drift.

## Reproduction

Any page where the two passes consume a different number of `useId()` calls before the tabs render. Forcing it directly:

```vue
<script setup lang="ts">
if (import.meta.server) {
  useId(); useId(); useId(); useId(); useId()
}
</script>
```

Build, serve, and compare each trigger's `aria-controls` against the panel ids. Before: 2 of 2 dangle. After: 0.

## Fix

`TabsContent` registers the id it rendered with, and `TabsTrigger` reads that back. Registration runs from a `watch` so it stays correct if the value changes. Triggers with no matching content still get no `aria-controls` (#2601 behaviour unchanged).

## Verified

2014 tests pass, `type-check` and eslint clean. Two regression tests, both fail on `v2`. Checked against the app this came from: axe on the real page went from 2 failures to 0.

## Breaking

`contentIds` becomes a `Map<StringOrNumber, string>` and `registerContent` takes the id as a second argument. Both are exported. Happy to keep the `Set` alongside if you would rather not break it.

### DIFF
--- a/packages/core/src/Tabs/Tabs.test.ts
+++ b/packages/core/src/Tabs/Tabs.test.ts
@@ -139,4 +139,46 @@ describe('given Tabs without TabsContent', () => {
     expect(triggers[0].attributes('aria-controls')).toBeDefined()
     expect(triggers[1].attributes('aria-controls')).toBeUndefined()
   })
+
+  it('should render aria-controls matching the TabsContent id', async () => {
+    const wrapper = mount({
+      components: { TabsRoot, TabsList, TabsTrigger, TabsContent },
+      template: `
+        <TabsRoot default-value="tab1">
+          <TabsList>
+            <TabsTrigger value="tab1">Tab 1</TabsTrigger>
+          </TabsList>
+          <TabsContent id="rendered-id" value="tab1">Content 1</TabsContent>
+        </TabsRoot>
+      `,
+    })
+    await flushPromises()
+
+    const trigger = wrapper.find('[role="tab"]')
+    const content = wrapper.find('[role="tabpanel"]')
+    expect(trigger.attributes('aria-controls')).toBe(content.attributes('id'))
+  })
+
+  it('should update aria-controls when the TabsContent value changes', async () => {
+    const wrapper = mount({
+      components: { TabsRoot, TabsList, TabsTrigger, TabsContent },
+      props: { value: String },
+      template: `
+        <TabsRoot :default-value="value">
+          <TabsList>
+            <TabsTrigger :value="value">Tab 1</TabsTrigger>
+          </TabsList>
+          <TabsContent :value="value">Content 1</TabsContent>
+        </TabsRoot>
+      `,
+    }, { props: { value: 'tab1' } })
+    await flushPromises()
+
+    await wrapper.setProps({ value: 'tab2' })
+    await flushPromises()
+
+    const trigger = wrapper.find('[role="tab"]')
+    const content = wrapper.find('[role="tabpanel"]')
+    expect(trigger.attributes('aria-controls')).toBe(content.attributes('id'))
+  })
 })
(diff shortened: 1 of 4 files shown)


TerrorSquad:
Thanks, the registration point was a real gap.

**Stale `contentIds`**: fixed. Registration now runs from a `watch` on `props.value` and the rendered id, and unregisters the old value when it changes. Added a test for a tab whose value changes while mounted; it fails on `v2`.

`watchEffect` deadlocks here, in case it comes up: it reads `currentElement`, writes `contentIds`, that re-renders the trigger, which re-triggers the effect. Explicit `watch` sources avoid it.

Not covered: changing the `id` attribute while the content stays mounted. `currentElement` only re-evaluates when the element identity changes, so that would mean watching DOM attributes.

**`@testing-library/vue`**: leaving as is. `Tabs.test.ts` is all `@vue/test-utils`, and repo-wide it is 66 files to 36, so one test in the other library would be the odd one out. Happy to convert the file separately if you want it standardised.

2014 tests pass, `type-check` and eslint clean.

zernonia:
Thanks for this, the fixed-id case works well. There's one gap: if the `id` on `TabsContent` changes after mount, `aria-controls` keeps the old value.

The watcher reads `currentElement.value?.id`, and Vue can't track a DOM attribute. So the watcher only re-runs when `props.value` or the element itself changes.

**Repro**

```ts
it('keeps aria-controls in sync when the id attr changes', async () => {
  const wrapper = mount({
    components: { TabsRoot, TabsList, TabsTrigger, TabsContent },
    props: { panelId: String },
    template: `
      <TabsRoot default-value="tab1">
        <TabsList>
          <TabsTrigger value="tab1">Tab 1</TabsTrigger>
        </TabsList>
        <TabsContent :id="panelId" value="tab1">Content 1</TabsContent>
      </TabsRoot>
    `,
  }, { props: { panelId: 'x' } })
  await flushPromises()

  await wrapper.setProps({ panelId: 'y' })
  await flushPromises()

  const trigger = wrapper.find('[role="tab"]')
  const content = wrapper.find('[role="tabpanel"]')
  expect(trigger.attributes('aria-controls')).toBe(content.attributes('id'))
  // AssertionError: expected 'x' to be 'y'
})
```

**Possible fix**

A changed `id` attr re-renders `TabsContent`, so refreshing the stored id in `onUpdated` catches it. Watching `attrs.id` wouldn't help, because `attrs` isn't reactive.

```ts
onUpdated(() => {
  const id = currentElement.value?.id
  if (id && rootContext.contentIds.value.get(props.value) !== id)
    rootContext.registerContent(props.value, id)
})
```

With this change and the test above, all 10 Tabs tests pass locally. Could you add both to the PR?

PR #2869, opened 2026-08-12
fix(Tooltip): receive the open broadcast on document instead of window

### LINKED ISSUE #2868: [Bug]: Tooltip's one-at-a-time close never fires (TOOLTIP_OPEN dispatched on document, listened on window, non-bubbling)
### Environment

```bash
reka-ui version: 2.10.1 (verified still present on main)
Vue version: 3.5.39
Browser: Chrome 140 (also reproduced under happy-dom)
```

### Describe the bug

`Tooltip`'s "close the others when one opens" mechanism never runs. The event is dispatched on
`document` and listened for on `window`, and it is not a bubbling event, so the listener is never
reached.

`TooltipRoot.vue:121` (dispatch):

```ts
document.dispatchEvent(new CustomEvent(TOOLTIP_OPEN))
```

`TooltipContentImpl.vue:90` (listen):

```ts
useEventListener(window, TOOLTIP_OPEN, rootContext.onClose)
```

`new CustomEvent(TOOLTIP_OPEN)` defaults to `bubbles: false`. For an event dispatched on `document`,
the propagation path is `[window, document]`, so `window` is only visited during the CAPTURE phase.
`useEventListener` registers in the bubble phase by default, and the bubble phase is skipped for a
non-bubbling event. The handler therefore never fires, for any tooltip, ever.

This is a different root cause from #1544 and #850, which were both about the grace area. Those
required a specific pointer path to reproduce; this one means the broadcast has simply never worked.

In practice it goes unnoticed with a mouse, because the pointer leaving trigger A closes A anyway
before B opens. It becomes visible as soon as anything opens a tooltip without a preceding
`pointerleave` on the other one, for example a touch or programmatic open, which is how I ran into
it.

### Steps to reproduce

Two tooltips, no pointer path between them:

```vue
<TooltipProvider :delay-duration="0">
  <TooltipRoot>
    <TooltipTrigger as-child><button data-testid="a">A</button></TooltipTrigger>
    <TooltipPortal><TooltipContent>tip A</TooltipContent></TooltipPortal>
  </TooltipRoot>
  <TooltipRoot>
    <TooltipTrigger as-child><button data-testid="b">B</button></TooltipTrigger>
    <TooltipPortal><TooltipContent>tip B</TooltipContent></TooltipPortal>
  </TooltipRoot>
</TooltipProvider>
```

```ts
const a = document.querySelector('[data-testid="a"]')!
const b = document.querySelector('[data-testid="b"]')!
a.dispatchEvent(new PointerEvent('pointermove', { pointerType: 'mouse', bubbles: true }))
// wait a tick for the delay timer
b.dispatchEvent(new PointerEvent('pointermove', { pointerType: 'mouse', bubbles: true }))

a.dataset.state // -> "delayed-open"   (expected "closed")
b.dataset.state // -> "instant-open"
```

Both tooltip contents stay mounted.

The propagation itself, independent of reka, in any console:

```js
let bubble = 0, capture = 0
window.addEventListener('tooltip.open', () => bubble++)
window.addEventListener('tooltip.open', () => capture++, true)
document.dispatchEvent(new CustomEvent('tooltip.open'))
console.log({ bubble, capture })   // -> { bubble: 0, capture: 1 }
```

I do not have a hosted reproduction, sorry. The console snippet above is self contained and the
component snippet reproduces it in any app that renders two tooltips.

### Expected behavior

Opening one tooltip closes any other that is open, which is what `TOOLTIP_OPEN` is there to do.

### Possible fix

Either side works:

```ts
// TooltipRoot.vue
document.dispatchEvent(new CustomEvent(TOOLTIP_OPEN, { bubbles: true }))
```

```ts
// TooltipContentImpl.vue
useEventListener(document, TOOLTIP_OPEN, rootContext.onClose)
```

Listening on `document` is the smaller change and keeps the event non-bubbling, so it cannot be
observed by unrelated listeners on `window`. Happy to open a PR if you would like one.

lazerg:
### 🔗 Linked issue

Fixes #2868

### ❓ Type of change

### 📚 Description

The "close the others when one opens" broadcast never actually ran. `TooltipRoot` dispatches `TOOLTIP_OPEN` on `document`, but `TooltipContentImpl` listens for it on `window`. The event isn't set to bubble, and for something dispatched on `document`, `window` only sees it during the capture phase, not the bubble phase the listener is registered for. So the handler was skipped every time.

This stayed hidden with a mouse because leaving one trigger already closes it before the pointer reaches the next one. It shows up as soon as a tooltip opens without a pointerleave in between, for example with touch input or a programmatic open.

Moved the listener to `document`, matching where the event is dispatched from.

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tooltip behavior so opening one tooltip correctly closes any tooltip that is already open.
  * Enhanced tooltip event handling for more reliable behavior across the application.

* **Tests**
  * Added coverage for tooltip replacement behavior and improved test cleanup.

### DIFF
--- a/packages/core/src/Tooltip/Tooltip.test.ts
+++ b/packages/core/src/Tooltip/Tooltip.test.ts
@@ -1,9 +1,10 @@
 import type { VueWrapper } from '@vue/test-utils'
 import { flushPromises, mount } from '@vue/test-utils'
-import { beforeEach, describe, expect, it } from 'vitest'
+import { afterEach, beforeEach, describe, expect, it } from 'vitest'
 import { axe } from 'vitest-axe'
 import Tooltip from './stories/_Tooltip.vue'
 import TooltipProvider from './TooltipProvider.vue'
+import { TOOLTIP_OPEN } from './utils'
 
 describe('given default Tooltip', () => {
   let wrapper: VueWrapper<InstanceType<typeof Tooltip>>
@@ -18,6 +19,11 @@ describe('given default Tooltip', () => {
     wrapper = mount(Tooltip, { attachTo: document.body })
   },
   )
+
+  afterEach(async () => {
+    wrapper.unmount()
+    await flushPromises()
+  })
 
   it('should pass axe accessibility tests', async () => {
     expect(await axe(wrapper.element)).toHaveNoViolations()
@@ -50,6 +56,16 @@ describe('given default Tooltip', () => {
       expect(document.body.innerHTML).not.toContain('Add to library')
     })
   })
+
+  it('should close when another tooltip broadcasts that it opened', async () => {
+    await wrapper.find('button').trigger('focus')
+    expect(document.body.innerHTML).toContain('Add to library')
+
+    document.dispatchEvent(new CustomEvent(TOOLTIP_OPEN))
+    await flushPromises()
+
+    expect(document.body.innerHTML).not.toContain('Add to library')
+  })
 })
 
 describe('given tooltip within TooltipProvider', () => {
@@ -68,6 +84,11 @@ describe('given tooltip within TooltipProvider', () => {
       },
       attachTo: document.body,
     })
+  })
+
+  afterEach(async () => {
+    wrapper.unmount()
+    await flushPromises()
   })
 
   it('should use the provider content values', async () => {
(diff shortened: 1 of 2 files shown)


zernonia:
LGTM! Thanks for the PR @lazerg !

PR #2874, opened 2026-08-14
fix(DismissableLayer): remove the stale branch element from the registry on unmount

### LINKED ISSUE #2873: DismissableLayerBranch leaks unmounted elements into the shared context.branches Set (stale currentElement.value read in onUnmounted, same bug as #2778)
I traced #2778 ("fix(Stepper): remove stale trigger element from totalStepperItems on unmount"), which fixed a stale-ref bug in `StepperTrigger.vue`: the component added `currentElement.value` to a shared reactive `Set` on mount and tried to remove it with `rootContext.totalStepperItems.value.delete(currentElement.value)` on unmount, but by the time `onUnmounted` runs, Vue has already unset the ref backing `currentElement` (a computed over `unrefElement(currentRef)`), so `currentElement.value` reads `null`/`undefined` at that point and `delete()` is a no-op against the actual registered element. The fix captures the element in a plain local variable at mount time and deletes that captured reference on unmount instead of re-reading the now-stale computed.

`packages/core/src/DismissableLayer/DismissableLayerBranch.vue` has the exact same shape, untouched by that fix:

```vue
const { forwardRef, currentElement } = useForwardExpose()
onMounted(() => {
  context.branches.add(currentElement.value)
})
onUnmounted(() => {
  context.branches.delete(currentElement.value)
})
```

`context.branches` is a module-level, app-wide `reactive` `Set<HTMLElement>` (defined in `DismissableLayer/context.ts`), so this isn't scoped per-instance the way it might be in Stepper. `currentElement` here is the exact same `useForwardExpose()`-provided computed as `StepperTrigger.vue` used before the fix, so it reads `null`/`undefined` in `onUnmounted` for the same reason, and `context.branches.delete(currentElement.value)` never actually removes the element that was added on mount.

Effect: every `DismissableLayerBranch` instance that unmounts leaks its DOM node into `context.branches` permanently. `DismissableLayerBranch` is a public export (`DismissableLayer/index.ts`) and is also used internally by `ToastViewport.vue`, so any app that mounts/unmounts toasts, or uses the primitive directly to exclude a portal from a `DismissableLayer`'s outside-click detection, accumulates detached nodes in this Set for the lifetime of the page. Every `DismissableLayer` instance's outside-pointerdown/focus handling does a linear scan over `context.branches` (`[...context.branches].some(branch => branch?.contains(target))`), so the leak also makes that check progressively more expensive the longer the app runs.

Suggested fix, mirroring #2778's approach:

```vue
let registeredElement: HTMLElement | null = null
onMounted(() => {
  registeredElement = currentElement.value
  if (registeredElement) context.branches.add(registeredElement)
})
onUnmounted(() => {
  if (registeredElement) context.branches.delete(registeredElement)
})
```

Checked against the current `main` source of `DismissableLayerBranch.vue`, `useForwardExpose.ts`, and `DismissableLayer/context.ts`.

lazerg:
### 🔗 Linked issue

Fixes #2873

### ❓ Type of change

### 📚 Description

`DismissableLayerBranch` adds `currentElement.value` to `context.branches` on mount and deletes it again on unmount. By the time `onUnmounted` runs, Vue has already cleared the ref behind that computed, so the delete is a no-op and the element stays in the Set forever. Since `context.branches` is module-level, every branch that unmounts leaks a detached node for the lifetime of the page, and the outside-pointerdown and focus checks in `DismissableLayer` scan the whole Set on every event.

Same fix as #2778 did for `StepperTrigger`: keep the registered element in a local variable and delete that exact reference on unmount.

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dismissal behavior for nested layers by reliably registering branch elements when mounted and removing them when unmounted.
  * Prevented invalid or unavailable elements from being registered during the component lifecycle.
* **Tests**
  * Added coverage confirming that branch registration and cleanup work correctly when components mount and unmount.

### DIFF
--- a/packages/core/src/DismissableLayer/DismissableLayer.test.ts
+++ b/packages/core/src/DismissableLayer/DismissableLayer.test.ts
@@ -5,7 +5,7 @@ import { beforeEach, describe, expect, it } from 'vitest'
 import { defineComponent, h, nextTick, ref } from 'vue'
 import { useBodyScrollLock } from '@/shared/useBodyScrollLock'
 import { sleep } from '@/test'
-import { DismissableLayer as DismissableLayerPrimitive } from '.'
+import { DismissableLayerBranch, DismissableLayer as DismissableLayerPrimitive } from '.'
 import { context } from './context'
 import DismissableLayer from './story/_DismissableLayer.vue'
 import { isLayerExist } from './utils'
@@ -21,6 +21,23 @@ describe('isLayerExist', () => {
 
     expect(isLayerExist(layer, document as any)).toBe(false)
     expect(isLayerExist(layer, document.createTextNode('x') as any)).toBe(false)
+  })
+})
+
+describe('given a DismissableLayerBranch', () => {
+  beforeEach(() => {
+    document.body.innerHTML = ''
+    context.branches.clear()
+  })
+
+  it('should leave the branch registry empty after unmounting', async () => {
+    const wrapper = mount(DismissableLayerBranch, { attachTo: document.body })
+    await nextTick()
+    expect(context.branches.size).toBe(1)
+
+    wrapper.unmount()
+    await nextTick()
+    expect(context.branches.size).toBe(0)
   })
 })
 
--- a/packages/core/src/DismissableLayer/DismissableLayer.test.ts
+++ b/packages/core/src/DismissableLayer/DismissableLayer.test.ts
@@ -33,10 +33,13 @@ describe('given a DismissableLayerBranch', () => {
   it('should leave the branch registry empty after unmounting', async () => {
     const wrapper = mount(DismissableLayerBranch, { attachTo: document.body })
     await nextTick()
+    const branch = wrapper.element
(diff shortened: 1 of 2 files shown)


MILLERMARRU (commented):
I pulled the actual source for DismissableLayerBranch.vue (note: this repo's default branch is v2, not main) and confirmed both `onMounted` and `onUnmounted` read `currentElement.value` fresh rather than capturing it once. Since `context.branches` is a module-level Set shared across the whole app and consulted by `useBodyScrollLock` before it releases the body's `pointer-events` style, a branch element that fails to deregister isn't just a local leak, it can affect scroll-lock behavior for every future dismissable layer in the app for the lifetime of the page. The fix of capturing `registeredElement` at mount time and nulling it after cleanup is the standard pattern for this class of bug, and the added test directly asserts the registry is empty after unmount, which is the right thing to assert given the bug is specifically about staleness rather than about registration ever working in the first place.

PR #2881, opened 2026-08-16
fix(Drawer): toggle open state on trigger click

### LINKED ISSUE #2872: [Bug]: DrawerTrigger cannot close an open drawer — click always calls onOpenChange(true), unlike DialogTrigger and vaul
### Environment

```bash
Development OS: Linux (Debian 12, kernel 6.1)
Node version: 22.23.2
Package manager: npm
Reka UI version: 2.10.3
Vue version: 3.5.18
Nuxt version: — (plain Vue 3 + Vite)
CSS framework: tailwindcss@4.1.1
Client OS: Linux
Browser: Chromium 139
```

The bug is directly visible in one line of source, so per the template's escape hatch I'm providing source permalinks plus an inline SFC instead of a sandbox (happy to add a StackBlitz if you still want one):

- `DrawerTrigger.vue` click handler hard-codes **open**: [`@click="rootContext.onOpenChange(true, 'trigger-press')"`](https://github.com/unovue/reka-ui/blob/14866201d179b8bae3c8b4346a1ca8eff1c5eaa4/packages/core/src/Drawer/DrawerTrigger.vue#L40)
- `DrawerRoot.onOpenChange` early-returns when the value is unchanged, so a repeated trigger click doesn't even emit `update:open`: [DrawerRoot.vue#L166-L168](https://github.com/unovue/reka-ui/blob/14866201d179b8bae3c8b4346a1ca8eff1c5eaa4/packages/core/src/Drawer/DrawerRoot.vue#L166-L168)

Minimal SFC:

```vue
<script setup lang="ts">
import { DrawerContent, DrawerDescription, DrawerRoot, DrawerTitle, DrawerTrigger } from 'reka-ui'
</script>

<template>
  <DrawerRoot modal="trap-focus">
    <DrawerTrigger>Toggle drawer</DrawerTrigger>
    <DrawerContent>
      <DrawerTitle>Drawer</DrawerTitle>
      <DrawerDescription>Click the trigger again — nothing happens.</DrawerDescription>
    </DrawerContent>
  </DrawerRoot>
</template>
```

### Steps to reproduce

1. Render a `DrawerRoot` in a mode where the trigger stays clickable while the drawer is open — e.g. `modal="trap-focus"` (or `:modal="false"`, or default modal with `@interact-outside` prevented for the trigger area).
2. Click the `DrawerTrigger` → the drawer opens.
3. Click the same `DrawerTrigger` again.

**Actual:** nothing happens — the drawer stays open. No `update:open` is emitted either, so a controlled parent can't even intercept the second click through Reka.

### Describe the bug

`DrawerTrigger`'s click handler always calls `onOpenChange(true, 'trigger-press')`, and `DrawerRoot.onOpenChange` starts with `if (open.value === value) return`. Together this makes a second click on the trigger of an open drawer a complete no-op.

In the default `modal` mode this is masked: the overlay covers the trigger, and pointer-down-outside dismisses the drawer before the click lands. But in `trap-focus` / non-modal modes — which exist precisely so the surrounding UI stays interactive — the trigger is reachable while open, and it silently does nothing.

This diverges from both of Reka's own reference points:

- Reka's `DialogTrigger` **toggles**: its click handler is [`onClick: onOpenToggle`](https://github.com/unovue/reka-ui/blob/14866201d179b8bae3c8b4346a1ca8eff1c5eaa4/packages/core/src/Dialog/DialogTrigger.vue).
- The original vaul (React), which the Drawer primitive ports, doesn't define its own trigger at all — [`Trigger: DialogPrimitive.Trigger`](https://github.com/emilkowalski/vaul/blob/main/src/index.tsx#L1142) is the raw Radix Dialog trigger, whose [click handler is `onOpenToggle`](https://github.com/radix-ui/primitives/blob/main/packages/react/dialog/src/dialog.tsx#L135). So in the React original a repeated trigger click closes the drawer.

Real-world case where this bites: a mobile bottom tab bar whose slots open drawers (`modal="trap-focus"`, interact-outside prevented for the bar so that switching slots works). Tapping the active slot again is expected to close its drawer — the universal tab-bar convention — but with the current trigger it's impossible without bypassing `DrawerTrigger` entirely.

### Expected behavior

Clicking the trigger of an open drawer closes it — i.e. `DrawerTrigger` behaves like `DialogTrigger` (`onOpenToggle`), matching the aria pattern (`aria-expanded` on the trigger already advertises toggle semantics) and vaul's behavior.

If the always-open behavior is intentional for swipe-related reasons, documenting it and providing an opt-in (e.g. a `toggle` prop) would also work — but parity with `DialogTrigger` seems like the least surprising default.

Found while implementing the tab-bar-with-drawers pattern described above; verified against both v2.10.3 (npm latest) and current `main` (1486620).

NasreddinHodja:
### 🔗 Linked issue

Fixes #2872

### ❓ Type of change

### 📚 Description

`DrawerTrigger`'s click handler hardcodes `onOpenChange(true, 'trigger-press')`, so it can only ever open the drawer. That's invisible in the default modal mode, where an overlay sits over the page and blocks the trigger once the drawer is open — but in `:modal="false"` or `"trap-focus"` mode the trigger stays clickable while open, and a second click does nothing instead of closing the drawer. `DialogTrigger` doesn't have this problem: it calls `onOpenToggle`, exposed on `DialogRootContext`.

This mirrors that pattern for Drawer: `DrawerRootContext` gains its own `onOpenToggle`, built on the existing guarded `onOpenChange` logic (tagged with the same `'trigger-press'` reason the old hardcoded call used), and `DrawerTrigger`'s click handler now calls it instead of hardcoding `true`.

Worth flagging: while extracting `onOpenChange` into its own function, I noticed `handleClose` (used by the root slot's `close` prop) duplicates the exact same guard/emit shape and could be reduced to `handleOpenChange(false, 'close-press')`. Left it as-is to keep this PR scoped to the toggle bug — happy to fold that in here or as a follow-up, whichever you'd prefer.

---

This is my first contribution to reka-ui, and to open source generally — paired with Claude Code to work through the codebase's conventions while writing this, but the reasoning and decisions here are mine. Happy to take feedback on approach, style, anything.

### 📸 Screenshots

_(none — behavior-only change, no visual difference)_

### 📝 Checklist

- [ ] I have updated the documentation accordingly. _(not applicable — no public props/emits/slots changed)_

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the Drawer trigger so clicking it while the Drawer is open now closes the Drawer correctly.
  * Drawer state updates now report the `trigger-press` reason consistently.

* **Tests**
  * Added regression coverage for toggling the Drawer closed through its trigger.

### DIFF
--- a/packages/core/src/Drawer/Drawer.test.ts
+++ b/packages/core/src/Drawer/Drawer.test.ts
@@ -1,4 +1,5 @@
 import type { Mock, MockInstance } from 'vitest'
+import userEvent from '@testing-library/user-event'
 import { findByText, fireEvent, render } from '@testing-library/vue'
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest'
 import { axe } from 'vitest-axe'
@@ -159,9 +160,10 @@ describe('update:open change event details', () => {
     components: { DrawerRoot, DrawerTrigger, DrawerPortal, DrawerContent, DrawerTitle, DrawerClose },
     props: {
       onOpenChange: { type: Function, required: true },
+      modal: { type: [Boolean, String], default: true },
     },
     template: `
-      <DrawerRoot @update:open="onOpenChange">
+      <DrawerRoot @update:open="onOpenChange" :modal="modal">
         <DrawerTrigger>Open</DrawerTrigger>
         <DrawerPortal>
           <DrawerContent>
@@ -204,6 +206,19 @@ describe('update:open change event details', () => {
     onOpenChange.mockClear()
     await fireEvent.click(getByText('Open'))
     await nextTick()
+    expect(onOpenChange).toHaveBeenCalledWith(false, { reason: 'trigger-press' })
+  })
+
+  it('closes on a second trigger click in non-modal mode', async () => {
+    const user = userEvent.setup()
+    const onOpenChange = vi.fn()
+    const { getByText } = render(DrawerWithReason, { props: { onOpenChange, modal: false } })
+    await user.click(getByText('Open'))
+    await nextTick()
+    onOpenChange.mockClear()
+    await user.click(getByText('Open'))
+    await nextTick()
+    expect(onOpenChange).toHaveBeenCalledTimes(1)
     expect(onOpenChange).toHaveBeenCalledWith(false, { reason: 'trigger-press' })
   })
 })
--- a/packages/core/src/Drawer/Drawer.test.ts
+++ b/packages/core/src/Drawer/Drawer.test.ts
@@ -195,4 +195,15 @@ describe('update:open change event details', () => {
     await nextTick()
     expect(onOpenChange).toHaveBeenCalledWith(false, { reason: 'close-press' })
   })
+
+  it('closes on a second trigger click (toggle)', async () => {
+    const onOpenChange = vi.fn()
+    const { getByText } = render(DrawerWithReason, { props: { onOpenChange } })
+    await fireEvent.click(getByText('Open'))
+    await nextTick()
+    onOpenChange.mockClear()
+    await fireEvent.click(getByText('Open'))
+    await nextTick()
+    expect(onOpenChange).toHaveBeenCalledWith(false, { reason: 'trigger-press' })
+  })
 })
--- a/packages/core/src/Drawer/DrawerRoot.vue
+++ b/packages/core/src/Drawer/DrawerRoot.vue
@@ -206,6 +206,10 @@ provideDrawerRootContext({
   descriptionId,
 })
 
+function handleClose() {
+  handleOpenChange(false, 'close-press')
+}
+
 function handleOpenChange(value: boolean, reason?: DrawerOpenChangeReason) {
   if (open.value === value)
     return
@@ -240,6 +244,6 @@ onUnmounted(() => {
 <template>
   <slot
     :open="open"
-    :close="() => handleOpenChange(false, 'close-press')"
+    :close="handleClose"
   />
 </template>
--- a/packages/core/src/Drawer/DrawerRoot.vue
+++ b/packages/core/src/Drawer/DrawerRoot.vue
@@ -206,13 +206,6 @@ provideDrawerRootContext({
   descriptionId,
 })
 
-function handleClose() {
-  if (!open.value)
-    return
-  uncontrolledOpen.value = false
-  emit('update:open', false, { reason: 'close-press' })
-}
-
 function handleOpenChange(value: boolean, reason?: DrawerOpenChangeReason) {
   if (open.value === value)
     return
@@ -247,6 +240,6 @@ onUnmounted(() => {
 <template>
   <slot
     :open="open"
-    :close="handleClose"
+    :close="() => handleOpenChange(false, 'close-press')"
   />
 </template>
--- a/packages/core/src/Drawer/DrawerRoot.vue
+++ b/packages/core/src/Drawer/DrawerRoot.vue
@@ -69,6 +69,7 @@ export interface DrawerRootContext {
   isSwiping: Ref<boolean>
   nestedSwipeProgressStore: NestedSwipeProgressStore
   onOpenChange: (value: boolean, reason?: DrawerOpenChangeReason) => void
+  onOpenToggle: () => void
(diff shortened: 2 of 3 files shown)


zernonia:
Thanks for this — really well-scoped first contribution. Right pattern chosen, conventions matched, and flagging the adjacent cleanup in the description instead of silently folding it in is exactly what makes a PR easy to review.

The fix itself is correct and I'm happy to take it. One change I'd like on the test before merging.

### The fix

`onOpenToggle` on `DrawerRootContext` placed right after `onOpenChange`, built on the shared guard, mirrors `DialogRoot`/`DialogTrigger` exactly, which is what I'd want here. Controlled mode still behaves correctly (the `uncontrolledOpen` write is inert, the emit still fires), and no docs regeneration is needed since no public props/emits/slots changed.

I also checked the interaction most likely to break from this change, and it holds: in `:modal="false"` / `"trap-focus"`, the `pointerdown` on the trigger does reach `DismissableLayer` as an outside-press, but `DrawerContent`'s non-modal `onInteractOutside` handler `preventDefault()`s when the target is inside `triggerElement`, so no dismiss fires and the following `click` toggle isn't a close-then-reopen. Verified with a real pointer sequence (`userEvent`) in both modes — exactly one `update:open` emit, `false` with `trigger-press`.

Full core suite passes, `vue-tsc` and eslint clean.

### Test: please cover the mode the bug actually lives in

`DrawerWithReason` renders the default `modal={true}` drawer, and `fireEvent.click` dispatches only a bare `click`. That bypasses both things that make this bug real:

- the overlay / `disable-outside-pointer-events` that makes the trigger unreachable in modal mode — so the test asserts toggling in a configuration where a real user can never click the trigger anyway;
- the `pointerdown` → `DismissableLayer` outside-press path, which is the actual risk in the modes issue #2872 is about.

It does fail against the old hardcoded `onOpenChange(true, …)`, so it has regression value — but it locks in a behavior nobody can reach, and leaves the `triggerElement.contains(target)` guard (the thing that makes the fix work in practice) uncovered. Something like:

```ts
it('closes on a second trigger click in non-modal mode', async () => {
  const user = userEvent.setup()
  const onOpenChange = vi.fn()
  const { getByText } = render(DrawerWithReason, { props: { onOpenChange, modal: false } })
  await user.click(getByText('Open'))
  await nextTick()
  onOpenChange.mockClear()
  await user.click(getByText('Open'))
  await nextTick()
  expect(onOpenChange).toHaveBeenCalledTimes(1) // not close-then-reopen
  expect(onOpenChange).toHaveBeenCalledWith(false, { reason: 'trigger-press' })
})
```

`DrawerWithReason` needs a `modal` prop threaded through; `@testing-library/user-event` is already a dev dep and used elsewhere in the suite. The `toHaveBeenCalledTimes(1)` is the assertion that actually pins the outside-press guard. Keep the existing modal-mode test alongside it if you like.

### On your `handleClose` question

You're right, and yes — please fold it into this PR. `handleClose` is behaviorally identical to `handleOpenChange(false, 'close-press')` (the `!open.value` guard collapses to the `open.value === value` guard when `value` is `false`, same emit shape). It's a five-line delete in a function this PR is already touching, so it costs nothing in review surface.

Two things I noticed while reading that are **not** yours to fix here, just noting them:

- `pendingDismissReason` in `DrawerContentImpl` is set even when the dismiss ends up prevented, and never cleared. Harmless today because the fallback is the same `'outside-press'`, but it's stale state.
- After a toggle-close in non-modal, `hasInteractedOutside` is `true`, so `onCloseAutoFocusNonModal` skips the explicit `triggerElement.focus()`. That's the same structure `DialogContentNonModal` has always had, so this PR brings Drawer to parity rather than introducing anything new.

NasreddinHodja:
Thanks for the thorough review!

Pushed two commits addressing both:

- `refactor(Drawer): fold handleClose into handleOpenChange` — folded as suggested.
- `test(Drawer): cover trigger toggle in non-modal mode with a real pointer sequence` — threaded `modal` through `DrawerWithReason` and added the `userEvent` + `modal: false` test from your snippet, asserting `toHaveBeenCalledTimes(1)` to pin the outside-press guard. Kept the existing `modal: true` test alongside it.

PR #2885, opened 2026-08-18
fix(Combobox): defer highlight scroll until Popper placement

### LINKED ISSUE #2884: [Bug]: Combobox highlights scroll the document before Popper placement
### Environment

```text
Development OS: macOS 26.6
Node: 24.15.0
Package manager: pnpm 10.34.4
Reka UI: 2.10.1
Vue: 3.5.33
Nuxt: N/A
Build tool: Vite 8.1.5
CSS framework: Tailwind CSS 4.3.0
Client OS: macOS 26.6
Browser: Chrome
```

https://stackblitz.com/edit/hn2yox-2uvmbefw?file=src%2FApp.vue

### Steps to reproduce

1. Open the reproduction and wait for the preview.
2. Scroll until the Combobox input is comfortably visible.
3. Focus the input and type a character.

### Describe the bug

Highlighting the first option scrolls the document even though the input is already visible.

`ListboxRoot.changeHighlight` calls `scrollIntoView` before Floating UI has placed the portalled Popper content. The wrapper still has its off-screen measurement transform (`translate(0, -200%)`), so the option scroll can move the page. In one trace, `scrollY` changed from 1342 to 0 before `placed`.

We have a narrow tested downstream patch and can submit a PR.

### Expected behavior

Highlight state, events, and `aria-activedescendant` should update immediately without moving the document. Once Popper is placed, the current highlighted option may be scrolled into view.

Our working approach suppresses only highlight `scrollIntoView` while a Popper-positioned Combobox is unplaced, then scrolls the current connected highlight on `placed`. Standalone Listbox behavior remains unchanged.

gavgrego:
### 🔗 Linked issue

Resolves #2884

### ❓ Type of change

### 📚 Description

Portalled Combobox content is rendered off-screen while Popper measures it. Highlighting an option during that window called `scrollIntoView`, which could move the page.

This defers only the highlight scroll until Popper emits `placed`, then scrolls the current connected highlight once. Highlight state, focus, inline Comboboxes, standalone Listboxes, and positioning strategy remain unchanged. Autocomplete receives the same coordination because it shares Combobox content.

Regression tests cover pre-placement keyboard changes, repeated placement, late-mounted options, force-mounted content, remounts, and inline behavior.

Reproduction: https://stackblitz.com/edit/hn2yox-2uvmbefw?file=src%2FApp.vue

### 📝 Checklist

- [x] I have updated the documentation accordingly. No public API or documentation changes are required.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved highlighted-option scrolling in comboboxes using popper-positioned content.
  * Scrolling now waits for placement to complete, including delayed mounting and remounting scenarios.
  * Preserved immediate scrolling for inline combobox content.
  * Prevented unwanted scrolling during content repositioning, including nested listbox interactions.

### DIFF
--- a/packages/core/src/Autocomplete/AutocompleteRoot.vue
+++ b/packages/core/src/Autocomplete/AutocompleteRoot.vue
@@ -147,7 +147,7 @@ const inputElement = ref<HTMLInputElement>()
 const triggerElement = ref<HTMLElement>()
 
 const highlightedElement = computed(() => primitiveElement.value?.highlightedElement ?? undefined)
-const contentPositioning = useComboboxContentPositioning(open, highlightedElement)
+const contentPositioning = useComboboxContentPositioning(open)
 
 const allItems = ref<Map<string, string>>(new Map())
 const allGroups = ref<Map<string, Set<string>>>(new Map())
--- a/packages/core/src/Autocomplete/AutocompleteRoot.vue
+++ b/packages/core/src/Autocomplete/AutocompleteRoot.vue
@@ -3,6 +3,7 @@ import type { Ref } from 'vue'
 import type { PrimitiveProps } from '@/Primitive'
 import type { Direction, GenericComponentInstance } from '@/shared/types'
 import { provideComboboxRootContext } from '@/Combobox/ComboboxRoot.vue'
+import { useComboboxContentPositioning } from '@/Combobox/useComboboxContentPositioning'
 import { usePrimitiveElement } from '@/Primitive'
 import { createContext, useDirection, useFilter } from '@/shared'
 
@@ -146,6 +147,7 @@ const inputElement = ref<HTMLInputElement>()
 const triggerElement = ref<HTMLElement>()
 
 const highlightedElement = computed(() => primitiveElement.value?.highlightedElement ?? undefined)
+const contentPositioning = useComboboxContentPositioning(open, highlightedElement)
 
 const allItems = ref<Map<string, string>>(new Map())
 const allGroups = ref<Map<string, Set<string>>>(new Map())
@@ -218,6 +220,7 @@ provideComboboxRootContext({
   disabled,
   open,
   onOpenChange,
+  ...contentPositioning,
   contentId: '',
   isUserInputted,
   isVirtual,
--- a/packages/core/src/Combobox/Combobox.test.ts
+++ b/packages/core/src/Combobox/Combobox.test.ts
@@ -1,10 +1,10 @@
 import type { DOMWrapper, VueWrapper } from '@vue/test-utils'
-import { mount } from '@vue/test-utils'
+import { flushPromises, mount } from '@vue/test-utils'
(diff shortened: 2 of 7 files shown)


zernonia:
Thanks for the detailed writeup and tests — the approach looks right: highlight state and `aria-activedescendant` stay immediate, only the scroll defers, and inline/standalone Listbox behaviour is untouched. Verified on `e10d878`: full core suite (2021 tests), `vue-tsc` and lint all pass.

Two things I'd like addressed before merge:

**1. `focus()` bypasses the new guard.** In `ListboxRoot.changeHighlight` only `scrollIntoView` is gated — the `highlightedElement.value.focus()` line above it still runs, and `focus()` scrolls by default. `focusable` is only flipped to `false` by `ListboxFilter` (rendered by `ComboboxInput`), so a trigger-only Combobox (`ComboboxTrigger` + `ComboboxContent`, no input) still jumps the page before `placed`. Checked with a stubbed `PopperContent`: pre-placement `scrollIntoView` calls = 0 ✅ but `focus()` on `[role="option"]` = 1. `highlightSelected` already couples the two for exactly this reason (`const focus = scroll ? undefined : false`), so `focus({ preventScroll: true })` while suppressed would be consistent. Same hole at `ListboxVirtualizer.vue:129` (`if (event) item?.focus()`).

**2. The deferred scroll ignores callers that asked for no scroll.** `onContentPlaced` scrolls whatever happens to be highlighted, but several call sites pass `scrollIntoView = false` on purpose — `highlightOnHover` (`ListboxItem.vue:96`) and the virtualizer's mount highlight (`ListboxVirtualizer.vue:127`/`143`), the latter specifically so a virtualized list below the fold doesn't pull the page. Suggest recording inside `changeHighlight` that a scroll *was* suppressed and flushing only that on `placed`; it also removes the need to pass `highlightedElement` into the composable.

Minor: `onContentUnmount` resets `contentPosition` unconditionally, so an outgoing instance unmounting after an incoming one's `setup` desyncs the state (fails open, low risk, but worth scoping to the current content); and `@placed` is bound even in `inline` mode, where it lands on `Primitive` as a native DOM listener.

gavgrego:
@zernonia sounds good! Will get these updates up.  Thanks for the review.

gavgrego:
Addressed in `a4331eed`:

- pre-placement option focus now uses `focus({ preventScroll: true })`; the virtualizer routes its event-driven focus through the same guarded path
- deferred scrolling now records the actual suppressed request in `changeHighlight`, so `scrollIntoView = false` cancels the pending flush instead of scrolling the current highlight on placement
- content lifecycle events are scoped to the active content instance, and `placed` is only bound for popper content

Added regressions for trigger-only focus, hover opt-out, virtualizer mount/focus, and overlapping content replacement. Full core suite passes (97 files / 2026 tests), along with `vue-tsc` and changed-file lint.

PR #2890, opened 2026-08-25
perf(useFilter): short-circuit contains on exact substring match

Cherry:
### ❓ Type of change

### 📚 Description

`contains()` walks the haystack one position at a time, calling `Intl.Collator.compare` on a fresh `slice()` at each step. For a match late in the string that is O(n) collator comparisons, and `ComboboxRoot`'s `filterState` calls it once per item per keystroke.

Three changes, all behaviour-preserving:

1. **Exact-substring fast path.** `String.prototype.includes` is checked first. Identical strings collate equal under any `Intl.Collator` options, so a plain substring hit is  always a collator hit. This only short-circuits positive results; negatives still run
   the full scan.
2. **Hoist `collator.value.compare` out of the loop.** It was a reactive `computed` deref *per character position*, plus a fresh bound-function access each iteration.
3. **Length bail.** `substring.length > string.length` returns early. The loop guard `scan + sliceLen <= string.length` already implied this, so it is a pure shortcut.

`startsWith` / `endsWith` are deliberately untouched: they are already O(1) in collator calls, and adding a length precheck there could change results for collator-ignorable characters (e.g. a needle whose extra character is ignorable).

2000-item list, `useFilter({ sensitivity: 'base' })`, labels shaped `"<word> server <n>"`,
timing one full pass over the list. Node 24.19.0, best of 5.

| needle | matches at | before | after | |
|---|---|---|---|---|
| `"minecraft"` | position 0 | 3.27 ms | 3.19 ms | 1.03x |
| `"server"` | mid-string | 3.25 ms | 0.40 ms | **8.12x** |
| `"zzzzzz"` | no match | 4.23 ms | 4.07 ms | 1.04x |
| `"a"` | early | 1.74 ms | 1.20 ms | 1.45x |

useFilter had no tests, so this PR adds 15 covering all three functions, sensitivity options, reactive options, and NFC/NFD normalization.

They were written against the existing implementation and verified to pass on it before the perf change was applied, so they characterize current behaviour rather than the new code's behaviour.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved filter matching for exact matches, oversized search terms, and locale-aware text comparisons.
  - Enhanced handling of Unicode normalization and collator sensitivity options.
  - Ensured filter results update correctly when matching options change, including reactive updates and computed results.

- **Tests**
  - Added comprehensive coverage for prefix, suffix, substring, empty, oversized, absent, exact, and Unicode matching scenarios.
  - Added coverage for locale-sensitive comparisons, reactive option updates, and computed-state invalidation.

### DIFF
--- a/packages/core/src/shared/useFilter.test.ts
+++ b/packages/core/src/shared/useFilter.test.ts
@@ -1,0 +1,90 @@
+import { describe, expect, it } from 'vitest'
+import { ref } from 'vue'
+import { useFilter } from './useFilter'
+
+describe('useFilter', () => {
+  const { startsWith, endsWith, contains } = useFilter()
+
+  describe('startsWith', () => {
+    it('matches a prefix', () => {
+      expect(startsWith('hello', 'he')).toBe(true)
+      expect(startsWith('hello', 'lo')).toBe(false)
+    })
+
+    it('treats an empty substring as a match', () => {
+      expect(startsWith('hello', '')).toBe(true)
+    })
+
+    it('returns false when the substring is longer than the string', () => {
+      expect(startsWith('he', 'hello')).toBe(false)
+    })
+  })
+
+  describe('endsWith', () => {
+    it('matches a suffix', () => {
+      expect(endsWith('hello', 'lo')).toBe(true)
+      expect(endsWith('hello', 'he')).toBe(false)
+    })
+
+    it('treats an empty substring as a match', () => {
+      expect(endsWith('hello', '')).toBe(true)
+    })
+
+    it('returns false when the substring is longer than the string', () => {
+      expect(endsWith('lo', 'hello')).toBe(false)
+    })
+  })
+
+  describe('contains', () => {
+    it('matches at the start, middle and end', () => {
+      expect(contains('hello', 'he')).toBe(true)
+      expect(contains('hello', 'ell')).toBe(true)
+      expect(contains('hello', 'lo')).toBe(true)
+    })
+
+    it('returns false when absent', () => {
+      expect(contains('hello', 'xyz')).toBe(false)
+    })
+
+    it('treats an empty substring as a match', () => {
+      expect(contains('hello', '')).toBe(true)
+    })
+
+    it('returns false when the substring is longer than the string', () => {
+      expect(contains('he', 'hello')).toBe(false)
+    })
+
+    it('matches an exact whole-string substring', () => {
+      expect(contains('hello', 'hello')).toBe(true)
+    })
+
+    it('matches across NFC/NFD normalization forms', () => {
+      // Precomposed "é" in the haystack, decomposed "e" + U+0301 in the needle.
+      expect(contains('café latte', 'café')).toBe(true)
+    })
+  })
+
+  describe('collator options', () => {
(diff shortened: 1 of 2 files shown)


zernonia:
Thanks for this — the optimization is well reasoned and the write-up made it easy to verify. I checked each of the load-bearing claims and they hold:

- **Detaching `compare` is safe.** `Intl.Collator.prototype.compare` returns a spec-bound function, so hoisting it doesn't lose `this`. `base.contains('Café Latte', 'cafe')` proves it — that case misses `includes` and exercises the detached `compare` on the loop path.
- **`includes` ⊆ collator match.** If `string.includes(substring)` at index *i*, then `slice(i, i + len) === substring`, and identical strings collate equal under any options. Strictly a positive short-circuit; negatives still run the full scan.
- **The length bail is a pure shortcut.** The old guard `0 + sliceLen <= string.length` already returned `false` without iterating.
- **The NFD test is genuine.** I byte-checked it — the needle in `contains('café latte', 'café')` really is decomposed `e` + U+0301, matching the comment.
- **The tests characterize existing behaviour as claimed.** All 15 pass against `v2`'s implementation unchanged.

One thing I think needs a fix before merge, though.

### The `includes()` fast path can drop the collator's reactive dependency

`contains` can now return before `collator.value` is ever dereferenced. When it does, a consuming `computed` never tracks the collator, so it won't re-evaluate when reactive options change:

```js
const options = ref({ sensitivity: 'variant' })
const { contains } = useFilter(options)
const filtered = computed(() => ['cat', 'cart', 'cabin'].filter(i => contains(i, 'ca')))

filtered.value                        // every item hits includes() — collator.value never read
options.value = { sensitivity: 'base' }
filtered.value                        // ✗ stale: the computed was never invalidated
```

I confirmed this both directions: the probe passes on `v2` (the old loop always ran at least one iteration, so `collator.value` was always read) and fails on this branch.

It needs *every* item to short-circuit — a single collator-path item re-arms tracking for the whole computed — which is why the new reactive-options test doesn't catch it: it calls `contains` directly rather than from inside an effect.

No in-repo consumer is affected today (`ComboboxRoot` and `AutocompleteRoot` both pass a static `{ sensitivity: 'base' }`), but `useFilter` is public API and its signature explicitly accepts `MaybeRef<Intl.CollatorOptions>`.

### Suggested fix

Hoist the deref above the fast path. It costs nothing — still one deref per call instead of one per character position, so the benchmark numbers should be unchanged:

```js
if (substring.length > string.length)
  return false

const compare = collator.value.compare // read before the fast path so the dep stays tracked
// Identical strings always collate equal, so an exact match short-circuits the scan.
if (string.includes(substring))
  return true

const sliceLen = substring.length
for (let scan = 0; scan + sliceLen <= string.length; scan++) {
  if (compare(substring, string.slice(scan, scan + sliceLen)) === 0)
    return true
}
```

Could you also add a test that filters from inside a `computed` with all items hitting the fast path? The current reactive test can't catch this class of bug, and it'd be an easy regression to reintroduce.

Cherry:
Good catch, thanks! Fixed in https://github.com/unovue/reka-ui/pull/2890/commits/e5d10e15ccc353e1919f55c05c454e7cef237256

PR #2901, opened 2026-08-26
feat(v3): headless Accordion/Collapsible composables

rastuhacode:
### 🔗 Linked issue

Continues #2723

### ❓ Type of change

### 📚 Description

Hey! I've had a free time and wanted to both help and take a look at a new release of `v3` version of Reka UI. To challenge myself and improve my skills I've created the headless composable for Accordion following existent recipe. I've decided to open a PR for it, if it is okay and you accept external help on this release. Thanks in advance!

#### Changes

- Added public `useAccordion.ts` and item/header/trigger/content surface builders.
- Converted Accordion SFCs into thin shells while preserving Collapsible wrappers, SSR IDs, events, keyboard navigation, and public APIs.
- Added characterization and composable tests in `useAccordion.test.ts`.
- Updated the `Superpowers migration recipe` from Accordion changes (added checkmark and the Disclosure refinement).

### 📝 Checklist

## Summary by CodeRabbit

* **New Features**
  * Added standalone Accordion and Collapsible composition APIs for custom interfaces.
  * Added cancellable state updates with change reasons and event details.
  * Collapsible content discovery events now include the originating browser event.

* **Documentation**
  * Updated event documentation to reflect payloads that may be undefined.
  * Expanded guidance for building custom Accordion and Collapsible interfaces.

* **Bug Fixes**
  * Improved Accordion keyboard navigation, disabled-item handling, controlled state, and server-side rendering.


packages/core/src/Accordion/useAccordion.ts
```diff
@@ -0,0 +1,260 @@
...
+
+    const values = Array.isArray(modelValue.value)
+      ? [...modelValue.value]
+      : [modelValue.value].filter((current): current is string => Boolean(current))
+    const index = values.indexOf(value)
+    if (index === -1)
+      values.push(value)
+    else
+      values.splice(index, 1)
+    modelValue.value = values
```
zernonia: **Regression: an `AccordionRoot` with no `type`, `v-model` or `default-value` becomes multiple.**

The shell now discards `isSingle`/`changeModelValue` from `useSingleOrMultipleValue` (`AccordionRoot.vue:96` keeps only `modelValue`) and this block re-infers the mode from the value. But `useSingleOrMultipleValue.getDefaultValue` seeds an untyped, valueless model with `[]` (`shared/useSingleOrMultipleValue.ts:39`), so `Array.isArray(inferredValue)` says multiple, while base `getDefaultType` falls back to `'single'`. Net effect: base emits `'one'`, this branch emits `['one']`. Documented usage hits it (`<AccordionRoot orientation="horizontal">` in `accordion.md`, `v-model` only in `AccordionSingle.story.vue`).

This `changeModelValue` is also a string-only re-implementation of the shared helper's (`isValueEqualOrExist` / `AcceptableValue`), so there are now two toggle implementations to keep in sync.

Suggested fix for both: accept `isSingle` and `changeModelValue` (or the whole `useSingleOrMultipleValue` return) as optional `UseAccordionProps`, have the shell pass them in, and keep the local implementation only as the standalone default. Then add a characterization test for `<AccordionRoot>` with no `type` and no model asserting the emitted value is a string.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/core/src/Accordion/AccordionItem.vue
```diff
@@ -58,26 +60,18 @@ defineSlots<{
...
+const { open, disabled } = surface
 
 const dataDisabled = computed(() => (disabled.value ? '' : undefined))
-
-const dataState = computed(() =>
-  open.value ? AccordionItemState.Open : AccordionItemState.Closed,
-)
+const dataState = computed<AccordionItemState>(() => surface.item.state.value.state)
 
 defineExpose({ open, dataDisabled })
```
zernonia: **Regression: `useForwardExpose()` now runs before `defineExpose({ open, dataDisabled })`.**

Base order was `defineExpose` first, then `useForwardExpose`. `useForwardExpose` snapshots `instance.exposed` and then replaces it; the compiled `__expose({ open, dataDisabled })` that now follows overwrites that replacement, dropping the forwarded prop getters, and Vue warns "expose() should be called only once per setup()" on every item in dev.

Please restore the original order (`defineExpose` above `useForwardExpose()`) and add a test on the item's exposed `open` / `dataDisabled` so this stays locked.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/core/src/Accordion/useAccordion.ts
```diff
@@ -0,0 +1,260 @@
...
+      unmountOnHide: toValue(options.unmountOnHide) ?? rootContext.unmountOnHide.value,
+      onKeydown: handleKeydown,
+    })),
+    state,
+  }
+  const itemContext: AccordionItemSurfaceContext = {
+    open,
+    disabled,
+    state,
+    triggerId: options.triggerId ?? '',
```
zernonia: **Standalone use yields empty ids.** With no `triggerId`, `getItemSurface('one')` renders `id=""` on the trigger and `aria-labelledby=""` on the content (line 90). The recipe's rule is that the composable defaults `baseId` to a literal so it stays callable outside `setup()`; please derive a default trigger id from a `baseId` option the same way `useTabs` does (`reka-accordion-trigger-<value>`), or make `triggerId` a getter/ref so the content surface's `props` computed actually tracks it (today it has no reactive dependency and is cached forever).

---
_Generated by [Claude Code](https://claude.ai/code)_

zernonia (changes requested):
Thanks for taking this on, and for following the recipe closely. This is exactly the kind of help #2723 needs, so apologies for the wait.

Overall the structure is right: the surface builders are clean, the Collapsible seam is handled correctly, and I verified parity on every rendered attribute (`role="region"`, `aria-labelledby`, `aria-expanded`, `aria-disabled`, `disabled`, the `reka-accordion-trigger-*` and Collapsible content ids, `data-reka-collection-item`, `data-state`/`data-disabled`/`data-orientation`, the CSS-variable `style`), keyboard navigation, `open`/`disabled`/`unmountOnHide` forwarding, the context shape, and the recipe amendment. `mergeProps` everywhere, handlers in `props`, no `useId`/`provide`/lifecycle inside the composable.

Two small regressions slipped through the shell, plus one standalone-mode gap. They are inline. Summary of what I'd like before merge:

1. Feed `isSingle` / `changeModelValue` from `useSingleOrMultipleValue` into `useAccordion` instead of re-inferring the mode, and add a characterization test for an untyped `<AccordionRoot>` with no model.
2. Restore `defineExpose` before `useForwardExpose()` in `AccordionItem.vue`, with a test on the exposed `open` / `dataDisabled`.
3. Give the standalone path real default ids (or a reactive `triggerId`).

Nice-to-have, not blocking: `AccordionItem` instantiates header/trigger/content surfaces it never uses (cheap, but dead), and Home/End, `dir="rtl"`, `loop`, and `unmountOnHide` through the shell have no characterization coverage yet.

Note that CI is green on this base but `docs:gen` was not run; no props/emits/slots changed, so I expect an empty diff.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/core/src/Accordion/useAccordion.ts
rastuhacode: Fixed: `AccordionRoot` now passes the existing `isSingle` and `changeModelValue` bridge through to `useAccordion`. Also added regression tests.

packages/core/src/Accordion/AccordionItem.vue
rastuhacode: Fixed: now `AccordionItem` restores `defineExpose()` before forwarding, also added regression tests.

packages/core/src/Accordion/useAccordion.ts
rastuhacode: Fixed: Surfaces now derive reactive IDs from `(baseId, value)`, defaulting to `reka-accordion-trigger-<value>`, while retaining explicit triggerId overrides. Tests were added.

zernonia:
Heads-up for the rebase: the composable foundation landed on `v3` in #2910, and the `data-state` rename is up in #2914. Two things change what this Accordion conversion should target:

- `useAccordion` should own its model through `useControllableState` (from `@/shared`) instead of `useVModel`, emit `beforeUpdate:modelValue` / `update:modelValue` as `[value, details]` with an `AccordionChangeReason`, and build each part with `createPartSurface` so the SFCs bind `part.attrs.value`. Switch, Tabs and Menu on `v3` are the reference conversions, and the recipe in `docs/superpowers/recipes/2723-headless-composable-recipe.md` is updated to the settled contract.
- `state.state` for the item/header/trigger/content parts should be typed `DisclosureState` and computed with `disclosureState(open)` from `@/shared`; the `AccordionItemState` enum is gone on `v3` after #2914.

Happy to answer questions on the recipe if anything is unclear.

---
_Generated by [Claude Code](https://claude.ai/code)_

rastuhacode:
rebased and refactored Accordion according to new changes (no pun intended)!

packages/core/src/Accordion/AccordionHeader.vue
```diff
@@ -24,9 +26,7 @@ useForwardExpose()
   <Primitive
     :as="props.as"
     :as-child="props.asChild"
-    :data-orientation="rootContext.orientation"
-    :data-state="itemContext.dataState.value"
-    :data-disabled="itemContext.dataDisabled.value"
+    v-bind="surface.attrs.value"
```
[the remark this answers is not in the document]
rastuhacode: Migration recipe explicitly permits this binding and Vue already merges inherited consumer listeners for this single-root component. The suggested mergeProps(surface.attrs.value) adds no additional listeners to merge.

docs/content/meta/AccordionRoot.md
```diff
@@ -70,10 +70,15 @@
...
+  {
+    'name': 'beforeUpdate:modelValue',
+    'description': '<p>Event handler called before a change; call details.cancel() to keep the current value.</p>\n',
+    'type': '[value: (ExplicitType extends \'single\' ? string : string[]), details: ChangeEventDetails&lt;AccordionChangeReason, Event&gt;]'
+  },
   {
     'name': 'update:modelValue',
     'description': '<p>Event handler called when the expanded state of an item changes</p>\n',
-    'type': '[value: (ExplicitType extends \'single\' ? string : string[])]'
+    'type': '[value: (ExplicitType extends \'single\' ? string : string[]), details: ChangeEventDetails&lt;AccordionChangeReason, Event&gt;]'
```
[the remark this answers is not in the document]
rastuhacode: Fixed `autogen.ts` and regenerated Accordion’s documentation. Both events now correctly include undefined. Props and slots are unchanged.

rastuhacode:
Hey!
I've decided to also add composable for the `Collapsible` in the same PR because `Accordion` and `Collapsible` have the same logic and `Accordion` integrates `Collapsible` behavior

packages/core/src/Collapsible/Collapsible.characterization.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  })
+}
+
+describe('collapsible characterization', () => {
+  it('links trigger and content and chains consumer clicks', async () => {
+    const onClick = vi.fn()
+    const wrapper = mountCollapsible({}, { onClick })
+    await nextTick()
+    const trigger = wrapper.find('button')
+    expect(trigger.attributes('aria-controls')).toBe(wrapper.find('[hidden]').attributes('id'))
```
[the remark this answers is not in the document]
rastuhacode: The content wrapper remains mounted because `Presence` uses `:force-mount="true"`. `unmountOnHide` only unmounts its slot children; the wrapper retains hidden and id. Therefore the ARIA-linkage assertion works with the default configuration.
Verified by running the unchanged characterization suite: all 3 tests pass.

PR #2923, opened 2026-09-05
fix(shared): suppress background with inert instead of aria-hidden alone

### LINKED ISSUE #2922: [Bug]: Modal content sets aria-hidden on the rest of the page without inert (axe aria-hidden-focus)
## Environment

```bash
Development OS: macOS 26.6.2
Node version: 26.7.0
Package manager: pnpm 11.23.0 (repo checks run with the pinned pnpm 10.13.1)
Reka UI version: 2.10.4
Vue version: 3.5.42
CSS framework: tailwindcss 4.3.3
Client OS: macOS 26.6.2
Browser: Chromium (Playwright build 1234), axe-core 4.10
```

## Link to minimal reproduction

https://reka-ui.com/docs/components/dropdown-menu

The docs demo reproduces it as-is — no sandbox needed, and nothing about the bug depends on app code. Steps below.

## Steps to reproduce

1. Open https://reka-ui.com/docs/components/dropdown-menu and click the trigger of the first example.
2. With the menu open, inspect the page wrapper element (the app root, `#__nuxt` on the docs site). It now carries `aria-hidden="true"` and `data-aria-hidden="true"`.
3. Note that the elements inside it are still focusable — no `inert`, no `tabindex="-1"`.
4. Run axe against the document while the menu is open (`axe.run()` in the console, or `@axe-core/playwright`):

```
[serious] ARIA hidden element must not be focusable or contain focusable elements
  https://dequeuniversity.com/rules/axe/4.10/aria-hidden-focus
  Selector: #app
  HTML: <div id="app" data-v-app="" data-aria-hidden="true" aria-hidden="true">
```

The same happens with `Dialog`, `Drawer`, `Popover`, `Select` and `Combobox` in modal mode.

## Describe the bug

`useHideOthers` marks everything outside the open content with `aria-hidden` only, so the elements underneath stay focusable in the DOM. That is exactly what axe-core's `aria-hidden-focus` rule flags, with **serious** impact.

`packages/core/src/shared/useHideOthers.ts`:

```ts
import { hideOthers } from 'aria-hidden'
// …
if (el && !isInsideClosedPopover)
  undo = hideOthers(el)
```

Used by `MenuRootContentModal`, `DialogContentModal`, `DrawerContent`, `PopoverContentModal`, `SelectContentImpl` and `ComboboxContentImpl`. The non-modal variants never call it.

**Why this probably went unnoticed:** the composable early-returns in test mode —

```ts
// disable hideOthers on test mode
if (import.meta.env.MODE === 'test')
  return
```

— so the colocated `vitest-axe` assertions never see the attribute being applied, and CI stays green.

**How bad is it in practice?** Worth saying plainly: users are unlikely to be affected today. `FocusScope` registers a `focusin` listener on `document` and pulls focus back into the container, so Tab does not actually reach the hidden elements. The violation is structural — axe reads the static DOM and cannot see a JS focus trap.

The practical cost is on the consuming side: any app that keeps an automated a11y assertion has to either suppress `aria-hidden-focus` globally (losing a rule that catches real problems elsewhere) or skip the assertion on every screen that opens a menu.

## Expected behavior

The background should be suppressed in a way that also removes it from the tab order, so the DOM matches the intent and the rule passes.

The `aria-hidden` package Reka already depends on exports exactly that:

```ts
/** Automatic function to "suppress" DOM elements - _hide_ or _inert_ in the best possible way */
export declare const suppressOthers: (originalTarget, parentNode?, markerName?) => Undo
```

```js
// aria-hidden 1.2.6
export var suppressOthers = function (originalTarget, parentNode, markerName) {
    return (supportsInert() ? inertOthers : hideOthers)(originalTarget, parentNode, markerName);
};
```

It applies `inert` where supported and falls back to today's behavior otherwise, so switching to it needs no new dependency and no version bump.

I intend to submit a PR for this — a two-line change in `useHideOthers.ts`. If changing the default is considered too broad, I am happy to gate it behind a prop or a `ConfigProvider` option instead.

## Context

Possibly related:

- #1280 — "[Bug]: Blocked aria-hidden on a `<input>` element" (open) looks like the same root cause, seen through the browser's own console warning.
- #354 — "[Feature]: Introduce Aria-Hidden `hideOthers`" (closed) is where the current behavior came from; `suppressOthers` may simply not have been considered then.
- #2240 — "[Bug]: Combobox with `open=true` adds `aria-hidden` on everything else" (closed) is about the scope of the hiding, not the missing `inert`.

geffersonluis:
### 🔗 Linked issue

Resolves #2922

### ❓ Type of change

### 📚 Description

`useHideOthers` marks everything outside the open content with `aria-hidden` only, leaving those elements focusable in the DOM. axe-core flags that as `aria-hidden-focus` (**serious**) on every screen where a modal primitive is open:

```
[serious] ARIA hidden element must not be focusable or contain focusable elements
  Selector: #app
  HTML: <div id="app" data-v-app="" data-aria-hidden="true" aria-hidden="true">
```

This switches to `suppressOthers()`, already exported by the `aria-hidden` dependency:

```js
// aria-hidden 1.2.6
export var suppressOthers = function (originalTarget, parentNode, markerName) {
    return (supportsInert() ? inertOthers : hideOthers)(originalTarget, parentNode, markerName);
};
```

It applies `inert` where the browser supports it and falls back to the current behavior otherwise.

**The change** — `packages/core/src/shared/useHideOthers.ts`:

```diff
 import type { MaybeElementRef } from '@vueuse/core'
 import { unrefElement } from '@vueuse/core'
-import { hideOthers } from 'aria-hidden'
+import { suppressOthers } from 'aria-hidden'
 import { onUnmounted, watch } from 'vue'

 export function useHideOthers(target: MaybeElementRef) {
-  let undo: ReturnType<typeof hideOthers>
+  let undo: ReturnType<typeof suppressOthers>
   watch(() => unrefElement(target), (el) => {
@@
     if (el && !isInsideClosedPopover)
-      undo = hideOthers(el)
+      undo = suppressOthers(el)
     else if (undo)
       undo()
   })
```

Both share the `(target, parentNode?, markerName?) => Undo` signature, so the call site, the teardown and the `[popover]:not(:popover-open)` guard are untouched. I kept the composable's name — renaming it would break anyone importing it from the shared entry, and "hide the others" still describes what it does.

**Scope.** Affects the primitives that use it in modal mode: `MenuRootContentModal`, `DialogContentModal`, `DrawerContent`, `PopoverContentModal`, `SelectContentImpl`, `ComboboxContentImpl`. Non-modal variants never called it.

**Compatibility.** No new dependency (`aria-hidden@^1.2.4` already ships both functions), and on browsers without `inert` the emitted attributes are identical to today. Where `inert` is supported, the background also leaves the tab order at the DOM level, so correctness no longer depends on `FocusScope`'s `focusin` listener being attached — the trap itself is unchanged.

**On tests.** I did not add one, and I would like your call on it. `useHideOthers` early-returns when `import.meta.env.MODE === 'test'`, so nothing under `vitest` ever exercises this path — which is also why the current behavior never showed up in the colocated `vitest-axe` assertions. Options I see:

1. Leave as is (manual verification, described below).
2. Narrow the guard so the composable runs in jsdom, then assert the attributes on the background in one colocated test. jsdom does not implement `inert` behavior, but `supportsInert()` checks `HTMLElement.prototype.inert !== undefined`, so the assertion would have to accept either branch.
3. Cover it in a browser-mode test if you have somewhere that fits better.

Tell me which you prefer and I will push it in this PR.

Not applicable — the change is an attribute on the background element. Before/after can be seen by inspecting the app root with any modal primitive open.

**Verified**

Repo checks, on this branch:

| | |
|---|---|
| `pnpm --filter reka-ui build` | passes (7.5s, including the `vue-tsc` step) |
| `pnpm --filter reka-ui exec vitest run` | 97 files, 2041 tests, all passing |
| `eslint` on the changed file | clean |

The suite passing only shows nothing regressed — it does not exercise this path, because of the test-mode early return described above.

So I verified the actual behavior end to end, by building the package and dropping the `dist` into a real Vue 3.5.42 + Vite app, then driving Chromium (Playwright) against a page with an open `DropdownMenu`:

```js
// before (reka-ui 2.10.4 as published)
{ ariaHidden: "true", inert: false }   // axe: 1 serious violation (aria-hidden-focus)

// after (this branch)
{ ariaHidden: null,   inert: true  }   // axe: no violations
```

Worth noting for the review: where `inert` is supported, `inertOthers` does not set `aria-hidden` at all — `inert` already removes the subtree from the accessibility tree and the tab order, so the end state is stronger than today's, not merely equivalent.

**To reproduce the check manually**

1. `pnpm --filter reka-ui build && pnpm docs:dev`
2. Open the DropdownMenu docs page and open the menu.
3. Inspect the app root: it carries `inert` instead of `aria-hidden`.
4. Run `axe.run()` in the console — the `aria-hidden-focus` violation is gone.

One gotcha when auditing by hand: wait for the open animation to finish before running axe. Reading computed styles mid-`fade-in` produces unrelated false positives (I measured a label at ~11% opacity reporting a contrast of 1.25 that reads 16:1 once settled).

### 📝 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accessibility behavior when isolating active interface elements by suppressing other content from assistive technologies.

### DIFF
--- a/packages/core/src/shared/useHideOthers.ts
+++ b/packages/core/src/shared/useHideOthers.ts
@@ -1,6 +1,6 @@
 import type { MaybeElementRef } from '@vueuse/core'
 import { unrefElement } from '@vueuse/core'
-import { hideOthers } from 'aria-hidden'
+import { suppressOthers } from 'aria-hidden'
 import { onUnmounted, watch } from 'vue'
 
 /**
@@ -11,7 +11,7 @@ import { onUnmounted, watch } from 'vue'
  * to hide other elements when it is clicked or focused.
  */
 export function useHideOthers(target: MaybeElementRef) {
-  let undo: ReturnType<typeof hideOthers>
+  let undo: ReturnType<typeof suppressOthers>
   watch(() => unrefElement(target), (el) => {
     // disable hideOthers on test mode
     if (import.meta.env.MODE === 'test')
@@ -24,7 +24,7 @@ export function useHideOthers(target: MaybeElementRef) {
     }
     catch {}
     if (el && !isInsideClosedPopover)
-      undo = hideOthers(el)
+      undo = suppressOthers(el)
     else if (undo)
       undo()
   })


zernonia:
Thanks for looking into this! I tried the change in Chromium, and switching to `suppressOthers` (which uses `inert`) breaks a couple of things:

- **Focus doesn't return on close.** Dialog, Select and modal Popover focus the trigger synchronously while closing, but `inert` is only removed later in `onUnmounted`. The trigger is still inert at that point, so focus ends up on `<body>`.
- **Toasts and live regions stop working.** `hideOthers` leaves `[aria-live]` elements alone, but `inertOthers` doesn't. With a Dialog open, toast buttons (Reka Toast, vue-sonner) can't be clicked. Outside-click handlers also get `<html>` as `event.target`, so `closest(...)` checks stop matching.
- The marker attribute changes from `data-aria-hidden` to `data-suppressed`, even in browsers without `inert` support.

For comparison, Radix keeps `hideOthers` (radix-ui/primitives#2459), and Floating UI makes `inert` opt-in. Could you rework this as an opt-in option (e.g. a prop or a ConfigProvider setting)? It would also need to remove `inert` before the close handlers run, keep `[aria-live]` regions interactive, and include a test for focus returning to the trigger. Thanks!

PR #2935, opened 2026-09-10
fix(Drawer): apply initialFocus and finalFocus on the content

### LINKED ISSUE #2934: [Bug]: Drawer: initialFocus and finalFocus are declared and documented but never applied
### Environment

```
reka-ui 2.10.4 (also present on the v2 branch source)
Vue 3.5.42, Vite, Chrome / Safari / Firefox
```

The behaviour is visible in the source. `packages/core/src/Drawer/DrawerContentImpl.vue` on `v2`:

- The props declare `initialFocus?: boolean | HTMLElement | null` and `finalFocus?: boolean | HTMLElement | null`, with the doc comment "`false`: do not focus anything" and "`false`: do not restore focus".
- The template renders `<FocusScope :trapped="props.trapFocus" @mount-auto-focus="emits('openAutoFocus', $event)" ...>`.
- Nothing in the file reads `props.initialFocus` or `props.finalFocus`. `DrawerContent.vue` declares the same two props and forwards them, so they end in `DrawerContentImpl` unread.

### Steps to reproduce

1. Render a `DrawerContent` with `:initial-focus="false"` and an `<input>` inside it.
2. Open the drawer.
3. The input gets focus. On a phone, the keyboard opens while the sheet is still sliding in.
4. Render with `:final-focus="false"`, open and close the drawer. Focus returns to the trigger.

### Describe the bug

`initialFocus` and `finalFocus` on `DrawerContent` are documented and typed but have no effect. `docs/superpowers/plans/2026-04-09-drawer-baseui-parity-fixes.md` lists the intended work: "add `initialFocus` / `finalFocus` props, wire to `FocusScope`" and "accept and pass through `initialFocus`/`finalFocus`". The props landed; the wiring did not.

A consumer can get the `false` case today by cancelling the `openAutoFocus` / `closeAutoFocus` events. The element form has no workaround short of focusing by hand.

### Expected behavior

`initialFocus` and `finalFocus` do what the doc comment says:

- `true` or unset: `FocusScope` focuses the first tabbable element on open, and the trigger on close.
- `false`: no focus on open, no restore on close.
- an element: that element receives focus.

We hit this while moving a bottom sheet from vaul to Reka. A sheet that focuses a field on open pulls the on-screen keyboard up during the slide-in, so we want `initial-focus="false"` as the default for drawers.

lazerg:
### 🔗 Linked issue

Fixes #2934

### ❓ Type of change

### 📚 Description

`DrawerContent` declares and documents `initialFocus` and `finalFocus`, but nothing ever reads them, so a drawer always focuses its first tabbable element on open and always sends focus back to the trigger on close.

`initialFocus` is now read in `DrawerContentImpl`'s `mount-auto-focus` handler: `false` cancels the event so `FocusScope` skips its default focus, an element cancels it and focuses that element instead. `finalFocus` is read in `DrawerContent`, where the close handlers already decide what to restore focus to — `false` restores nothing, an element gets the focus the trigger would have had. Both keep the current behaviour when unset or `true`, and a consumer that calls `preventDefault()` on `openAutoFocus`/`closeAutoFocus` still wins.

n/a

### 📝 Checklist

## Summary by CodeRabbit

- **New Features**
  - Added configurable focus behavior when opening drawers, including disabling automatic focus or selecting a specific element.
  - Added configurable focus behavior when closing drawers, including restoring focus to a chosen element, the trigger, or nowhere.
  - Drawer focus options now default to focusing on open and returning focus on close.

- **Bug Fixes**
  - Improved focus handling for modal and non-modal drawers.
  - Added support for selecting text when focusing an element on open.

- **Documentation**
  - Documented the default focus behavior for drawer props.

### DIFF
--- a/docs/content/meta/DrawerContent.md
+++ b/docs/content/meta/DrawerContent.md
@@ -25,7 +25,8 @@
     'name': 'finalFocus',
     'description': '<p>Final focus target when the drawer closes.</p>\n<ul>\n<li><code>true</code> / default: focus the trigger</li>\n<li><code>false</code>: do not restore focus</li>\n<li>element ref: focus that specific element</li>\n</ul>\n',
     'type': 'boolean | HTMLElement | null',
-    'required': false
+    'required': false,
+    'default': 'true'
   },
   {
     'name': 'forceMount',
@@ -37,7 +38,8 @@
     'name': 'initialFocus',
     'description': '<p>Initial focus target when the drawer opens.</p>\n<ul>\n<li><code>true</code> / default: focus the first focusable element inside</li>\n<li><code>false</code>: do not focus anything</li>\n<li>element ref: focus that specific element</li>\n</ul>\n',
     'type': 'boolean | HTMLElement | null',
-    'required': false
+    'required': false,
+    'default': 'true'
   }
 ]" />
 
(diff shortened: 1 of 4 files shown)


zernonia:
Thanks for the fix! One issue before we can merge:

`initialFocus` / `finalFocus` are typed `boolean | HTMLElement | null`, so Vue casts them to `false` when they're not passed. A plain `<DrawerContent>` no longer focuses its content when it opens or returns focus to the trigger when it closes. The tests don't catch this because the wrapper always binds the props, and passing `undefined` explicitly skips the cast.

Could you please:
- add `withDefaults(..., { initialFocus: true, finalFocus: true })` in both `DrawerContent.vue` and `DrawerContentImpl.vue`
- add a test that renders `DrawerContent` without binding either prop
- run `pnpm docs:gen` so the docs show the new defaults

Thanks!

lazerg:
Good catch, that cast bit exactly as you describe. Fixed in 52db0362:

- `withDefaults` with `initialFocus: true` / `finalFocus: true` in both `DrawerContent.vue` and `DrawerContentImpl.vue`. Both needed it, since `DrawerContent` forwards its own resolved props down to the impl.
- New test mounts `DrawerContent` with neither prop bound and checks focus on open and on close. It fails on 52bf27d and passes now.
- Ran `pnpm docs:gen`, `DrawerContent.md` now lists `true` as the default for both.

PR #2936, opened 2026-09-12
fix(ContextMenu): support controlled open state

### LINKED ISSUE #2906: [Bug]: Cannot open/close ContextMenu programmatically
### Environment

```bash
Developement/Production OS: Arch Linux Rolling Release
Node version: 24.18.0
Package manager: npm@12.0.2
Reka UI version: 2.10.4
Vue version: 3.5.32
Nuxt version: 4.4.2
Nuxt mode: csr?
Nuxt target: client
CSS framework: tailwindcss@4.2.2
Client OS: Arch Linux Rolling Release
Browser: Chromium 151.0.7922.173 and Firefox 154.0.1
```

https://stackblitz.com/edit/vitejs-vite-rvsbwr4v?file=src%2FApp.vue

### Steps to reproduce

Try to open context menu from the button

### Describe the bug

Updating the v-model:open ref via a button doesn't reflect on the context menu

### Expected behavior

Chanigng the v-model:open ref should be used by context menu as expected

might give more context or be useful
Similar issue I found on radix: https://github.com/radix-ui/primitives/issues/2795
their fix: https://github.com/radix-ui/primitives/pull/3891

Chaweki:
### 🔗 Linked issue

Fixes #2906

### ❓ Type of change

### 📚 Description

`ContextMenuRoot` currently owns an internal `open` ref and emits `update:open`, but it does not react when a controlled `open` value changes from the parent.

This updates `ContextMenuRoot` to support controlled `open` state using the existing `useVModel` pattern while preserving the current uncontrolled right-click behavior.

Programmatic opening cannot rely on the pointer-based virtual anchor because no pointer event occurs. Programmatically opened menus therefore use the trigger element as their positioning reference, while pointer-opened menus continue to use the pointer coordinates.

This also prevents a programmatic reopen from reusing stale pointer coordinates from a previous right-click.

A controlled variant was added to the ContextMenu story for manual verification.

Regression coverage was added for:
- programmatically opening the context menu
- programmatically closing the context menu
- positioning a programmatically opened menu relative to its trigger
- resetting pointer-based positioning before a programmatic reopen
- preserving the existing right-click behavior

Validation:
- ContextMenu tests: 7 passing
- `pnpm lint`: 0 errors
- `pnpm --filter reka-ui type-check`: passing
- `pnpm --filter reka-ui build`: passing
- `pnpm docs:gen`: currently blocked by the existing `@babel/traverse` issue on `v2`

N/A

### 📝 Checklist

## Summary by CodeRabbit

* **New Features**
  * Added support for controlling context menu open state programmatically, including external open/close controls.
  * Added examples for controlled context menus and rejecting open requests.
  * Context menus now position correctly when opened programmatically after pointer-based interactions.

* **Bug Fixes**
  * Prevented stale pointer coordinates from affecting the position of subsequently opened menus.

* **Documentation**
  * Documented the controlled `open` state and `v-model:open` usage.

### DIFF
--- a/packages/core/src/ContextMenu/ContextMenu.test.ts
+++ b/packages/core/src/ContextMenu/ContextMenu.test.ts
@@ -1,17 +1,19 @@
 import type { VueWrapper } from '@vue/test-utils'
-import { findByRole } from '@testing-library/vue'
+import { findByRole, queryByRole } from '@testing-library/vue'
 import { mount } from '@vue/test-utils'
 import { beforeEach, describe, expect, it } from 'vitest'
 import { axe } from 'vitest-axe'
 import ContextMenu from './story/_ContextMenu.vue'
+import ControlledContextMenu from './story/_ControlledContextMenu.vue'
+
+globalThis.ResizeObserver = class ResizeObserver {
+  observe() {}
+  unobserve() {}
+  disconnect() {}
+}
 
 describe('given default ContextMenu', () => {
   let wrapper: VueWrapper<InstanceType<typeof ContextMenu>>
-  globalThis.ResizeObserver = class ResizeObserver {
-    observe() {}
-    unobserve() {}
-    disconnect() {}
-  }
 
   beforeEach(() => {
     document.body.innerHTML = ''
@@ -33,6 +35,33 @@ describe('given default ContextMenu', () => {
 
     it('should render the menu', async () => {
       expect(await findByRole(document.body, 'menu')).toBeTruthy()
+    })
+  })
+})
+
+describe('given controlled ContextMenu', () => {
+  let wrapper: VueWrapper<InstanceType<typeof ControlledContextMenu>>
+
+  beforeEach(() => {
+    document.body.innerHTML = ''
+    wrapper = mount(ControlledContextMenu, { attachTo: document.body })
+  })
+
+  describe('when open state is changed programmatically', () => {
+    it('should render the menu', async () => {
+      await wrapper.find('#open-button').trigger('click')
+
+      expect(await findByRole(document.body, 'menu')).toBeTruthy()
+    })
+
+    it('should close the menu', async () => {
+      await wrapper.find('#open-button').trigger('click')
+
+      expect(await findByRole(document.body, 'menu')).toBeTruthy()
+
+      await wrapper.find('#close-button').trigger('click')
+
+      expect(queryByRole(document.body, 'menu')).toBeNull()
     })
   })
 })
--- a/packages/core/src/ContextMenu/ContextMenuRoot.vue
+++ b/packages/core/src/ContextMenu/ContextMenuRoot.vue
@@ -13,7 +13,7 @@ type ContextMenuRootContext = {
   pressOpenDelay: Ref<number>
 }
 
-export interface ContextMenuRootProps extends Omit<MenuProps, 'open'> {
+export interface ContextMenuRootProps extends MenuProps {
   /**
    * The duration from when the trigger is pressed until the menu opens.
    *
@@ -28,7 +28,8 @@ export const [injectContextMenuRootContext, provideContextMenuRootContext]
 </script>
 
 <script setup lang="ts">
-import { ref, toRefs, watch } from 'vue'
+import { useVModel } from '@vueuse/core'
+import { ref, toRefs } from 'vue'
 import { MenuRoot } from '@/Menu'
 
 defineOptions({
@@ -38,13 +39,17 @@ defineOptions({
 const props = withDefaults(defineProps<ContextMenuRootProps>(), {
   modal: true,
   pressOpenDelay: 700,
+  open: undefined,
 })
 const emits = defineEmits<ContextMenuRootEmits>()
 const { dir: propDir, modal, pressOpenDelay } = toRefs(props)
 useForwardExpose()
 const dir = useDirection(propDir)
 
-const open = ref(false)
+const open = useVModel(props, 'open', emits, {
+  defaultValue: false,
+  passive: (props.open === undefined) as false,
+}) as Ref<boolean>
 const triggerElement = ref<HTMLElement>()
 
 provideContextMenuRootContext({
@@ -56,10 +61,6 @@ provideContextMenuRootContext({
   modal,
   triggerElement,
   pressOpenDelay,
-})
-
-watch(open, (value) => {
-  emits('update:open', value)
 })
 </script>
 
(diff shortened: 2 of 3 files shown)


zernonia:
Thanks for this! One issue: when the menu is opened via the `open` prop, it's positioned wrong. `point` in `ContextMenuTrigger.vue` is only set in `handleOpen` (right-click / long-press), so:

- before any right-click, the menu opens at the viewport's top-left (`{ x: 0, y: 0 }`)
- after a right-click, it reopens at that old cursor position

I confirmed this in jsdom with the new `_ControlledContextMenu` story (the menu wrapper got `translate(2px, 0px)`, then `translate(298px, 400px)` on reopen after a right-click at 300,400). The current tests only check that the menu exists, so they don't catch it.

Suggested fix: when `open` turns true without a pointer event, anchor to the trigger element instead, e.g. set `point` from `rootContext.triggerElement.value.getBoundingClientRect()` (or use the element itself as the reference). A test that asserts the position would be good too.

Chaweki:
> Thanks for this! One issue: when the menu is opened via the `open` prop, it's positioned wrong. `point` in `ContextMenuTrigger.vue` is only set in `handleOpen` (right-click / long-press), so:
> 
> * before any right-click, the menu opens at the viewport's top-left (`{ x: 0, y: 0 }`)
> * after a right-click, it reopens at that old cursor position
> 
> I confirmed this in jsdom with the new `_ControlledContextMenu` story (the menu wrapper got `translate(2px, 0px)`, then `translate(298px, 400px)` on reopen after a right-click at 300,400). The current tests only check that the menu exists, so they don't catch it.
> 
> Suggested fix: when `open` turns true without a pointer event, anchor to the trigger element instead, e.g. set `point` from `rootContext.triggerElement.value.getBoundingClientRect()` (or use the element itself as the reference). A test that asserts the position would be good too.

@zernonia Thanks for catching this! Fixed in 7fc243b.

Programmatic opens now use the trigger element as the positioning reference, while pointer-triggered opens continue using the virtual pointer position. This also prevents programmatic reopens from reusing stale pointer coordinates.

I added regression coverage for both the initial programmatic open and reopening programmatically after a right-click, plus a controlled story variant for manual verification.

zernonia:
Thanks @Chaweki — the positioning fix works. I checked out 7fc243b and confirmed programmatic opens now anchor to the trigger; tests (7/7) and `type-check` pass.

One thing left, which is the edge case CodeRabbit raised — I reproduced it. `handleOpen` sets `isPointerOpen = true` before asking the parent to open. If a controlled parent refuses (keeps `open` false), `rootContext.open` never changes, so the watcher never resets the flag, and the next programmatic open reuses the refused right-click's coordinates.

With a parent that rejects open requests: right-click at (300, 400) refused, then a programmatic open placed the menu at `translate(298px, 400px)` instead of the trigger's `translate(98px, 100px)`.

Clearing the flag when the request is refused fixes it, and your 7 tests still pass with it applied:

```ts
async function handleOpen(event: MouseEvent | PointerEvent) {
  isPointerOpen.value = true
  point.value = { x: event.clientX, y: event.clientY }
  rootContext.onOpenChange(true)
  await nextTick()
  // a controlled parent may have refused the request
  if (!rootContext.open.value)
    isPointerOpen.value = false
}
```

Two smaller notes:

- `ContextMenuContent` hardcodes `side="right"` / `align="start"` / `sideOffset={2}`, which was tuned for a zero-size cursor point. Anchored to a real element, a programmatic open lands just off the trigger's right edge — for a large trigger area that may be further from the trigger than expected. Anchoring to a virtual point at the trigger's top-left corner would feel closer to a native context menu. Happy either way, just flagging it since jsdom can't show how it looks.
- `docs/content/meta/ContextMenuRoot.md` still doesn't list `open` — needs a `pnpm docs:gen` run before merge.

Chaweki:
> Thanks @Chaweki — the positioning fix works. I checked out [7fc243b](https://github.com/unovue/reka-ui/commit/7fc243bed6b66cdd3440e5fe18549ed1f49a7212) and confirmed programmatic opens now anchor to the trigger; tests (7/7) and `type-check` pass.
> 
> One thing left, which is the edge case CodeRabbit raised — I reproduced it. `handleOpen` sets `isPointerOpen = true` before asking the parent to open. If a controlled parent refuses (keeps `open` false), `rootContext.open` never changes, so the watcher never resets the flag, and the next programmatic open reuses the refused right-click's coordinates.
> 
> With a parent that rejects open requests: right-click at (300, 400) refused, then a programmatic open placed the menu at `translate(298px, 400px)` instead of the trigger's `translate(98px, 100px)`.
> 
> Clearing the flag when the request is refused fixes it, and your 7 tests still pass with it applied:
> 
> ```ts
> async function handleOpen(event: MouseEvent | PointerEvent) {
>   isPointerOpen.value = true
>   point.value = { x: event.clientX, y: event.clientY }
>   rootContext.onOpenChange(true)
>   await nextTick()
>   // a controlled parent may have refused the request
>   if (!rootContext.open.value)
>     isPointerOpen.value = false
> }
> ```
> 
> Two smaller notes:
> 
> * `ContextMenuContent` hardcodes `side="right"` / `align="start"` / `sideOffset={2}`, which was tuned for a zero-size cursor point. Anchored to a real element, a programmatic open lands just off the trigger's right edge — for a large trigger area that may be further from the trigger than expected. Anchoring to a virtual point at the trigger's top-left corner would feel closer to a native context menu. Happy either way, just flagging it since jsdom can't show how it looks.
> * `docs/content/meta/ContextMenuRoot.md` still doesn't list `open` — needs a `pnpm docs:gen` run before merge.

Thanks @zernonia for reproducing this and for the detailed feedback!

I’ve addressed the rejected controlled-open case by resetting the pointer state when the parent doesn’t accept the open request after the next tick. I also added regression coverage for the exact sequence you described: rejected right-click at (300, 400) followed by a programmatic open, which now correctly anchors back to the trigger instead of reusing the stale pointer coordinates.

The ContextMenu suite is now 8/8 passing.

I kept the trigger-element reference for programmatic opens for now, while pointer opens continue to use the cursor position.

I also regenerated ContextMenuRoot.md, so the controlled open prop is now included in the generated API docs.

Thanks again for catching the edge case!

Chaweki:
Hi @zernonia , just following up on this one when you have a chance. I addressed the rejected controlled-open case and regenerated the ContextMenuRoot docs in the latest commit. The ContextMenu suite is 8/8 passing and there are no remaining CodeRabbit findings. Thanks again!

Issue #2200, opened 2025-10-02
[Bug]: NumberField increment/decrement button click bugs on iOS


lubomirblazekcz:
### Environment

```bash
not relevant
```

Bug is present on iOS 18 or 26

### Link to minimal reproduction

https://reka-ui.com/docs/components/number-field

### Steps to reproduce

Try to click on the demo page on the increment/decrement, often when you click on the edge the button registers the focus/active state but the number will not change.

### Describe the bug

When the area of the increment/decrement button is small, half the time the click is not registered, even when focus/active state is visualy visible.

### Context & Screenshots (if applicable)

https://github.com/user-attachments/assets/892b3fea-0aab-4a76-a977-c4f3b33e0978

AANDDARK:
Hi @lubomirblazekcz  i have a draft with maybe fix of your bug can you some check it because i dont have iPhone

lubomirblazekcz:
nope, that didn't fix the issue, the `handleChangingValue` is not triggered at all when the "not working" click happens

lubomirblazekcz:
after some debuging, the [pointerdown](https://github.com/unovue/reka-ui/blob/b681496f59a3b7bf81eb284f2e3b1d73dc252dda/packages/core/src/NumberField/utils.ts#L54) isn't triggered at all sometimes (often when clicking on the button side). When I added `@click` on the button it is triggered everytime.

AANDDARK:
idk how fix it because close request, but thanks @lubomirblazekcz  for collaborate.

JustusBaumhove:
Two related things in `NumberField` that combine into click bugs.

**1. The trigger becomes `disabled` between `pointerdown` and `pointerup` of the same press.**

`usePressedHold` applies the step on `pointerdown`. So on the press that reaches `min`/`max`, `NumberFieldIncrement`/`NumberFieldDecrement` recompute `isDisabled` and set `disabled` on the element **while it is still being pressed**.

Consumers commonly style `disabled` with `pointer-events: none` (the Tailwind `disabled:pointer-events-none` convention). That drops the pressed element out of hit-testing mid-gesture, so the browser retargets the rest of the tap at whatever.

On iOS this is considerably worse: WebKit synthesises `click` at `touchend` and hit-tests the touch point at dispatch time, so the click can land on an entirely unrelated element elsewhere on the page.

**2. `focusOnChange` focuses the input from inside the pointerdown handler.**

[...]

PR #2948, opened 2026-09-17
fix(NumberField): defer touch stepper activation

### LINKED ISSUE #2947: [Bug]: NumberField stepper buttons activate while starting a touch scroll
### Environment

```bash
Development/Production OS: macOS Tahoe 26.3.1 (a)
Node version: 24.14.0
Package manager: npm@11.6.2
Reka UI version: 2.8.0; also reproducible on the current v2 branch (2.10.4)
Vue version: 3.5.28
Nuxt version: N/A
Nuxt mode: N/A
Nuxt target: N/A
CSS framework: tailwindcss@4.1.14
Client OS: macOS Tahoe 26.3.1 (a)
Browser: Chrome 152.0.7977.76
```

https://www.reka-ui.com/docs/components/number-field

### Steps to reproduce

1. Open the NumberField documentation on a touchscreen device.
2. Place a finger on an increment or decrement button.
3. Without releasing the finger, immediately drag vertically as if scrolling the page.
4. Observe that the value changes on touch-down, before the gesture has resolved into either a tap or a scroll.
5. Repeat while holding the button stationary and confirm that the existing hold-to-repeat behavior starts immediately.

### Describe the bug

NumberField increment and decrement controls activate on `pointerdown` for
touch pointers. Consequently, merely beginning a touch gesture over a stepper
button changes the value, even when the user's intended gesture is to scroll.

A touch interaction should not be committed until it can be distinguished from
a scroll gesture. Native button clicks normally activate on release and are
cancelled when the touch becomes a scroll.

The existing hold-to-repeat behavior should be preserved, but touch and pen
pointers should first wait for either:

- release, producing one step;
- a stationary long press, starting repeat; or
- sufficient pointer movement / `pointercancel`, cancelling the press.

Mouse behavior can remain immediate.

This appears related to the broader iOS pointer-down interaction described in
#2200, although this report specifically concerns accidental activation while
scrolling.

I have implemented and tested a fix and intend to submit a PR.

### Expected behavior

- A touch tap increments or decrements once when the finger is released.
- A stationary touch hold starts repeating after a short delay.
- Moving beyond a small tolerance before the hold activates cancels the step
  and allows native scrolling.
- Moving beyond the touch tolerance after repeat begins stops further repetition.
- `pointercancel` cancels the interaction.
- Mouse presses retain their current immediate behavior.

_No response_

kaspernowak:
### 🔗 Linked issue
Closes #2947

Related to #2200.

### ❓ Type of change

### 📚 Description

NumberField stepper controls currently trigger immediately on `pointerdown`.
On touch devices this changes the value before the browser can distinguish a
tap or long press from a scrolling gesture.

This change preserves immediate mouse activation while changing touch and pen
behavior to:

- trigger a tap on release;
- begin hold-to-repeat after the existing 400 ms delay;
- cancel when movement exceeds a 10 px tolerance;
- stop repetition when movement exceeds the 10 px tolerance;
- cancel on `pointercancel`.

The implementation remains inside the existing NumberField `usePressedHold`
utility and does not change the public API.

Regression tests cover mouse activation, touch taps, stationary holds,
movement before and after hold activation, and pointer cancellation.

Not applicable; this changes pointer interaction behavior without changing
rendered output.

### 📝 Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved NumberField interactions for mouse, touch, and pen input.
  - Mouse presses activate immediately, while touch and pen taps activate on release.
  - Press-and-hold controls repeat after a brief delay.
  - Small movements no longer interrupt an active hold.
  - Larger touch or pen movements and pointer cancellation stop activation.
  - Disabled controls no longer trigger value changes.


packages/core/src/NumberField/utils.ts
```diff
@@ -3,59 +3,120 @@ import type { Ref } from 'vue'
...
+  const onPressMove = (event: PointerEvent) => {
+    if (event.pointerId !== activePointerId || activePointerType === 'mouse')
+      return
+
+    const distanceX = event.clientX - startX
+    const distanceY = event.clientY - startY
+    const movedBeyondTolerance = distanceX * distanceX + distanceY * distanceY > TOUCH_MOVE_TOLERANCE ** 2
+
+    if (movedBeyondTolerance)
+      resetPress()
```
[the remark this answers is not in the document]
kaspernowak: Thanks for flagging this. The 10 px touch slop is intentional both before and after hold activation so normal finger jitter does not cancel hold-to-repeat. The previous implementation did not explicitly cancel on pointerleave; it stopped on pointerup or pointercancel. The new implementation preserves those cancellation paths and additionally cancels when movement exceeds the tolerance.
I added regression coverage in 8a23105d: a 5 px movement after hold activation continues repetition, while the existing test verifies that movement beyond 10 px stops it.

zernonia (changes requested):
Thanks, the logic and tests look good. Tests, lint and type-check pass locally on the head commit. One change I'd like before merging, plus two optional nits.

**Keep `event.preventDefault()` for touch and pen too** (see inline comment). Cancelling `pointerdown` never blocked scrolling, but it did suppress the compatibility `mousedown`, which is what keeps focus on the input after a tap on Chrome for Android.

Optional:
- Check `disabled` at the top of `onPressStart` before scheduling. Right now a touch press on a disabled stepper sets `isPressed` / `data-pressed` for up to 400 ms or until release, because the disabled check only runs inside `trigger()`.
- The window `pointermove` listener is now registered permanently for every stepper. The early return makes it cheap, so fine as is, but it could be attached only while a press is active.

Nice that `onScopeDispose(resetPress)` also fixes the pre-existing leaked repeat timer on unmount mid-hold, and that release activation sidesteps the mid-press `disabled` retargeting from #2200 for taps.

packages/core/src/NumberField/utils.ts
```diff
@@ -3,59 +3,120 @@ import type { Ref } from 'vue'
...
+
+    if (event.pointerType === 'mouse') {
+      event.preventDefault()
+      holdTriggered = true
+      if (trigger())
+        scheduleRepeat(HOLD_DELAY)
+      return
+    }
+
+    scheduleRepeat(HOLD_DELAY)
```
zernonia, line 80: Please hoist `event.preventDefault()` above the `pointerType` branch so it runs for touch and pen as well.

Per the Pointer Events spec, panning/zooming is not a default action of any pointer event, so cancelling `pointerdown` does not affect native scrolling (and the browser still fires `pointercancel` when it takes over the gesture). What cancelling does do is suppress the compatibility `mousedown`.

Without it, on Chrome for Android a tap goes: `pointerup` → `trigger()` → root `focusOnChange` focuses the input → compat `mousedown` on the stepper (`tabindex="-1"`, so focusable by pointer) → input blurs. The input's `@blur` reformats the value and the soft keyboard opens and closes inside one tap, which is the layout-shift problem from #2200, just on Android. iOS Safari doesn't focus buttons on tap, so it isn't affected.

```suggestion
    event.preventDefault()

    if (event.pointerType === 'mouse') {
      holdTriggered = true
      if (trigger())
        scheduleRepeat(HOLD_DELAY)
      return
    }

    scheduleRepeat(HOLD_DELAY)
```

The `defaultPrevented` assertion in the touch tap test (`NumberField.test.ts:137`) would flip to `true`.

kaspernowak: Thanks for the detailed explanation. Addressed in 4526e72e: `preventDefault()` now runs before the pointer-type branch, and the touch assertion verifies that compatibility mouse events are suppressed for touch and pen.

I also accepted the disabled-state suggestion by returning before initializing press state or scheduling a timer, with regression coverage confirming that `data-pressed` is not set and the value remains unchanged.

I left the window `pointermove` listener registered through `useEventListener`, since its inactive path is inexpensive and this keeps the pointer lifecycle straightforward.

Verification completed:
- NumberField tests: 49 passed
- ESLint: passed
- Type-check: passed
- Core build: passed

PR #2961, opened 2026-09-22
fix(Calendar): keep the visible month when modelValue is a new object for the same day

### LINKED ISSUE #2960: [Bug]: Calendar month resets when modelValue is a new object for the same day
### Environment

```bash
Developement/Production OS: Windows 10
Node version: 24.5.0
Package manager: pnpm
Reka UI version: 2.10.5 (also reproduced on 2.6.0; still present on the v2 branch)
Vue version: 3.5
Nuxt version: n/a
Nuxt mode: n/a
Nuxt target: n/a
CSS framework: n/a
Client OS: Windows 10
Browser: Chrome
```

https://codesandbox.io/p/sandbox/7ctgcx

### Steps to reproduce

1. Open the sandbox. The calendar shows July 2026 because `modelValue` is `2026-07-01`.
2. Click **Next page**. The heading becomes August 2026.
3. Click **Pass a new CalendarDate for the same day**.

The button only increments a counter so the computed returns `new CalendarDate(2026, 7, 1)` again. The selected calendar day does not change.

```vue
<script setup>
import { CalendarDate } from '@internationalized/date'
import { computed, ref } from 'vue'

const tick = ref(0)

const modelValue = computed(() => {
  tick.value
  return new CalendarDate(2026, 7, 1)
})
</script>
```

### Describe the bug

After paging to another month, passing a new `DateValue` object for the **same calendar day** moves `placeholder` back to the selected day. The heading snaps from August 2026 back to July 2026.

`CalendarRoot` watches `modelValue` and compares the new value with the current `placeholder`, not with the previous selected day:

```ts
watch(modelValue, (_modelValue) => {
  if (Array.isArray(_modelValue) && _modelValue.length) {
    const lastValue = _modelValue.at(-1)
    if (lastValue && !isEqualDay(placeholder.value, lastValue))
      onPlaceholderChange(lastValue)
  }
  else if (!Array.isArray(_modelValue) && _modelValue && !isEqualDay(placeholder.value, _modelValue)) {
    onPlaceholderChange(_modelValue)
  }
})
```

Vue runs that watch whenever the object identity changes. Next/prev only updates `placeholder` (for example to 1 August). A later render that constructs another `CalendarDate` for 1 July is a different object, `isEqualDay(1 August, 1 July)` is false, and the visible month jumps back.

This shows up whenever a parent builds the date during render, which is a normal way to pass `@internationalized/date` values into a controlled calendar.

I can open a PR for this.

Suggested change: ignore the update when the selected day is unchanged, and only move `placeholder` when it actually changes.

```ts
watch(modelValue, (value, previous) => {
  if (isSameSelectedDate(previous, value))
    return

  const focused = Array.isArray(value) ? value.at(-1) : value
  if (focused && !isEqualDay(placeholder.value, focused))
    onPlaceholderChange(focused)
})
```

`isSameSelectedDate` should use `isEqualDay` (and compare each day for `multiple`). Clearing the value should still leave `placeholder` where it is, which is what the current watch does.

Two cases worth locking in a test:

- Same day, new object, after clicking Next page: the heading stays on the next month.
- Selected day changes from 1 July to 23 September: the heading follows the new day.

`DatePickerRoot` has the same shape (`value.compare(placeholder) !== 0` with no previous-value check). This report is only for `CalendarRoot`.

### Expected behavior

A new `modelValue` object for the same calendar day does not change the visible month. The visible month moves only when the selected day itself changes.

Still present in `packages/core/src/Calendar/CalendarRoot.vue` on `v2` (`_modelValue.at(-1)`), so this is not fixed in 2.10.5.

jevin98:
## Summary

- `CalendarRoot` moved `placeholder` whenever `modelValue` changed identity and was not the same day as the current view. A parent re-render that builds a new `DateValue` for the same day undid month paging.
- Compare the new value with the previous selected day. The visible month still follows a real day change, including `multiple`.
- Fixes #2960. Reproduction: https://codesandbox.io/p/sandbox/7ctgcx

## Test plan

- [x] `pnpm exec vitest run src/Calendar/Calendar.test.ts` in `packages/core` (57 passed)
- [ ] Click Next page, then pass `selected.copy()` and confirm the heading stays on the next month
- [ ] Change the selected day to another month and confirm the heading follows it

## Summary by CodeRabbit

* **Bug Fixes**
  * Calendar views now remain on the current month when selections are replaced with equivalent date values.
  * Calendar views correctly move to the newly selected month when the selected day changes.
  * Multiple-date selections now synchronize the visible month based on the latest selected date.
  * Placeholder updates are prevented when selections are empty or unchanged.


zernonia:
Thanks! The `CalendarRoot` change looks good. However, DatePicker still jumps back to the selected month. `DatePickerRoot.vue` has its own `modelValue` watcher, and it compares the new value against `placeholder` rather than against the previous value. It then pushes that date down to `CalendarRoot`.

To reproduce: open a controlled DatePicker, go to the next month, then have the parent pass a new `CalendarDate` for the same day. The heading goes back to the selected month.

`MonthPickerRoot.vue` and `YearPickerRoot.vue` have the same watcher, so they jump back by month and by year. Could you apply the same comparison fix to those three roots in this PR, with a test for each?

jevin98:
Applied the same previous-value comparison to DatePickerRoot, MonthPickerRoot, and YearPickerRoot.

DatePickerRoot alone was not enough for the controlled DatePicker repro. DatePickerField forwards placeholder updates from DateFieldRoot, and DateFieldRoot had the same watcher, so a same-day object still wrote the selected day back onto the DatePicker placeholder. That watcher now bails out when the calendar day is unchanged.

When the day is unchanged, DatePickerRoot also skips closeOnSelect. A real day change still closes the popover when closeOnSelect is true.

Each root has a test that pages forward, replaces modelValue with an equal new object, and checks the heading stays put.

jevin98:
Follow-up on the duplicated comparison: Calendar, MonthPicker, and YearPicker had the same selection check, differing only by period (`isEqualDay`, `isSameYearMonth`, `isSameYear`).

That logic now lives in the internal date module as `isSameDateSelection` and `focusedDateValue` (`packages/core/src/date/comparators.ts`). They are not part of the public export. Each root passes its own comparator. DatePicker and DateField use the same helper for the single-value case. `closeOnSelect` stays in `DatePickerRoot`: an unchanged day skips the whole watcher, and a real day change still closes the popover.

jevin98:
@zernonia  done

